PR #5, opened 2016-02-08
Renamed "HalfOpen" to "HalfClosed" 

roltean:
Renamed "HalfOpen" to "HalfClosed" everywhere in code (methods, variables etc.) and documentation to match the State.HALF_CLOSED value.


RobWin:
The name HALF-OPEN is correct and used in the pattern description of Martin Fowler.
I have to change the state diagram and the documentation to be consistent.

PR #9, opened 2016-07-01
Add decorateCompletableFuture

tomfitzhenry:
I can add tests and an overload for `Supplier<CompletableFuture<T>>` if you're happy with this sort of change.


RobWin:
Hi,

do you want to secure a function call which runs async?
Is it not sufficient to just decorate the function?

Similar to this? http://www.javaslang.io/javaslang-circuitbreaker/0.5.0/#_use_a_completablefuture

Do I miss something important?

PR #11, opened 2016-07-07
Fix to allow retry durations over 10 ms

charithe:
The validation logic on custom `RetryContext` wait duration contains a casting bug which prevents wait durations under one second from being specified. This is because `duration.getSeconds()` returns a `long` -- which is always zero for durations under one second.


RobWin:
Thx

RobWin:
Btw if you like the CircuitBreaker, I would love to get your GitHub Star

PR #16, opened 2016-11-26
cleanup in the retry code: better name & signature for function

emmanueltouzery:
I think isRetryAllowed... did not properly describe the purpose of the function. Also returning a boolean was not necessary.


RobWin:
Thx

PR #18, opened 2016-11-27
Retry: sleep fixes and tests

emmanueltouzery:
so the Retry mechanism only respected waiting times at the second granularity (!!!). The first commit fixes that.
The second commit makes the sleeping function overrideable by the tests and the tests do override it and check the sleep times to make sure they are as expected. Notice that the change also speeds up the tests after the first commit, since after that first commit they start sleeping by amounts of 500ms.


emmanueltouzery:
done!

RobWin:
Thx

RobWin:
If you like javaslang-circuitbreaker, I would love to get your GitHub Star

PR #19, opened 2016-11-27
Retry: add a backoff function feature, to modify the waiting interval between failures

emmanueltouzery:


RobWin:
Thx,

do you want to join the gitter channel?
https://gitter.im/javaslang/javaslang-circuitbreaker

emmanueltouzery:
thanks, won't join the gitter for now. Only using retry, a tiny part of the library actually

Issue #12, opened 2016-11-22
Implement a Rate Limiter


RobWin:
Rate limiters are used to restrict the rate at which a backend is allowed to be accessed.

storozhukBM:
Didn't notice this issue. So repost my question here.
I've already look through [this](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java) and several other implementations. And have overall understanding of pros and cons of different algorithms. But I don't sure that such feature should be added to javaslang-circuitbreaker because it is already concise and beautiful single responsibility library. 
On my current project I've created abstraction called `Gateway`. Every API call is proxied through this `Gateway` and it is responsible for invocation of circuitBreaker, requestLimit and/or metrics.

Like that:
```java
public class GatewayInterceptor implements MethodInterceptor {
    private final Logger log = LoggerFactory.getLogger(GatewayInterceptor.class);
    private final ApiType apiType;
    private final Try.CheckedFunction<MethodInvocation, Object> gatewayInvoke;

    public GatewayInterceptor(ApiType apiType, RequestLimit requestLimit, CircuitBreaker circuitBreaker) {
        this.apiType = apiType;
        Try.CheckedFunction<MethodInvocation, Object> circuitBreakerInvoke = CircuitBreaker.decorateCheckedFunction(
            MethodInvocation::proceed,
            circuitBreaker
        );
        Try.CheckedFunction<MethodInvocation, Object> requestLimitInvoke = RequestLimit.decorateCheckedFunction(
            circuitBreakerInvoke,
            requestLimit
        );
        gatewayInvoke = requestLimitInvoke;
    }

    @Override public Object invoke(final MethodInvocation invocation) throws Throwable {
        Object result = Try.success(invocation)
            .mapTry(gatewayInvoke)
            .onFailure(throwable -> log.error("Error in invocation of {} API: {}", apiType, throwable.getMessage()))
            .get();
        return result;
    }
}
```
@tomfitzhenry What do you think about it? Should we enhance javaslang-circuitbreaker or create additional library that will compose  few other libraries?

RobWin:
@tomfitzhenry is not the maintainer, but it's me :)

If a RateLimiter implementation is lightweight in the sense that we don't need any other 3rd party libraries, we can add it to javaslang-circuitbreaker. A RateLimiter is also some sort of CircuitBreaker. A RateLimiter can forbid further calls, if a rate limit has exceeded.

[...]

PR #23, opened 2016-12-03
Issue #12 Implement a Rate Limiter

storozhukBM:
Finally it's here with tests, documentation update and all other stuff.


storozhukBM:
This coverage situation is super weird, because IDEA shows me completely different numbers
screenshot at dec 04 00-17-31

PR #26, opened 2016-12-06
Ratelimiter default values

RobWin:


src/main/java/io/github/robwin/ratelimiter/RateLimiter.java
```diff
@@ -168,7 +168,7 @@ static Runnable decorateRunnable(RateLimiter rateLimiter, Runnable runnable) {
     /**
      * Will wait for permission within default timeout duration.
      *
-     * @param rateLimiter the RateLimiter to get permission from
+     * @param rateLimiter the RateLimiter to getValueFromCache permission from
```
storozhukBM: ?

src/main/java/io/github/robwin/ratelimiter/RateLimiterConfig.java
```diff
@@ -18,46 +18,60 @@
...
+
 public class RateLimiterConfig {
     private static final String TIMEOUT_DURATION_MUST_NOT_BE_NULL = "TimeoutDuration must not be null";
     private static final String LIMIT_REFRESH_PERIOD_MUST_NOT_BE_NULL = "LimitRefreshPeriod must not be null";
     private static final Duration ACCEPTABLE_REFRESH_PERIOD = Duration.ofNanos(1L);
 
-    private final Duration timeoutDuration;
-    private final Duration limitRefreshPeriod;
-    private final int limitForPeriod;
+    private final Context context;
```
storozhukBM: Maybe we can get rid of Context by making all fields private non final? And builder will just set this during construction.

PR #28, opened 2016-12-08
Fix for race conditions if two threads increase the max error counter…

RobWin:
… at the same time.


RobWin:
Hi @jburell,

I think, the RetryContext should not be used by multiple threads.
Retry is different to CircuitBreaker and RateLimiter in that case.

But anyway thank you for fixing this issue.

RobWin:
Btw, if you like javaslang-circuitbreaker, we would love to get your GitHub star :)

jburell:
@RobWin While I agree that the behavior is not thread safe, and that I don't need it for my use case, what is the purpose of the AtomicInteger?

RobWin:
Hi, there is no purpose. I will fix this. I think this was introduces in a PR and I did not noticed that change.

RobWin:
Thx for your start

PR #30, opened 2016-12-08
switch from jacoco to cobertura + new tests + travis caches

storozhukBM:


storozhukBM:
Hmm it seems that coveralls show totally different numbers then IDEA. I should work on it more.

storozhukBM:
At least this numbers are similar to IDEA coverage report 🙈

storozhukBM:
I've switched jacoco to cobertura because it is able to handle tests that are using code instrumentation like PowerMock tests.

RobWin:
Does it work with coveralls?

storozhukBM:
Yes coverals check passed. Results [here](https://coveralls.io/builds/9216210)

PR #40, opened 2017-01-05
Issue #39 ring bit set copying constructor

storozhukBM:


src/main/java/io/github/robwin/circuitbreaker/internal/RingBitSet.java
```diff
@@ -48,6 +48,17 @@ public RingBitSet(int bitSetSize) {
         bitSet = new BitSetMod(bitSetSize);
     }
 
+    public RingBitSet(int bitSetSize, RingBitSet sourceSet) {
+        this(bitSetSize);
+
+        int targetLength = Integer.min(bitSetSize, sourceSet.length);
+        int sourceIndex = sourceSet.index;
+        for (int i = 0; i < targetLength; i++) {
```
RobWin: I think this should be in reverse order  so that we copy the latest call results. Or?

src/main/java/io/github/robwin/circuitbreaker/internal/RingBitSet.java
```diff
@@ -48,6 +48,17 @@ public RingBitSet(int bitSetSize) {
...
+            this.setNextBit(sourceSet.bitSet.get(sourceIndex));
+            sourceIndex = (sourceIndex + 1) % sourceSet.size;
```
RobWin: Sry I mean this Code line

PR #46, opened 2017-02-02
DOC: Fixed link to javaslang in "Use a CircuitBreaker"

cpilsworth:
Minor change just to fix the link to the javaslang github repo


RobWin:
Hi. Thx. I fixed it because the url of javaslang is different.

RobWin:
Btw. if you like and use this library, we would love to  get your github star.

Issue #47, opened 2017-02-22
OSGI support


david-t-daniel:
It would be great if osgi support was able to be added.  There is a bndtools gradle plugin that may be able to add the correct exports https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/README.md#gradle-plugin-for-workspace-builds

RobWin:
Hi,
good idea. If you have experience wit this plugin, would you like to create a PR and add this?

david-t-daniel:
I will give it a try but it may take a couple weeks since I have never used Gradle and have lots to do now. Still it is probably not a pressing issue as I am the first to ask for it. Thank you

RobWin:
Thx, no hurry :)

cpilsworth:
Created a PR with some very basic changes to add OSGi manifest headers. Just need to know if the some dependencies can be considered _optional_... see #56

cpilsworth:
@david-t-daniel take a look at the linked PR for this issue and see if it does what you need.

david-t-daniel:
I will do that this weekend.  Thank you very much for putting in the effort on this.

markusheer:
Hi

I tried to use Resilience4j in Apache Felix (Adobe Experience Manager). Activation of the bundles works but during runtime there is a class loading issue in io.github.resilience4j.circuitbreaker because the io.github.resilience4j.core package is not defined in Import-Package.

This is because in publishing.gradle the io.github.resilience4j package is excluded in the Import-Package section.

A solution is to change the exclude to to current projects package:
`instruction 'Import-Package', "!${project.osgiManifest().symbolicName}.*,*"`

RobWin:
Thx. Fixed it.
Could you please test `0.11.1-SNAPSHOT` when the build is complete.

markusheer:
That was fast. Tested via local build from master and it works.

RobWin:
Thx 👍

storozhukBM:
@RobWin should we make 0.11.1 release or wait for 0.12.0 ?

PR #56, opened 2017-03-15
Add osgi support

cpilsworth:
This PR implements #47 

This generates the OSGi META-INF/MANIFEST.MF file that decribes the jar as an OSGi bundle.

See https://docs.gradle.org/current/userguide/osgi_plugin.html

The library depends upon `javaslang` & `slf4j`, which both seems to be an [OSGi bundle]:(https://github.com/javaslang/javaslang/blob/master/pom.xml#L271-L284)
```
javaslang.collection,version=[2.0,3) from io.javaslang (482)
javaslang.control,version=[2.0,3) from io.javaslang (482)
org.slf4j,version=[1.7,2) from slf4j.api (17)
```

It also has the following dependencies which it expects to import - I wanted to check whether the following dependencies could be considered _optional_ - only required for certain features?
```
io.reactivex,version=[2.0,3) -- Cannot be resolved
io.reactivex.disposables,version=[2.0,3) -- Cannot be resolved
io.reactivex.functions,version=[2.0,3) -- Cannot be resolved
io.reactivex.processors,version=[2.0,3) -- Cannot be resolved
javax.cache,version=[1.0,2) -- Cannot be resolved
org.reactivestreams,version=[1.0,2) -- Cannot be resolved
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ before_install:
 - chmod +x gradlew
 
 script:
-- ./gradlew clean check allTests -x test
+- ./gradlew clean check jar allTests -x test
 
 after_success:
 - ./gradlew coberturaReport coveralls
--- a/build.gradle
+++ b/build.gradle
@@ -103,7 +103,7 @@ task javadocJar(type: Jar, dependsOn: javadoc) {
 jar {
     manifest {
         instruction 'Export-Package', '!io.github.robwin.*.internal,*'
-        instruction 'Import-Package', '!io.github.robwin.*,io.reactivex.*;resolution:=optional,javax.cache;resolution:=optional,org.reactivestreams;resolution:=optional,*'
+        instruction 'Import-Package', '!io.github.robwin.*,*'
         instruction 'Bundle-Description', 'A lightweight fault tolerance library'
         instruction 'Bundle-DocURL', 'http://robwin.github.io/javaslang-circuitbreaker/'
     }
--- a/build.gradle
+++ b/build.gradle
@@ -100,6 +100,15 @@ task javadocJar(type: Jar, dependsOn: javadoc) {
     from javadoc.destinationDir
 }
 
+jar {
+    manifest {
+        instruction 'Export-Package', '!io.github.robwin.*.internal,*'
+        instruction 'Import-Package', '!io.github.robwin.*,*'
+        instruction 'Bundle-Description', 'A lightweight fault tolerance library'
+        instruction 'Bundle-DocURL', 'http://robwin.github.io/javaslang-circuitbreaker/'
+    }
+}
+
 artifacts {
     archives sourcesJar
     archives javadocJar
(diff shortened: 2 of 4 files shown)


cpilsworth:
Judging by the docs support for RxJava and JCache is optional - people may choose not to use that functionality - so making those dependencies optional.  Let me know if this isn't right.

RobWin:
Hi, unfortunately RxJava is not optional at the moment.
We should have a look at it again, after we have modularized the project. 
The new home of the project will be: https://github.com/resilience4j

RobWin:
Good work so far :)  :+1:

cpilsworth:
Thanks.  The OSGi integration test loads the generated bundle from the gradle build folder - this works locally, but not on travis it seems.  Need to work out why.

cpilsworth:
@RobWin I think this is at a good point - let me know what you think.  

I could have the osgi loading test verify all of the packages, but it might cause you more problems if you are planning to modularise anyway.  If I can be any help in modularising so it does not have a hard dependency on rxjava etc, then let me know.

RobWin:
Thx for your effort. I  have a look at it next week.

storozhukBM:
Hey @cpilsworth,
This is really great work, thank you. I'll try to use it as example for all modules.

david-t-daniel:
@cpilsworth I can see the manifest is generated correctly.  The unit test is not passing for me when I do a build from the command line.  The normal way I do integration tests is by including the test class in the jar itself and then specifying the test class in the manifest file http://bnd.bndtools.org/chapters/310-testing.html.  I can also see how including integration tests in the jar is undesirable though.  I was not able to figure out how to use gradle with eclipse to run the tests but I can run them from the command line and will see if I can figure out why things are not working for me.

cpilsworth:
The unit test is an integration test really, it relies upon the jar having being generated, which it then loads.  Travis has been executing `./gradlew clean check jar allTests -x test`.  I should probably indicated this as a (post-build) integration test _or_ load bundle from the build classpath rather than the jar artefact.

I've not used the test method that you've referenced before.  I've not had a chance to read through the document in detail, but it does seems strange that it would be tested from the inside.  Part of what is being tested if the external interface being presented (e.g. packages/classes exported).

RobWin:
Hi, do you think we can start another try to OSGIfy Resilience4j?

david-t-daniel:
The patch worked for me and generated the headers correctly it was just a unit test thing.   I can try and take a look at getting the unit test running for me but if it works for you and Chris then I am good with a merge.

RobWin:
Yes, but we did a major refactoring and the branch cannot be merged anymore.

david-t-daniel:
I used Chris' technique for modifying the build.gradle file and verified that it works at the individual project level.  In 2 weeks during our cleanup sprint I will update our code to vavr from javaslang and make sure working headers are in the circuitbreaker and other projects we use.  Thenk you Chris for this work again.

cpilsworth:
Sorry, I've not found the time to update this since the project was split into modules.  There's been a lot of activity on the project since I looked at this!  

Not sure if it's still an issue, but at one time a couple of the modules shared a common base package - this could cause problems with multiple bundles exporting the same package.

PR #82, opened 2017-03-31
adding ratpack support, closes #74

### LINKED ISSUE #74: Host ratpack integration module
Hi,

I created https://github.com/drmaas/ratpack-resilience4j, and was wondering if you're interested in hosting it here. I see you have a number of other integrations. It provides method interceptors for circuit breaking methods that return promises (and completion stage, observable, flowable, and normal values), and operators to circuit break and retry promises reactively.

Ratpack is a reactive, non-blocking web framework built on Netty that we've been using in production for over a year now, and have been happy with it. https://ratpack.io/

Dan

drmaas:
reopened PR for ratpack support.

@RobWin PTAL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .gradle
 build
 classes
+/resilience4j-ratpack/src/test/groovy/io/github/resilience4j/ratpack/annotation/RateLimiterSpec.groovy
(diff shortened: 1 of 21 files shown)


RobWin:
Could you please add some tests, so that the coverage is stable.

drmaas:
I have to disagree somewhat with the coverage stats - they are claiming that some boilerplate code is not being executed (CircuitBreakerProvider and RateLimiterProvider). However, the tests wouldn't work at all if the providers were not being executed.

Other missed lines of code are extreme exception handling edge cases that I think would be difficult to test.

Issue #72, opened 2017-03-27
Create Dropwizard Metrics integration module 


storozhukBM:

drmaas:
@storozhukBM @RobWin is this being actively worked on? If not there are people in our org who want to work on it.

drmaas:
See #74 for reference

fromanator:
I work with @drmaas and would be happy to help with integrating dropwizard for metrics.

storozhukBM:
Hi guys, I'm also interested in this issue, but haven't started work yet (quite busy at work atm). So if you have time it would be great.

storozhukBM:
The only thing I've done so far is draft API design.
My plan was to create factory methods that can basically wrap our `io.github.resilience4j.circuitbreaker.CircuitBreaker.Metrics` or `io.github.resilience4j.ratelimiter.RateLimiter.Metrics` into dropwizard gauges.

API example
```java
//so we have some registries
CircuitBreakerRegistry circuitBreakerRegistry;
com.codahale.metrics.MetricRegistry metricsRegistry;

CircuitBreaker breaker = circuitBreakerRegistry.ofDefaults("name");
//this factory method creates metrics set for specific limiter
com.codahale.metrics.MetricSet metrics = CircuitBreakerGauges.of(breaker); 
//now you can register all this metrics with one call
registry.registerAll(metrics);

// ResilinceMetrics.of can also take iterable
MetricSet metrics = CircuitBreakerGauges.of(circuitBreakerRegistry.getAllCircuitBreakers()); 
registry.registerAll(metrics);
```

Maybe you guys have other ideas?

drmaas:
LGTM

The above would just need to execute after all circuitbreakers/ratelimiters are created.

RobWin:
I think we habe two use cases. 

1) Use Dropwizard Metrics to publish the metrics of the CircuitBreaker, Retry, RateLimiter.
2) Create a Metrics decorator for Counter and Timer to measure the Response time of functions and the number of invocations

`Supplier<String> = Metrics.decorateSupplier(counter, () -> ...)`

RobWin:
I create a new metrics branch. I created an initial resilience4j-metrics module for Dropwizard metrics and moved the Stopwatch class into a new resilience4j-core module.
@storozhukBM @drmaas Would you like to review it?

RobWin:
@fromanator Do you want to go on with the task and add your code into the resilience4j-metrics module?

[...]

PR #92, opened 2017-04-04
Adds retry metrics for #72

fromanator:
I also refactored `id` to `name` to match the conventions set for the rateLimiter and circuitBreaker classes. 

It looks like there's not a registry for AsyncRetry, and it looks like this feature isn't quite complete? So I didn't bother setting up metrics for it. Let me know if you also want this in a separate PR.


RobWin:
Thx. Why do you think AsyncRetry is not complete?

PR #93, opened 2017-04-05
adding retry method annotation, closes #91

### LINKED ISSUE #91: Ratpack support for retry method interceptor
Just like `@CircuitBreaker` and `@RateLimiter`.

drmaas:
also:

* adds methods to retry registry to ensure concurrent isolation where needed

### DIFF
--- a/resilience4j-ratpack/README.adoc
+++ b/resilience4j-ratpack/README.adoc
@@ -58,8 +58,11 @@ public Promise<String> methodWhichReturnsAPromise() {
 
 == Guice AOP
 
-Guice provides method interception capabilities. Here are provided some annotations which allow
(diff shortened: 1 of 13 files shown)


drmaas:
@RobWin let me know if the completionstage stuff should be refactored here or in a separate issue. Currently the tests pass, but they may be 'off by 1' if the circuitbreaker check happens after the completionstage has already executed.

RobWin:
I think we should disable CompletionStage interception and then merge the PR. We can find a solution in a separate issue..

drmaas:
See #95 for the remaing work to be done for CompletionStage.

drmaas:
@RobWin PTAL

PR #94, opened 2017-04-06
suggestion

aping-foo:
a suggestion for this queue. when the queue full,we need to wait?
or cover it?


RobWin:
Hi, thanks for your contribution.
But please create an issue and describe your problem before creating a Pull Requests.
We don't know your issue and what this change improves.
Please also don't use chinese comments or Git commit messages. The common language in this project is english.

Issue #97, opened 2017-04-09
Add support for rate limiting to the retrofit module


cpilsworth:
The initial functionality contained only circuit breaking, support for rate limiting should be added in a similar fashion.

RobWin:
@cpilsworth Do you want to join our Slack channel?

cpilsworth:
ok, sure

PR #98, opened 2017-04-09
Add rate limiting to the retrofit module

cpilsworth:
Adds support for rate limiting http client calls as described in #97 

* Added RateLimiterCallAdapter
* Created internal DecoratedCall to minimise duplication
* Added tests for Retrofit Call without type parameters for better coverage
* Updated docs and included rate limiting examples

### DIFF
--- a/resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/RetrofitRateLimiter.java
+++ b/resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/RetrofitRateLimiter.java
@@ -20,6 +20,7 @@ package io.github.resilience4j.retrofit;
 
 import io.github.resilience4j.circuitbreaker.CircuitBreaker;
 import io.github.resilience4j.ratelimiter.RateLimiter;
+import io.github.resilience4j.ratelimiter.RequestNotPermitted;
 import io.github.resilience4j.retrofit.internal.DecoratedCall;
 import javaslang.control.Try;
 import okhttp3.MediaType;
@@ -47,9 +48,9 @@ public interface RetrofitRateLimiter {
     /**
      * Decorate {@link Call}s allow {@link CircuitBreaker} functionality.
      *
-     * @param rateLimiter  {@link RateLimiter} to apply
-     * @param call            Call to decorate
-     * @param <T> Response type of call
+     * @param rateLimiter {@link RateLimiter} to apply
+     * @param call        Call to decorate
+     * @param <T>         Response type of call
      * @return Original Call decorated with CircuitBreaker
      */
     static <T> Call<T> decorateCall(final RateLimiter rateLimiter, final Call<T> call) {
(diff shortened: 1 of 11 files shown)


resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/RetrofitRateLimiter.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param rateLimiter  {@link RateLimiter} to apply
+     * @param call            Call to decorate
+     * @param <T> Response type of call
+     * @return Original Call decorated with CircuitBreaker
+     */
+    static <T> Call<T> decorateCall(final RateLimiter rateLimiter, final Call<T> call) {
+        return new DecoratedCall<T>(call) {
+            @Override
+            public Response<T> execute() throws IOException {
+                Try.CheckedSupplier<Response<T>> restrictedSupplier = RateLimiter.decorateCheckedSupplier(rateLimiter, call::execute);
```
cpilsworth: Is this an idiomatic way of using Javaslang's Try @RobWin?

RobWin: Can you use Try.of(restrictedSupplier).recover(ex -> tooManyRequestsError()).get() ?

RobWin:
Good work 👍

resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/RetrofitRateLimiter.java
cpilsworth: I have updated the code to this now:
```
public Response<T> execute() throws IOException {
                final Try.CheckedSupplier<Response<T>> restrictedSupplier = RateLimiter.decorateCheckedSupplier(rateLimiter, call::execute);
                return Try.of(restrictedSupplier)
                        .recover(ex -> Match(ex).of(
                                Case(instanceOf(RequestNotPermitted.class), tooManyRequestsError()),
                                Case(instanceOf(IllegalStateException.class), tooManyRequestsError()))
                        ).get();
            }
```
However, Try seems to want to always swallow the exceptions and return some Response<T>.  I really only want to consume the `RequestNotPermitted` and perhaps `IllegalStateException` exceptions so as not to change the interface for `Call` in decorating.  

The code above will give a MatchError on the a IOException from Call.execute - I would like to re-throw on the "else" match if possible.  Is there any way of doing this?

RobWin: Yes, Try might not be a good match for this type of use case. Might be simpler if you just use an imperative style.

cpilsworth:
Thanks for reviewing @RobWin :eyeglasses:

PR #99, opened 2017-04-11
handle completionstage appropriately from ratpack method annotations

### LINKED ISSUE #95: Fix CompletionStage handling support for Ratpack annotations
* Retry
* CircuitBreaker
* RateLimiter

For each, since CompletionStage is not lazy, once `invocation.proceed()` is called, the future will execute, whether it is desired or not.

A better approach will be to switch on `invocation.getMethod().getReturnType()`, and then decide how to proceed for each return type. 

Specifically for `CompletionStage`, it will be useful to only invoke the method if it is allowed to be invoked, based on the appropriate `Retry`, `CircuitBreaker`, or `RateLimiter` specified.

The other types can retain their logic as-is.

drmaas:
closes #95


drmaas:
@RobWin PTAL

PR #114, opened 2017-04-26
simple bulkhead operator

gibffe:
We needed a bulkhead operator on top of a circuitbreaker in our project, I figured I would open a pull, maybe others can find it usefull.


storozhukBM:
Hi @gibffe,
Thank you for your contribution. I think it will be really useful and I'll be able to use this in my projects too.
After some brief review it looks very good. I'm going to make in depth review tomorrow (EEST).

For now please fix all Codacy issues and try increase coverage reporting at least a little bit so we can meet our goal of 90%. Also please pay your attention on our `resilience4j-documentation` module. It'd be great if you provide some brief explanation and usage examples (or maybe diagram) for your Bulkhead pattern implementation.

Can't make any guaranties that this PR will be included to `0.9.0` version, because we've already planned it for this Friday (28 Apr). But maybe it will be included into `0.9.1`.

gibffe:
@storozhukBM I Will take care of the failed coverage & quality warnings.

RobWin:
Awesome. Thx.
Could you please add some Tests.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
```diff
@@ -0,0 +1,109 @@
...
+
+        // init event publisher
+        this.eventPublisher = PublishProcessor.<BulkheadEvent>create()
+                                              .toSerialized();
+    }
+
+    @Override
+    public boolean isCallPermitted() {
+
+        final boolean callPermitted = semaphore.tryAcquire();
```
RobWin: Do you think it makes sense to make the waiting time configurable?
Please have a look at our [SemaphoreBasedRateLimiter](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java#L118).

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
```diff
@@ -0,0 +1,109 @@
...
+    private final Semaphore semaphore;
+    private final FlowableProcessor<BulkheadEvent> eventPublisher;
+
+    /**
+     * Creates a bulkhead.
+     *
+     * @param name  the name of the bulkhead
+     * @param depth the depth of the bulkhead
+     */
+    public SemaphoreBulkhead(String name, int depth) {
```
RobWin: All of our components are configured using a Config class: `RateLimiterConfig`, `CircuitBreakerConfig`.
I think we should stay consistent here, even if we can only configure the semaphore depth at the momemt. But we could also add the waiting time as a config value.  See [RateLimiterConfig](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java#L30)

RobWin:
Do you think we should also provide a Thread-Pool based BulkHead or don't provide any bulkhead with a particular scheduler model?
Because a Thread-Pool based BuldHead would not work if you use Vert.x, for example.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
RobWin: I think we should rename depth to `maxConcurrentCalls`.  The name abstracts from the implementation.

gibffe: That makes sense - I'll create a config & rename the depth stuff.

The config will have maxConcurrentCalls, maxWaitTime.  Wait time not always makes sense, say for anything executing on an event loop, anything above 0 would probably be a terrible idea. I will add some generic note in the javadoc.

gibffe:
@RobWin Current, semaphore based bulkhead is a lowest common denominator that I felt could be useful across a wide spectrum of concurrency and io models. 

It works out of the box for event-loop/non-blocking scenarios. For traditional applications with thread pools and blocking io this can also do well, but I guess that's not necessarily obvious at first. Hystrix approach is shadow thread pools, which is fine. Here, the "firewall" limiting the saturation of your pool is virtual, but it is still there. It lacks of course any metrics on shadow pools, etc, which could be really useful.

I think it is a good idea to do further research into viability of thread-pool based bulkhead, possibly outside of this pull, as this may be quite an endeavor.

RobWin:
Yes, I totally agree.

RobWin:
@gibffe Do you think you can incorporate the changes within this week?

gibffe:
@RobWin Yes, I will be pushing updates later today.

PR #131, opened 2017-05-11
Histogram decorators for Prometheus

goldobin:


RobWin:
Please pull the latest changes from master.
The Dropwizard-Module Metrics interface does not exists anymore.

RobWin:
Could you please create a `Timer` Interface and a `TimerContext` class which can be instantiated similar to the other modules like `Timer.ofDefaults("label")` (which uses `CollectorRegistry.defaultRegistry`) or `Timer.ofCollectorRegistry("label", collectorRegistry)` (which uses a provided CollectorRegistry).
Please have a look at the resilience4j-metrics module.
The Timer should measure the number of total calls, successful calls and failed calls (Counters) and response time of successful calls (Histogram). I think the response time of failed calls is not necessary. What do you think?

resilience4j-prometheus/src/main/java/io/github/resilience4j/prometheus/Histograms.java
```diff
@@ -0,0 +1,273 @@
...
+ * distinguishing successful and failed executions correspondingly.
+ */
+public interface Histograms {
+    /**
+     * Decorates and executes the decorated Runnable.
+     *
+     * @param histogram the histogram to use
+     * @param runnable the original Callable
+     */
+    static void executeRunnable(Histogram histogram, Runnable runnable) throws Exception {
```
RobWin: When we have a TimerContext class, you can change this from `static` to `default`  and remove the Histogram parameter.

RobWin:
Any progress ? :)

goldobin:
Had very dense May with a lot of travels. Next month seems to be calm and I will finish it. :)

goldobin:
Hi @RobWin. Please take a look at the updated code. I have used different naming (CallMeter instead of Timer and Timer instead of Context) to fit into the Prometheus client library style of creating collectors.

RobWin:
Hi, thx.
Could you please add some documentation?

goldobin:
Hi, yes sure

goldobin:
Hi @RobWin. Please check updated PR.

RobWin:
I think I would prefer a different API which is more similar to resilience4j-metrics.
Instead of:
```
final CallMeter meter = CallMeter
  .build()
  .name("foo_call")
  .help("Foo call help")
  .labelNames("label_1")
  .register(registry);
```

Either: 
```
// Uses `CollectorRegistry.defaultRegistry` by default
final CallMeter meter = CallMeter
  .of("foo_call")
  .help("Foo call help")
  .labelNames("label_1")
  .build();
```

```
final CallMeter meter = CallMeter
  .ofCollectorRegistry("foo_call", registry)
  .help("Foo call help")
  .labelNames("label_1")
  .build();
```

What do you think?

goldobin:
As previously as far as this `CallMeter` is a kind of custom Prometheus meter I was trying to mimic Prometheus simple client's API for meters. But probably better to follow this library API design approach. So I agree `CallMeter.build().name("foo_call"). ... ` looks weird in the context of this library.  

One thing that I'm proposing to leave is an explicit call to the `register()` method like this:

```
final CallMeter meter = CallMeter
  .of("foo_call")
  .help("Foo call help")
  .labelNames("label_1")
  .build()
  .register(registry) // or without registry argument to use default registry 
```
-- so we're separating metric creation and registration in registry. For me it looks more convenient as far as Dropwizard and Prometheus clients has little bit different ideology.

What do you think?

RobWin:
I prefer your upper (first) version.
Would you still provide a `CallMeter.ofCollectorRegistry("foo_call", registry)`?

goldobin:
One more comment. Unlike Timer, CB and RateLimiter this meter creation is not one step. It has 5 parameters. So probably follow builder approach is better in this case rather than `of()` method. And actually, I can support both styles here.

RobWin:
Ok, then let's use the builder approach.
Your second example.

goldobin:
>> Would you still provide a CallMeter.ofCollectorRegistry("foo_call", registry)?

I think we can add this method as well as an alias for 
```
CallMeter
  .builder()
  .name("foo_call")
  .help("foo_call meter") // Help is mandadtory for Prometheus
  .build()
  .register(registry)
```

RobWin:
ok

goldobin:
Hey, @RobWin! Please take a look at updated PR

Issue #135, opened 2017-05-12
Support decoration/execution of CompletionStage in Ratelimiter


RobWin:
Please see CircuitBreaker or Bulkhead implementation.

storozhukBM:
Merged

PR #136, opened 2017-05-12
Issue #135 CompletionStage support in RateLimiter

storozhukBM:


RobWin:
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java#L305

Issue #141, opened 2017-05-19
Allow explicit ordering for CircuitBreaker and RateLimiter aspects in SpringBoot starter


storozhukBM:

RobWin:
Any idea how to solve it?

storozhukBM:
We have 2 options here:
1) We can strictly define aspects order using **org.springframework.core.annotation.Order** annotation. 
2) Or we can make it user definable via additional custom annotation.

Personally I'd use  strictly defined order where RateLimiter should be applied first and CircuitBreaker second.
This will lead to the following sequence:
`ProtectedOperation -> CircuitBreaker -> RateLimiter -> RealOperation`

Such sequence has few properties:
1) Frequent rate excesses with timeouts should move CB to open state.
2) Calls through the opened CB won't drain calling rates.

storozhukBM:
@RobWin any thoughts on it?

RobWin:
The Aspects could implement the `Ordered` and we could inject `CircuitBreakerProperties` or `RateLimiterProperties` and make the order configurable?

storozhukBM:
Merged to master. Will be included to the next release.

PR #165, opened 2017-10-02
#141 Allow explicit ordering for CircuitBreaker and RateLimiter aspects in SpringBoot starter

storozhukBM:
.


RobWin:
looks good. a test would be nice :)

PR #169, opened 2017-10-16
Adding Bulkhead integration with Ratpack and Bulkhead call finished metric

jansyk13:
* adding integration of Bulkhead with Ratpack via annotation(AOP/method interceptor)
* adding bulkhead finish event (to monitor whether the bulkhead is leaking permits or not)

I will add doc tomorrow. 

Notes:
* I am not sure about calling recovery function on both failed and rejected executions in transformer and method interceptor. IMO only rejected call should be handled by recovery function.
* I am not pleased with the tests I wrote... - not much support for asynchronous testing by Ratpack(both ExecHarness and TestingHttpClient have blocking methods) - working around it with executor and latches (to hit the bulkhead limit)

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
@@ -95,11 +95,8 @@ public class SemaphoreBulkhead implements Bulkhead {
 
     @Override
     public void onComplete() {
-        try {
-            publishBulkheadEvent(() -> new BulkheadOnCallFinishedEvent(name));
-        } finally {
-            semaphore.release();
-        }
+        semaphore.release();
+        publishBulkheadEvent(() -> new BulkheadOnCallFinishedEvent(name));
     }
 
     @Override
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
@@ -22,7 +22,7 @@ package io.github.resilience4j.bulkhead.internal;
 import io.github.resilience4j.bulkhead.Bulkhead;
 import io.github.resilience4j.bulkhead.BulkheadConfig;
 import io.github.resilience4j.bulkhead.event.BulkheadEvent;
-import io.github.resilience4j.bulkhead.event.BulkheadFinishedCallEvent;
+import io.github.resilience4j.bulkhead.event.BulkheadOnCallFinishedEvent;
 import io.github.resilience4j.bulkhead.event.BulkheadOnCallPermittedEvent;
 import io.github.resilience4j.bulkhead.event.BulkheadOnCallRejectedEvent;
 import io.github.resilience4j.core.EventConsumer;
@@ -96,7 +96,7 @@ public class SemaphoreBulkhead implements Bulkhead {
     @Override
     public void onComplete() {
         try {
-            publishBulkheadEvent(() -> new BulkheadFinishedCallEvent(name));
+            publishBulkheadEvent(() -> new BulkheadOnCallFinishedEvent(name));
         } finally {
             semaphore.release();
         }
@@ -134,6 +134,12 @@ public class SemaphoreBulkhead implements Bulkhead {
         @Override
         public EventPublisher onCallRejected(EventConsumer<BulkheadOnCallRejectedEvent> onCallRejectedEventConsumer) {
             registerConsumer(BulkheadOnCallRejectedEvent.class, onCallRejectedEventConsumer);
+            return this;
+        }
+
+        @Override
+        public EventPublisher onCallFinished(EventConsumer<BulkheadOnCallFinishedEvent> onCallFinishedEventConsumer) {
+            registerConsumer(BulkheadOnCallFinishedEvent.class, onCallFinishedEventConsumer);
             return this;
         }
 
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
@@ -22,6 +22,7 @@ package io.github.resilience4j.bulkhead.internal;
 import io.github.resilience4j.bulkhead.Bulkhead;
 import io.github.resilience4j.bulkhead.BulkheadConfig;
 import io.github.resilience4j.bulkhead.event.BulkheadEvent;
+import io.github.resilience4j.bulkhead.event.BulkheadFinishedCallEvent;
 import io.github.resilience4j.bulkhead.event.BulkheadOnCallPermittedEvent;
 import io.github.resilience4j.bulkhead.event.BulkheadOnCallRejectedEvent;
 import io.github.resilience4j.core.EventConsumer;
@@ -34,7 +35,7 @@ import java.util.function.Supplier;
 /**
  * A Bulkhead implementation based on a semaphore.
  */
-public class SemaphoreBulkhead implements Bulkhead{
+public class SemaphoreBulkhead implements Bulkhead {
 
     private final String name;
     private final Semaphore semaphore;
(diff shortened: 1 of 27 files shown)


jansyk13:
I have failing test when locally running clean build tasks even on master as well. 
```io.github.resilience4j.retry.AsyncRetryEventPublisherTest > shouldConsumeOnSuccessEvent FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at AsyncRetryEventPublisherTest.java:83
```
I will look at it tomorrow

RobWin:
@storozhukBM Do you have time to review it?

storozhukBM:
@jansyk13 Thank you for this PR
@RobWin Yes, I have time to review it. 
This evening or tomorrow morning. 
Stay tuned 😃

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+public class BulkheadMethodInterceptor implements MethodInterceptor {
+
+    @Inject(optional = true)
+    private BulkheadRegistry registry;
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Object invoke(MethodInvocation invocation) throws Throwable {
+        Bulkhead annotation = invocation.getMethod().getAnnotation(Bulkhead.class);
```
storozhukBM: I'm not particularly familiar with Ratpack ways to deal with AOP.
So maybe this comments is not relevant, but:
Using reflection on each invocation to determine annotations and recovery function is not very performant way.
There is two ways you can handle this:
1. Create Interceptor object for each intercepted method and specify `recoveryFunction` during interceptor creation.
2. Use `ConcurrentHashMap<Method, RecoveryFunction>` as cache for this heavy reflection operations.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
```diff
@@ -0,0 +1,133 @@
...
+        RecoveryFunction<?> recoveryFunction = annotation.recovery().newInstance();
+        if (registry == null) {
```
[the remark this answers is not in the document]
storozhukBM: Is there any real case where `registry` won't be injected before first `invoke` execution.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
```diff
@@ -0,0 +1,133 @@
...
+            registry = BulkheadRegistry.ofDefaults();
+        }
+        io.github.resilience4j.bulkhead.Bulkhead bulkhead = registry.bulkhead(annotation.name());
```
storozhukBM: one more reflection issue, please look at my comments above

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
```diff
@@ -0,0 +1,133 @@
...
+        if (bulkhead == null) {
+            return invocation.proceed();
+        }
+        Class<?> returnType = invocation.getMethod().getReturnType();
+        if (Promise.class.isAssignableFrom(returnType)) {
```
storozhukBM: ideally this decisions should be made during interceptor creation

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
storozhukBM: I know that currently all interceptor in ratpack module are implemented this way, but maybe we can find better solution.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
```diff
@@ -94,7 +95,11 @@ public boolean isCallPermitted() {
 
     @Override
     public void onComplete() {
-        semaphore.release();
+        try {
```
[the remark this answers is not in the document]
gibffe: Can event publishing be optional ? We are currently not consuming any bulkhead events, and have no immediate plans to, so this will immediately become just a gc pressure for us ...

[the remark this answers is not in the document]
storozhukBM: @gibffe  Actually, it is optional. Check out `publishBulkheadEvent` method it won't publish anything if there is no subscriptions.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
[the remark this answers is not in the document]
jansyk13: I kept the implementation same as other interceptors. But I don't think there are many cases, when this would occur. I think most use cases for these annotations are in with methods that make external calls(and requests are already being handled), which means that registry(context) is already created. The creation of context happens when Ratpack initializes Netty internals and adds handler adapter(contains all Ratpack handlers and registry) as last channel inbound handler. https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/server/internal/DefaultRatpackServer.java#L263

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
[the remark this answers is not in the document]
jansyk13: @storozhukBM agree

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
jansyk13: I will look into how Guice does AOP(see `Resilience4jModule` and usage of `bindInterceptor`). Probably some external library could be introduce to do "dirty work" with reflection, would that be okay? But generally I think the implementation should be consistent across all interceptors(bulkhead, retry, cb, ...)

jansyk13: @RobWin @storozhukBM is there a way to discuss stuff more directly? You mentioned Slack channel before.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
jansyk13: Probably this could be used https://github.com/google/guice/blob/master/core/src/com/google/inject/matcher/Matchers.java#L380 and create method interceptor for each case.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
RobWin: I need your E-Mail address to invite you.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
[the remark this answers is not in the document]
RobWin: I think it's fine if the implementation is similar to the other interceptors.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
RobWin: Let's make this refactoring independent from this PR.

storozhukBM:
Completely agree with @RobWin. Let's make some refactoring in scope of other issue.
As for now only tests fix required.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadMethodInterceptor.java
jansyk13: jansyk13 at gmail, thanks

RobWin:
Some test still failing?

PR #170, opened 2017-10-16
Removing usage of fork join pool in CircuitBreakerMethodInterceptor

jansyk13:
I have noticed that `CircuitBreakerMethodInterceptor` uses `whenCompleteAsync`. IMO it should be replaced with `whenComplete` to avoid using  fork-join pool.


RobWin:
Thx
@drmaas Would you like to review it?

storozhukBM:
Following callback seems to be very small and non-blocking, so in theory it should be OK to delegate it to common fork-join thread pool.

jansyk13:
I think all other types(Promise, Flowable, etc.) uses the same thread(one that fulfills the execution) to execute recovery function. So why it should be different?

Recovery function is a function defined by user, so it can be blocking(r4j can't control what the function does). Plus in my opinion it is a bad practice to use fork join pool from within the library, the user can't control the thread that executes it and it can have negative impacts.

RobWin:
I personally agree that `whenComplete` should be used.
But the code was implemented by @drmaas. I would like to understand his reason why he used `whenCompleteAsync`.

storozhukBM:
Yep, from this perspective I also agree.

PR #173, opened 2017-11-01
Fixing blinking tests

jansyk13:
There is a race condition in `AsyncRetryBlock` when executing `AsyncRetryEventPublisherTest.shouldConsumeOnSuccessEvent` test
```        
stage.whenComplete((result, t) -> {
    if (t != null) {
        onError(t);
    } else {
        promise.complete(result);
        retryContext.onSuccess();
    }
});
```
The callback is invoked after completing the promise on schedulers thread but main thread(test) already has result of promise. Currently callback is used only for publishing events, so doesn't really matter whether it gets executed immediately or not. But synchronization over latch in test makes sure that it gets called before asserting.


jansyk13:
There is one more blinking test `io.github.resilience4j.ratpack.ratelimiter.endpoint.RateLimiterChainSpec > test events`

jansyk13:
@RobWin Hey, could you look over these changes? Both issues are described in comments above. These changes should also fix my other PR. Thanks

RobWin:
Thx

Issue #174, opened 2017-11-17
Possible to create a dynamic rate limiter?


efenderbosch:
Some APIs have strict limits. However, we need to put a rate limit on API calls to servers of unknown CPU/RAM. We can probably write a plugin to get stats reported back to us, and scale the rate limit up/down, but then we need to modify the existing ``RateLimiter``. ``RateLimiterConfig`` doesn't have any mutator methods.

Could setters be added for ``timeoutDuration``,  ``limitRefreshPeriod`` and ``limitForPeriod``?

Alternatively, maybe something like a ``RateLimiterScalingStrategy`` and the default would be ``NoScalingStrategy`` that duplicates the current behavior. Then we could write a custom implementation?

storozhukBM:
Hi @efenderbosh,
Yes, it is possible. In fact I’ve already thought about few times, but haven’t found any real use case in my project, so decided to wait for some, before making any changes. Could you please describe your use case in details? I’m interested in such things: what params you want to change dynamically, only limits per cycle or limit refresh period also? How your program will make decision when to change params and what new values should be?
I’ll think about it more and will post some solution design here for discussion. When we agree on interfaces and other details, I’ll make required implementation changes.
If you have any other questions I’ll be happy to answer them. 
Have a nice day and don’t forget to give our library a star🙂

efenderbosch:
We've been using the old ``io.github.robwin`` ``CircuitBreaker`` and ``RateLimiter`` classes for a while, but we've always used them when connecting to an API that has specific limits, like SendGrid or Shopify.

We're creating new multi-tenant integrations that connect to APIs that don't have built in rate limits. We also won't necessarily know the size/capacity of the server that will be serving the API and don't want to bring our client's servers to their knees. We're thinking about running a small plugin/extension on their server to periodically report back stats like CPU utilization via http post. If CPU utilization gets too high, we'd scale back our ``RateLimiter``. As for scaling limits per cycle or refresh period, I'm open to suggestions.

Is that enough info?

storozhukBM:
@efenderbosch I think it should be enough. I'll think about possible solutions and will come back after few days.

storozhukBM:
Sorry closed by accident.

storozhukBM:
Hi @efenderbosch ,
Please checkout PR #177. I've added few methods for `RateLimiter` that should satisfy your use cases. If such API is OK for you, I'll add similar functionality to bulkhead component and it will be delivered in scope of the next release.

[...]

PR #177, opened 2017-11-26
#174 Dynamic ratelimiter configuration

storozhukBM:


storozhukBM:
Hi @RobWin 
I'm also concerned about it, but unfortunately I haven't found the way to make it safe from the API stand point. Due to the following: 
`rateLimiter.getRateLimiterConfig().updateTimeoutDuration(...)` - I think we want to keep config immutable for cases when it have been reused 
across different rate limiters, we should change behavior of only one of them.
`rateLimiter.updateRateLimiterConfig(RateLimiterConfig config)` - we should not allow to change limit refresh period, so we can't accept completely new config.

RobWin:
Ok.
`rateLimiter.getRateLimiterConfig().updateTimeoutDuration(...)` could create a new instance of the rateLimiterConfig internally.

RobWin:
But you mean I could be confusing, because people think you can modify an existing RateLimiterConfig?

storozhukBM:
rateLimiter.getRateLimiterConfig() should return configuration. like that
```java
RateLimiterConfig oldConfig = rateLimiter.getRateLimiterConfig();
```
Let's assume `updateTimeoutDuration(...)` method will produce new config without changing previous
```java
RateLimiterConfig newConfig = oldConfig.updateTimeoutDuration(...)
```
But then we should insert this new config to the rateLimiter which is actually your second variant
`rateLimiter.updateRateLimiterConfig(RateLimiterConfig config)` where we can't be sure that user changed only allowed options. 
And runtime checks for such disallowed changes aren't very convenient.

RobWin:
OK then lets keep your solution

storozhukBM:
Thank you

Issue #175, opened 2017-11-22
Add support for Maybe and Completable


danielimre:
RxJava2 support should include operators for `io.reactivex.Maybe` and `io.reactivex.Completable`
Now, the only workaround I found is converting them to `Observable` and use the appropriate operators.

RobWin:
Hi,

would you like to create a PR?

danielimre:
@RobWin , created a PR for bulkhead: #180 
Please check if I've got it right and will do the rest.

RobWin:
Merged

PR #181, opened 2017-12-05
Add RxJava bulkhead operator for Maybe and Completable.

danielimre:
Added bulkhead operator support for Maybe and Completable. Part of #175
Changed some of the observers to have stricter control on how to handle disposed state.
Also, split the BulkheadOperator itself and its tests into respective classes.


storozhukBM:
@danielimre please check out this Codacy report
https://www.codacy.com/app/storozhukBM/resilience4j/pullRequest?prid=1132768

RobWin:
I like the additions, but I think the other operators (CircuitBreaker and Ratelimiter) should also support Maybe and Completable so that they are functionally identical. Would it be possible?

danielimre:
@RobWin , I will make these changes for other operators as well. Just wanted to have a confirmation that the approach is ok. Writing RxJava operators is not trivial for me yet. Not sure if we really need that level of concurrency handling around cancellation and disposal. On the other hand, I don't think they would add too much overhead.

RobWin:
Yes RxJava operators are not trivial

PR #185, opened 2018-01-01
Feature/add more rx operator support

danielimre:
Add direct support for Maybe and Completable for CircuitBreaker RxJava operator.
Update documentation.


resilience4j-rxjava2/src/main/java/io/github/resilience4j/circuitbreaker/operator/CircuitBreakerSubscriber.java
```diff
@@ -0,0 +1,123 @@
...
+        markFailure(e);
+        if (isInvocationPermitted()) {
+            childSubscriber.onError(e);
+        }
+    }
+
+    @Override
+    public void onComplete() {
+        LOG.debug("onComplete");
+        markSuccess();
```
danielimre: I'm not 100% sure about this one but looked logical to me to mark the call as success when the subscription was cancelled after subscribing to it. For example some elements could have been already emitted in onNext but the call wasn't registered in CB. Unfortunately, it depends on upstream whether it honors cancellation or not. 
The same logic was applied in all other CB operators. 
What do you think about this?

danielimre:
Not sure why the CI build is failing..

storozhukBM:
Hi @danielimre,
Thank you for this contribution. I'll make review within next 2 days.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/circuitbreaker/operator/CircuitBreakerSubscriber.java
storozhukBM: For me it looks OK, but I think we should explicitly mention this behaviour in documentation.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java
```diff
@@ -15,7 +17,7 @@
     private Disposable disposable;
     private final Bulkhead bulkhead;
     private final AtomicBoolean disposed = new AtomicBoolean(false);
-    private final AtomicBoolean permitted = new AtomicBoolean(false);
+    private final AtomicReference<Permit> permitted = new AtomicReference<>(Permit.PENDING);
```
storozhukBM: @danielimre could you please describe reasons behind such solution.

RobWin: Yes, I would also like to understand it :)

danielimre: It's really an edge-case and maybe an overkill but `acquireCallPermit` wasn't safe from this point of view. One subscriber can try to acquire a permit and fail which results in `permitted=false` and the rest of the subscription should act according to this. But meanwhile another subscriber could try to acquire a permit again and succeed which can have weird side-effects on the first subscription.
On the other hand, some parts of the operator not really thread-safe anyway so it's a question whether we should ensure this level of protection at all.

RobWin: As far as I know all operators in RxJava which operate on one Observable are NOT thread safe.
Only operators which operate over multiple Observables are thread safe: merge(), combineLatest(), zip(), .....

RobWin:
We can merge the PR from my point of view. I think we can optimize the implementation when people use the operators and create issues.

RobWin:
@akarnokd Would you have time and passion to review the RxJava2 operators from Resilience4j?

akarnokd:
Sure.

RobWin:
Awesome. You could use this link. It contains the latest changes: https://github.com/danielimre/resilience4j/tree/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java
danielimre: So what should be done? Simplify it according to this?

akarnokd:
### RxJava2Adapter

- [L15](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/adapter/RxJava2Adapter.java#L15) it is recommended you document the backpressure behavior, especially when using a `Processor` that doesn't compose backpressure. Also disconnecting from the onEvent is not supported either.
- [L22](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/adapter/RxJava2Adapter.java#L22) can the `EventPublisher`'s callbacks happen concurrently?
- [L36](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/adapter/RxJava2Adapter.java#L36) Is `toSerialized` necessary?

### DisposableBulkhead

- [L20](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java#L20) You coud combine `disposed` and `permitted`into a a single atomic reference with a 4 state enum: `pending -> ((acquired -> released) | (rejected))`.

### DisposableCircuitBreaker

- [L22](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/circuitbreaker/operator/DisposableCircuitBreaker.java#L22). Same as with the bulkhead, combine into one state.
- [L31](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/circuitbreaker/operator/DisposableCircuitBreaker.java#L31) Stop the `StopWatch`?

### RateLimitOperator

- [L100](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/ratelimiter/operator/RateLimiterOperator.java#L100) Perhaps set `cancelled` to `true` here so an eager `onError` and `onComplete` may not get forwarded.
- [L116](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/ratelimiter/operator/RateLimiterOperator.java#L116). Same case, `cancelled.set(true);` here.
- [L158](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/ratelimiter/operator/RateLimiterOperator.java#L158) Use `compareAndSet` here (suggested) or `cancelled` could be a plain volatile field.

The remaining two operators have similar points.

### RetryTransformer

- [L61](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java#L61) No need for `fromPublisher`, just extend `Flowable` and implement `subscribeActual`.
- [L71](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java#L71): `new Observable<T>() { /* ... */ }`
- [L72](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java#L72) Why convert to `Flowable` and introduce a buffer?
- [L82](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java#L82): `new Single<T>() { /* ... */ }`
- [L83](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java#L83) Why convert to `Flowable` and introduce an indirection?
- [L123](https://github.com/danielimre/resilience4j/blob/5be2a73e3f6a2757fa0dbddf11025c792f964df2/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java#L123): This approach is a bit expensive. You could collect the [onNext](https://github.com/akarnokd/Reactive4JavaFlow/blob/master/src/main/java/hu/akarnokd/reactive4javaflow/impl/operators/FolyamRetry.java#L133) count [into a long](https://github.com/akarnokd/Reactive4JavaFlow/blob/master/src/main/java/hu/akarnokd/reactive4javaflow/impl/operators/FolyamRetry.java#L66), commit it [once per repetition](https://github.com/akarnokd/Reactive4JavaFlow/blob/master/src/main/java/hu/akarnokd/reactive4javaflow/impl/operators/FolyamRetry.java#L111)


---

The rest of the files seem okay to me.

RobWin:
Thank you

danielimre:
@akarnokd , I have applied some changes on `DisposableBulkhead`. Could you please confirm that those are ok and I will apply them on the circuit breaker as well. Thanks.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java
```diff
@@ -25,22 +24,53 @@
...
 
     @Override
     public boolean isDisposed() {
-        return disposed.get();
+        return DisposableHelper.isDisposed(get());
     }
 
-    protected void setDisposable(Disposable disposable) {
-        this.disposable = requireNonNull(disposable);
+    protected void onSubscribe(Disposable disposable, Consumer<Disposable> onSubscribe, Consumer<Throwable> onError) {
```
akarnokd: I would have implemented abstract methods instead so no allocation/lambdas have to happen.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java
```diff
@@ -25,22 +24,53 @@
...
+        DisposableHelper.setOnce(this, disposable);
```
akarnokd: You only need this if cancellation can happen before or during `onSubscribe`.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java
danielimre: I refactored that part a bit avoiding lambdas. Thanks.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/DisposableBulkhead.java
danielimre: I'm not sure if that can be ensured here. We can keep it just to be safe.

danielimre:
Can someone please kick of another CI build as it failed with some env issue.. Not sure how I can rebuild it.

RobWin:
Done

RobWin:
Please tell us when we can do the next review

danielimre:
I consider this PR complete from my side. Please go ahead with the review. Thanks.

RobWin:
Awesome work. Thank your for your effort.
For me it looks fine.

RobWin:
@storozhukBM Do you have any findings?

storozhukBM:
@RobWin 
no LGTM

RobWin:
@danielimre Thanks for your effort.

RobWin:
We would be happy, if you would like to contribute more features.

danielimre:
Thanks @RobWin , will do my best to contribute more.

PR #192, opened 2018-01-11
Retry Budget implementation

jansyk13:


RobWin:
Thank you very much for your PR.
Some findings I would like to talk about. I don't have a solution right now for all of them:

- RetryMetrics interface change is not backward compatible
- Tests use Spock framework instead of Junit, Mockito und Assertj
- We might need more documentation about the implementation details

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java
```diff
@@ -0,0 +1,208 @@
...
+    private RingBitSet ringBitSet;
+
+    public RetryBudgetImpl(String name, RetryConfig config) {
+        this.name = name;
+        this.config = config;
+        this.maxAttempts = config.getMaxAttempts();
+        this.intervalFunction = config.getIntervalFunction();
+        this.exceptionPredicate = config.getExceptionPredicate();
+        this.bufferSize = config.getBufferSize();
+        this._bufferSize = config.getBufferSize();
```
RobWin: Let's find a better name for this variable

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        private void waitIntervalAfterFailure() {
+            // wait interval until the next attempt should start
+            long interval = intervalFunction.apply(numOfAttempts.get());
+            Try.run(() -> sleepFunction.accept(interval))
+                    .getOrElseThrow(ex -> lastRuntimeException.get());
+        }
+
+        private boolean canRetry(final int currentNumOfAttempts) {
```
RobWin: Could you please add some JavaDocs for this `canRetry` method

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java
jansyk13: I will keep only `bufferSize` and internally it will be double

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java
jansyk13: I will add

jansyk13:
@RobWin I can rewrite the test to JUnit. It was just more convenient for me to use Spock. Or I can rewrite other JUnit tests to Spock.

gibffe:
rx operator could be handly (if makes sense)

jansyk13:
@gibffe It implements `Retry` interface so it should work out of the box. There already is `RetryTransformer` for rx

gibffe:
:+1: I will play around with it.

jansyk13:
Fixed:
- [x] Return Metrics to Retry - backward compatibility
- [x] Rewrite tests from Spock to JUnit
- [x] More documentation and explanation about how budget works

cc @RobWin

RobWin:
Awesome. Thx.
I hope I can review it tomorrow.
@storozhukBM Did you have some time to look at it?

storozhukBM:
Hi
Sorry I wasn't able to look at it.

jansyk13:
Cool, when building on top of this. I found out that the `reset` method for the `RetryBudgetImpl` could be helpful, because it is stateful.

jansyk13:
ping @RobWin @storozhukBM

RobWin:
Sry I'm currently reviewing another PR.

storozhukBM:
Hi @jansyk13,
Sorry for this delay with review, I'll do my best finish it by the end of this week.
I hope it's not critical for you, but if it is, please tell me and I'll raise its priority in my TODO list.

jansyk13:
@storozhukBM you can take your time, I will release "alpha version" of our fork into private repository and use that in mean time. I already built something on top of this and later I will switch to upstream version.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -52,6 +58,18 @@ public int getMaxAttempts() {
         return exceptionPredicate;
     }
 
+    public CheckedConsumer<Long> getSleepFunction() {
+        return sleepFunction;
+    }
+
+    public int getBufferSize() {
```
RobWin: I think I would prefer that we have to different Config classes for two different retry implementations. From the user perspective I might be overwhelmed by the configuration options even if they don't apply to the concrete Retry implementation. 
My suggestion: We should introduce a new `BudgetedRetryConfig` and a new `BudgetedRetry` interface and `BudgetedRetryImpl`. (Seperation of Concerns).

```` 
BudgetedRetryConfig config = BudgetedRetryConfig.custom()
                 .bufferSize(1000)
                 .retryThreshold(0.3)
                 .maxAttempts(3)
                 .build()
BudgetedRetry retry = BudgetedRetry.of("testName, config);
````

Comment (jansyk13) [reply resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java]: I get the reason, but I think it is better to keep `Retry` interface and benefit from it. Retry Budget(RetryBudgetImpl) is just another implementation. Better documentation of `RetryConfig`?

Comment (RobWin) [reply resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java]: Since we have multiple implementations of Retry now, I don't like to static factory method `Retry.of()` in the interface anymore. How do you control which implementation do you want to use. Either by using different Config classes or we need a `RetryFactory`.
A better place for the static factory methods would have been the implementation class itself. :(

Comment (RobWin) [reply resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java]: @storozhukBM What do you think?

Comment (storozhukBM) [reply resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java]: Yep, we have similar issue with two implementations of RateLimiter.

Comment (jansyk13) [reply resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java]: Ok, but this won't be backward compatible

Comment (RobWin) [reply resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java]: Yes we should first discuss pros and cons.

File: resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java [hunk 4]
```diff
@@ -0,0 +1,255 @@
...
+            // check max number of attempts
+            if (currentNumOfAttempts >= maxAttempts) {
+                // mark next as '0'
+                ringBitSet.setNextBit(false);
+                return false;
+            }
+            synchronized (this) {
+                // retrieve number of retries in bit set, add one and divided by buffer size
+                double retriesPctWithOneMore = (ringBitSet.cardinality() + 1) / bufferSize;
```
storozhukBM: I think you should call `ringBitSet.setNextBit` only from `onSuccess` and `onError` methods. And `canRetry` should only check threshold without any additional synchronisations. Perhaps, I'm missing something because of trying to read completely from GitHub, so correct me if I'm wrong.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java
```diff
@@ -0,0 +1,255 @@
...
+                // check against the threshold
+                if (retriesPctWithOneMore >= retryThreshold) {
```
storozhukBM: And also please rename `retriesPctWithOneMore` to `resultRetriesPercent` or something similar

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -52,20 +48,21 @@
...
         this.maxAttempts = config.getMaxAttempts();
         this.intervalFunction = config.getIntervalFunction();
         this.exceptionPredicate = config.getExceptionPredicate();
-        this.metrics = this.new RetryMetrics();
         this.eventProcessor = new RetryEventProcessor();
-        succeededAfterRetryCounter = new LongAdder();
-        failedAfterRetryCounter = new LongAdder();
-        succeededWithoutRetryCounter = new LongAdder();
-        failedWithoutRetryCounter = new LongAdder();
+        this.sleepFunction = config.getSleepFunction();
```
storozhukBM: I think we should actually embed all this fields to `RetryMetricsImpl` completely and provide methods to update them like `RetryMetricsImpl.failedWithoutRetry()` etc.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryBudgetImpl.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        /**
+         * Resolves whether the failed operation can retry.
+         *
+         * @param currentNumOfAttempts current number of attempts by this operation
+         * @return boolean whether the operation can retry
+         */
+        private boolean canRetry(final int currentNumOfAttempts) {
+            // check max number of attempts
```
storozhukBM: Both me and @RobWin separately noticed possible issue with this design created on top of BitBuffer.
The issue is that after `retriesPctWithOneMore >= retryThreshold` actually happens, retry will be blocked and you will you actually need one call to succeed without retries to recover from this state.

Is it your actual design choice? 
For examle Finagle actually has additional time constraints for this behaviour. https://finagle.github.io/blog/2016/02/08/retry-budgets/

That is why I was suggesting you to look through AtomicRateLimiter token bucket implementation.
For now it seems to me that correct design should be some combination of buckets with BitBuffer.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -95,25 +113,64 @@ public Builder waitDuration(Duration waitDuration) {
...
         }
 
+        /**
+         * Set a function that performs a sleep/pause between retries after failure. By default
+         * the {@link Thread#sleep(long)} is used.
+         *
+         * @param sleepFunction function that performs a sleep/pause between retries after failure
+         * @return the RetryConfig.Builder
+         */
+        public Builder sleepFunction(CheckedConsumer<Long> sleepFunction) {
```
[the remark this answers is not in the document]
storozhukBM: Agree

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
storozhukBM: I'm not sure that IOC principle is actually applies here.
From my perspective all this counters don't bring any behaviour to this Retry implementation.
Actually you can completely delete remove them and nothing should actually change.

That is why I think that this counters don't belong here and should be isolated behind some other class.

michael-barker:
@jansyk13 I have a pretty good use case where I'd like to use retry budgets.  Do you plan on continuing the work on this PR for the suggested changes?

storozhukBM:
Unfortunately this PR is in work too long. I think we can close it due too inactivity. @jansyk13 if you still want to improve it and push to upstream, you can reopen or recreate separate PR.

PR #194, opened 2018-01-16
Feature/159 allow circuit breaker disabled forced open

rLitto:
Missing the tests


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
```diff
@@ -279,8 +279,20 @@ public EventPublisher onStateTransition(EventConsumer<CircuitBreakerOnStateTrans
         }
 
         @Override
-        public EventPublisher onStateReset(EventConsumer<CircuitBreakerOnResetEvent> onStateResetEventConsumer) {
-            registerConsumer(CircuitBreakerOnResetEvent.class, onStateResetEventConsumer);
+        public EventPublisher onReset(EventConsumer<CircuitBreakerOnResetEvent> onResetEventConsumer) {
```
RobWin: I think we don't need special event handlers like `onReset`, `onDisable` and `onForceOpen`. I would suggest to use `onStateTransition` for all state transitions.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
```diff
@@ -160,48 +161,47 @@ public String toString() {
...
-        if (previousState.getState() != CLOSED) {
-            publishStateTransitionEvent(StateTransition.transitionToClosedState(previousState.getState()));
+        if (previousState.getState() != newState) {
+            publishStateTransitionEvent(StateTransition.transitionBetween(previousState.getState(), newState));
         }
     }
 
+
+    @Override
+    public void disable() {
```
RobWin: I would prefer `transitionToDisabled` over `disable()` and `transitionToForcedOpen` over `forceOpen`. Because it's more consistent to the other methods.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
rLitto: only for `onReset`

RobWin:
Good work so far :+1: 
Please don't forget to adapt our AsciiDoc-documentation. I can add an updated image of the circuitbreaker state diagram.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -229,13 +258,46 @@ public int getOrder(){
...
+                        .collect(Collectors.toMap(v -> new Tuple2(v.fromState, v.toState), Function.identity()));
+
+        private boolean matches(State fromState, State toState) {
+            return this.fromState == fromState && this.toState == toState;
+        }
 
-        State fromState;
-        State toState;
+        public static StateTransition transitionBetween(State fromState, State toState){
+            final StateTransition stateTransition = STATE_TRANSITION_MAP.get(new Tuple2<>(fromState, toState));
```
RobWin: Let's use `Tuple.of()`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -229,13 +258,46 @@ public int getOrder(){
...
+        DISABLED_TO_HALF_OPEN(State.DISABLED, State.HALF_OPEN);
+
+        private final State fromState;
+
+        private final State toState;
+
+        private static final Map<Tuple2<State, State>, StateTransition> STATE_TRANSITION_MAP =
+                Arrays
+                        .stream(StateTransition.values())
+                        .collect(Collectors.toMap(v -> new Tuple2(v.fromState, v.toState), Function.identity()));
```
RobWin: Let's use `Tuple.of()`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -298,12 +329,13 @@ public String toString(){
...
         EventPublisher onStateTransition(EventConsumer<CircuitBreakerOnStateTransitionEvent> eventConsumer);
 
-        EventPublisher onStateReset(EventConsumer<CircuitBreakerOnResetEvent> eventConsumer);
+        EventPublisher onReset(EventConsumer<CircuitBreakerOnResetEvent> eventConsumer);
 
         EventPublisher onIgnoredError(EventConsumer<CircuitBreakerOnIgnoredErrorEvent> eventConsumer);
 
         EventPublisher onCallNotPermitted(EventConsumer<CircuitBreakerOnCallNotPermittedEvent> eventConsumer);
 
+        EventPublisher onAny(EventConsumer<CircuitBreakerEvent> onCallNotPermittedEventConsumer);
```
RobWin: `onAny` is not required. There is already an `onEvent` in the `EventPublisher` interface.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/event/CircuitBreakerEvent.java
```diff
@@ -51,16 +51,26 @@
...
-        RESET;
+        RESET(true),
+        /** A CircuitBreakerEvent which informs the CircuitBreaker has been forced open */
+        FORCED_OPEN(false),
+        /** A CircuitBreakerEvent which informs the CircuitBreaker has been disabled */
+        DISABLED(false);
+
+        public final boolean forcePublish;
+
+        Type(boolean ignorable) {
```
RobWin: Please rename ignorable to forcePublish.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/ForcedOpenState.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Returns false, if the wait duration has not elapsed.
+     * Returns true, if the wait duration has elapsed and transitions the state machine to HALF_OPEN state.
+     *
+     * @return false, if the wait duration has not elapsed. true, if the wait duration has elapsed.
+     */
+    @Override
+    boolean isCallPermitted() {
+        return false;
```
RobWin: Do you think we should measure metrics here? It could be interesting how many calls have been rejected.
See OpenState
`circuitBreakerMetrics.onCallNotPermitted();`

rLitto: Do we? What is the usual behaviour in other CircuitBreakers? But then why not measure how many we allowed when we are `DISABLED`? Should then have a difference between `DISABLED` and `FORCED_CLOSED`?

rLitto: Shouldn't be better if `FORCED_CLOSED` was a real state, like `FORCED_OPEN` where both record metrics but do not generate events, while being disabled/enabled simply switches off/on the CB but does not change state or update metrics? So the cb is disabled -> it is like it is bypassed and no call arrives to it, but keeps the same state and metrics as before and resumes from that point when it is enabled?

RobWin: I think `DISABLED` should be a state where everything (metrics and events) is disabled. When I disable a CB I don't care about metrics anymore. It would be even confusing when metrics still change in my monitoring system. Since there is no `ON` state, the user can choose to transition either to `OPEN`, `FORCED_OPEN`, `HALF_OPEN` or  `CLOSED`. Therefore I don't think that it makes sense to keep the metrics of the previous state.
The state `FORCED_OPEN` is somehow similar to `OPEN` but does not change back to `HALF_OPEN`. But I would expect to see metrics (number of rejected calls) in my monitoring system.

rLitto: K, I will implement it in the weekend

PR #198, opened 2018-01-21
Add Maybe and Completable RateLimiter operator support.

danielimre:
Refactor RateLimiter operators to be in line with other operators.
Add direct support for Completable and Maybe RateLimiter.
Avoid superfluous permission request in Observable and Flowable onNext for first event.


storozhukBM:
Hi @danielimre,
Thank you very much for this contribution and effort you've made.
I've looked through a changes and everything looks fine for me, but unfortunately I'm not such a heavy RxJava user, so I think we should wait for additional review from @RobWin 

PS: I love this 92% of coverage 😍 good job.

danielimre:
Hi @RobWin , when is this planned to be released?

RobWin:
Hi @danielimre 
Good question. :)
I could create a new release soon.

PR #202, opened 2018-02-06
Making a small change that would allow getting the builder, and thus …

rogerschlachter:
Making a small change that would allow getting the builder, and thus allow adding a custom exception handler.


RobWin:
thx

PR #227, opened 2018-04-12
ratpack will now use reactor

drmaas:


RobWin:
Hi thx.
What's the main reason to switch to Reactor?

drmaas:
I think it was for a couple reasons - 

1. consistency with resilience4j Spring modules
2. Recommendations by David Karnok (RxJava2 lead and Reactor contributor) that Reactor be used in java8 applications. See https://www.reddit.com/r/java/comments/6acxe3/spring_reactor_vs_rxjava/ for discussions

Issue #228, opened 2018-04-13
[resilience4j-retry] Events not published on every failed try


olayinkasf:
I think it would be better if an event is published after every failed try. Why?

In a scenario where more than one type of exception is thrown, I would like to know where this exceptions occur so that I can at least attempt to fix them. Right now, event is only published at the end of the retry and only the last exception is accessible.

RobWin:
Hi,

we made the design decision so that the monitoring system is not overwhelmed by too many retry events if a remote system is down. If you want to log all exceptions, you could catch the exception of your service, log it and rethrow the exception.
Would this be sufficient for you?

olayinkasf:
Catching the exception and rethrowing seems reasonable however I doubt too many retry events would be overwhelming to the publisher. The publisher can still choose to not register for these events therefore avoiding the load. I personally would prefer to not write a try catch throw every time I use the retry when I can simple register for the retry events.

olayinkasf:
A list of exceptions might consume too much memory in case of having too many retries but publishing an event on every retry could replace the need for a list of events and can help monitor (while in progress) how well a resilience call is performing. I'd be glad to create a PR @RobWin

RobWin:
Feel free to create a PR. We are an open community.
But we should keep in mind that the change might not be backward compatible.
Maybe we need a config switch.

RobWin:
v0.13.0 has been released

PR #229, opened 2018-04-22
Publish retry event for every retry

olayinkasf:
#228


RobWin:
Thx. Looks good. 
I'm thinking about renaming `RetryOnRetryEvent `.
Any idea? :)

olayinkasf:
@RobWin I was skeptical about that as well but it seemed fitting considering the naming of the other events

Issue #178, opened 2017-11-27
Spring Boot 2.0 Support


gbrehmer:
I know Spring Boot 2 is not released yet, but most of the changes are already fixed (2.0.0.M6)
e.g. Endpoint configuration has changed significantly (AbstractEndpoint not available anymore). To support both versions, e.g. @ConditionalOnClass is required for configuration

RobWin:
Would you like to contribute the changes?

schosin:
Just as a heads-up, Spring Boot 2 also replaced Dropwizard with Micrometer. This might have an effect on resilience4j-metrics.

Looking forward to Spring Boot 2 support.

HarishaAkasapu:
Seeing this issue after SpringBoot2 migration, any plans to address this?

RobWin:
I started working on it.
Please see https://github.com/RobWin/resilience4j-spring-boot2-demo.
Add the dependencies as follows 
```
repositories {
	maven { url 'http://oss.jfrog.org/artifactory/oss-snapshot-local/' }
	mavenCentral()
}


ext{
	resilience4jVersion = '0.13.0-SNAPSHOT'
}

dependencies {
	compile('org.springframework.boot:spring-boot-starter-webflux')
	compile("io.github.resilience4j:resilience4j-spring-boot2:${resilience4jVersion}")
	compile("io.github.resilience4j:resilience4j-reactor:${resilience4jVersion}")
}
```

HarishaAkasapu:
Thanks, will look into.

radekkaminski:
How to apply the CircuitBreaker to Flux in this version ? CircuitBreakerOperator seems to be deleted...

RobWin:
Please have a look at the demo on github. There us a link some comments above.
I guess you need resilience4j-reactor as a compile dependency.

radekkaminski:
got it - thanks

radekkaminski:
I can't connect Prometheus cause the simpleclient_spring_boot is dependant on Spring Boot 1.x - Does anyone know if there is an alternative ?

RobWin:
Spring Boot2 uses MicroMeter which supports Prometheus out of the box.

RobWin:
https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#production-ready-metrics-export-prometheus

michael-barker:
[...]

PR #231, opened 2018-05-09
Add validation to CircuitBreakerProperties

kesslerj:
`CircuitBreakerProperties`:
- removed `@Component` (not necessary due to usage of
`@EnableConfigurationProperties` in `CircuitBreakerAutoConfiguration`)
- Spring-Boot 2 supports using java.time.Duration as Property type, see [Spring-Boot documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-conversion-duration). So I
changed the type of `waitDurationInOpenState` - and renamed it. This also fixes a
difference between the javadoc and the code (javadoc said it would be
seconds, but it was milliseconds).
- added validation for properties. Application should fail fast on
mis-configuration. For this feature I had to add the hibernate-validator
dependency.
- changed default behaviour to `registerHealthIndicator=true`. Convention
over configuration - if you provide a feature, you should activate by
default.

`CircuitBreakerMetricsAutoConfiguration`:
- add possibility to disable registering of circuitbreaker metrics

Changes related to #178

### DIFF
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -23,6 +23,7 @@ ext {
     reactorVersion = '3.1.5.RELEASE'
     reactiveStreamsVersion = '1.0.2'
     micrometerVersion = '1.0.0'
+    hibernateValidatorVersion = '6.0.9.Final'
 
     libraries = [
             // compile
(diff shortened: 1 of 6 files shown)


RobWin:
Hi. Thx. I have a look at it.

PR #234, opened 2018-05-26
Non-blocking API for RateLimiter

storozhukBM:


vy:
@storozhukBM Thanks so much for the prompt action! Though if I understand it right, Reactor/Rx operators still use the old `getPermission()` call rather than the introduced `reservePermission()`. While the fix is trivial, I doubt if that is indeed a solution suitable for the general user profile. Anyway... Thanks again. Appreciated.

storozhukBM:
@vy 
Don't worry I'll make all necessary changes in operators too.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
```diff
@@ -268,6 +270,16 @@ static void waitForPermission(final RateLimiter rateLimiter) throws IllegalState
...
 
+    /**
+     * Reserves a permission from this rate limiter and returns nanoseconds you should wait for it.
+     * If returned long is negative, it means that you failed to reserve permission,
+     * possibly your {@code timeoutDuration} is less then time to wait for permission.
+     *
+     * @param timeoutDuration the maximum time you want to wait.
+     * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
+     */
+    long reservePermission(Duration timeoutDuration);
```
RobWin: From an API-perspective I would prefer to return a Duration instead of a long. But I guess this would lead to more. GC

vy:
@storozhukBM, for the reactive operators, my point was maybe it is best to not change them. Because, if I am not mistaken, this will prevent the reactive stream from propagating the back-pressure. Yes... I know... I was the one requesting them and I still want to use them. Though I am not sure if it is suitable for general consumption. Consider the following: You have a producer with a faster pace than the consumer and you piped them through a rate limiter delaying emitted items. Such a design necessitates  certain `onBackpressure...()` methods which will highly likely to be forgotten by users.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
storozhukBM: Yep, it's a low level API where we want to avoid unnecessary GC pressure.

storozhukBM:
@vy agree

thekalinga:
@vy @storozhukBM 

Looks like I am wrong. I just looked at the implementation & realised that we will have backpressure issue with current implementation as we are direct requesting upstream for data even before we are certain that there is no token available

The current implementation we have is wrong & it does not rate limit at all. It is pretending to do rate limit, but reality it is blindly letting upstream requests to go thru without any token check & when the upstream emits an event, it is verifying whether a token is available or not. This is completely wrong impelmentation

The rate limiter is supposed to limit calls to upstream (say a network call), but we are sending `request`s blindly, so there is no rate limit

Check `ResilienceBaseSubscriber#request` implementation. It does no checks and blindly lets the upstream calls be made. The rate limiting should happen here. Instead they are happening in `RateLimiterSubscriber#hookOnNext` which means we are rate limiting downstream events when we are supposed to rate limit upstream `request`s

thekalinga:
I have raised an issue with more contex

https://github.com/resilience4j/resilience4j/issues/311

storozhukBM:
@thekalinga great if you know how to make some things better, it is more than welcome. But please, be careful about your subjective judgments about code quality of other contributions, it doesn't bring anything valuable to this conversation.

thekalinga:
@storozhukBM You are right. We are in opensource world where everyone is doing their best to help everyone else. I should have been more gracious than I was

madgnome:
As the person who wrote "the mess" I agree with @storozhukBM comment...

Issue #222, opened 2018-04-05
[resilience4j-ratpack] Documentation is misleading, incomplete, or invalid


ndugger:
While attempting to implement the resilience4j-ratpack package into my ratpack application, I am encountering a few issues regarding the [documentation](http://resilience4j.github.io/resilience4j/#_ratpack_starter)

### 1. The first of which is the following line of code from the [basic usage](http://resilience4j.github.io/resilience4j/#_basic_usage):
```java
install(Resilience4jModule.class);
```
The error I get from that can be seen below:

![](https://i.imgur.com/pQuBKS5.png)

### 2. Later in the docs, about [adding configuration](http://resilience4j.github.io/resilience4j/#_adding_bulkheads_circuitbreakers_ratelimiters_and_retries), it uses an instance instead of passing in the class, but in calling the `configure` method, I am getting an exception:
```
java.lang.ClassNotFoundException: io.github.resilience4j.metrics.CircuitBreakerMetrics
```
I'm not even trying to use `CircuitBreaker`. Below is my code:
```java
Resilience4jModule resilience = new Resilience4jModule();
resilience.configure(config -> config
    .retry("elastic", retry -> retry
        .maxAttempts(3)
        .waitDurationInMillis(100)
    )
);
install(resilience);
```

No idea why I'd be getting an exception about circuit breaking. I included the curcuitbreaker package in my gradle, and still I get the exception.

### 3. Missing explanations/definitions. For example, in the docs for [retry](http://resilience4j.github.io/resilience4j/#_retry_2), it references a `retry` variable without any context around what that is. It also is too simple of an example to be useful, resolving with a string.

What if my promise resolves with a `ReceivedResponse` from an external service that I am attempting to retry on failures? I can't just instantiate one. Maybe some clarification needs to exist on what you are allowed to resolve from your promises that you attempt to retry.

---

At this point, I am attempting to use the annotation approach, but cannot get past the exception being thrown in bullet point 2.

RobWin:
Thx. I have a look at it. The upgrade to the latest Ratpack version might have broken something.

See docs for a Retry introduction:

http://resilience4j.github.io/resilience4j/#_retry

RobWin:
Regarding point 2) 
Our resilience4j-ratpack module has a compileOnly dependency to resilience4j-metrics or resilience4j-prometheus.
I guess you've enabled metrics in your config. If that's the case you have to add resilience4j-metrics as a compile dependency to your application.

[...]

PR #238, opened 2018-06-13
specify recordFailurePredicate in ratpack configs, closes #237, #222

### LINKED ISSUE #237: [resilience4j-ratpack] Integrate use of recordFailurePredicate in configs
Currently the ratpack configs do not allow the configuration of `recordFailurePredicate`. This should be added to `io.github.resilience4j.ratpack.circuitbreaker.CircuitBreakerConfig` in 2 ways:

1) as a predicate in the same way as in `io.github.resilience4j.circuitbreaker.CircuitBreakerConfig`. Default to `(exception) -> true`
2) as a list of classes that should be recorded as omitted (will NOT cause recording of an error). Default to empty list.

Then `Resilience4jModule` can set up circuitbreakers to honor this predicate and the provided config.

drmaas:
Also addresses some documentation inconsistencies.


drmaas:
@RobWin ptal

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/circuitbreaker/CircuitBreakerTransformer.java
```diff
@@ -72,16 +87,10 @@ public void success(T value) {
...
-                            if (recoverer != null) {
-                                down.success(recoverer.apply(throwable));
-                            } else {
-                                down.error(throwable);
-                            }
-                        } catch (Throwable t) {
-                            down.error(t);
+                        if (recordFailurePredicate.test(throwable)) {
+                            circuitBreaker.onError(durationInNanos, throwable);
                         }
```
amityatagiri: in the else, shouldn't we have circuitBreaker.onSuccess(durationInNanos) call?

drmaas: @amityatagiri @RobWin This appears to be redundant - as the same check is present in `CircuitBreakerStateMachine.onError`

RobWin: @amityatagiri @drmaas This is the case when a decorated service throws a business exception (not a technical exception) and the exceptions should be ignored and not count as an error or success.
The CircuitBreakerStateMachine emits a `CircuitBreakerOnIgnoredErrorEvent` in that case.

From my point of view the `CircuitBreakerTransformer` can invoke `circuitBreaker.onError` directly and must not do the check `circuitBreakerConfig.getRecordFailurePredicate().test(throwable)`, because this is implemented in `circuitBreaker.onError` anyway.

Issue #242, opened 2018-06-22
Meters are duplicated when applying common tags with Spring Boot 2 auto-configuration


michael-barker:
Using the example from the the [Spring Boot docs](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-getting-started) to register common tags it appears that meters are duplicated, a meter with no tags an a meter with the common tags.

Using the [JVM MeterBinders](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.java#L83) as a reference it looks like a `MeterBinder` bean should be registered as a bean and we let Spring auto-configuration take care of binding it to a `MeterRegistry` instead of directly binding it.

PR #243, opened 2018-06-22
Allow Spring to bind MeterBinders instead of directly binding

michael-barker:
When binding directly to a registry we create our initial meters without any additional tags that might be applied by meter filters. This causes the meter to be duplicated, once with no tags and again with
the common tags.

Resolve: #242

### DIFF
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -24,7 +24,7 @@ ext {
     prometheusSimpleClientVersion = '0.3.0'
     reactorVersion = '3.1.5.RELEASE'
     reactiveStreamsVersion = '1.0.2'
-    micrometerVersion = '1.0.4'
+    micrometerVersion = '1.0.5'
     hibernateValidatorVersion = '6.0.9.Final'
 
     libraries = [
--- a/resilience4j-spring-boot2/src/main/java/io/github/resilience4j/circuitbreaker/autoconfigure/CircuitBreakerMetricsAutoConfiguration.java
+++ b/resilience4j-spring-boot2/src/main/java/io/github/resilience4j/circuitbreaker/autoconfigure/CircuitBreakerMetricsAutoConfiguration.java
@@ -15,16 +15,14 @@
  */
 package io.github.resilience4j.circuitbreaker.autoconfigure;
 
+import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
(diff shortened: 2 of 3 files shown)


resilience4j-spring-boot/src/main/java/io/github/resilience4j/circuitbreaker/autoconfigure/CircuitBreakerMetricsAutoConfiguration.java
```diff
@@ -39,9 +36,7 @@
 @AutoConfigureBefore(MetricRepositoryAutoConfiguration.class)
 public class CircuitBreakerMetricsAutoConfiguration {
     @Bean
-    public CircuitBreakerMetrics registerCircuitBreakerMetrics(CircuitBreakerRegistry circuitBreakerRegistry, MetricRegistry metricRegistry){
-        CircuitBreakerMetrics circuitBreakerMetrics = CircuitBreakerMetrics.ofCircuitBreakerRegistry(circuitBreakerRegistry);
-        metricRegistry.registerAll(circuitBreakerMetrics);
```
RobWin: I think your described solution only works with Spring Boot 2 and Micrometer, but not with Spring Boot 1 and Dropwizard Metrics, or?

RobWin: I think in Spring Boot 1 it's still necessary to register all Metrics from a `MetricSet`

michael-barker: Thanks, I'll take a look at it.

michael-barker: Thanks for catching that breaking change!  I've been using Micrometer's support for Spring Boot 1.5 and completely missed that this was registering it with a `MetricRegistry` and not a `MeterRegistry`.

I reverted the Spring Boot 1.x change as there should be no issue when using Dropwizard Metrics since there are no tags :)

PR #244, opened 2018-06-27
Corrected link to Prometheus Metrics Integration

kdombeck:
Old one was invalid


storozhukBM:
@kdombeck  thanks for fixing this

PR #246, opened 2018-07-03
make async retrofit call not make the request when circuit is open

szmg:
Hi, when using the Retrofit binding in async mode, the request is fired regardless of the state of the circuit. This PR contains a test case and the fix.


storozhukBM:
Thank you for this fix, @szmg

szmg:
Thanks for merging this so quickly :)

Issue #247, opened 2018-07-13
TimeLimiter brings rxjava2 dependency


edudar:
I don't see any code in `io.github.resilience4j.timelimiter` that would require `rxjava2` dependency but it's in `compile` configuration nevertheless. This also means it's exposed as transitive one further down the road.

edudar:
Opened https://github.com/resilience4j/resilience4j/pull/248

PR #248, opened 2018-07-13
Removed rxjava2 dependency for time limiter

edudar:
There is no code that requires it in the module. Fixes https://github.com/resilience4j/resilience4j/issues/247


RobWin:
Thx

Issue #249, opened 2018-08-01
[Spring Boot2] CircuitBreakerMetricsAutoConfiguration scans CircuitBreakerRegistry eagerly and does not bind metrics for circuit breakers registered after the bean is created


berrueta:
Spring Boot2 `CircuitBreakerMetricsAutoConfiguration` calls `CircuitBreakerMetrics.ofCircuitBreakerRegistry`, which immediately calls `circuitBreakerRegistry.getAllCircuitBreakers()`. Therefore it only binds metrics for circuit breakers that have already been registered against the `CircuitBreakerRegistry` at the time the `CircuitBreakerMetrics` bean is instantiated.

For applications that create and register circuit breaker instances as standalone Spring beans or as part of the initialisation of other beans, that means their circuit breakers will not publish metrics.

A possible solution would be to receive a `List<CircuitBreaker>` instead of the `CircuitBreakerRegistry`. Spring would populate that list with all the circuit breaker instances in the context. As part of that solution, we would also copy the static factory methods from `io.github.resilience4j.metrics.CircuitBreakerMetrics` to `io.github.resilience4j.micrometer.CircuitBreakerMetrics`.

RobWin:
Thx. Would you like to create a PR?

berrueta:
Yes, I will raise a PR.
Do you have a CLA for this project?

berrueta:
Closed by #254

PR #254, opened 2018-08-18
Additional factory methods - issue #249

berrueta:
These factory methods are identical to the ones already existing in metrics/CircuitBreakerMetrics.java. Their purpose is to allow Spring Boot applications to instantiate CircuitBreakerMetrics and have fine-grained control over which circuit breakers are bound to metrics, in particular in scenarios where the CircuitBreakerRegistry may not be completely populated by the time the CircuitBreakerMetrics class is instantiated.

This is not a complete solution for issue #249 . In fact, applications experiencing the problem described in that issue will still suffer the problem. However, these changes allow those applications to define their own CircuitBreakerMetrics bean independently of the CircuitBreakerRegistry.

An ideal solution would involve changing CircuitBreakerMetricsAutoConfiguration to ensure that the provided CircuitBreakerMetrics bean performs the metrics binding only after all the circuit breakers have been registered with the CircuitBreakerRegistry. However, that approach might not be backwards compatible, so I have decided to leave it for later.


storozhukBM:
@berrueta thank you very much for this contributions and ideas.

PR #255, opened 2018-08-18
Circuit breaker elapsed timer metric

berrueta:
Publish a timer metric tagged with the outcome of the operation.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/MetricNames.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/MetricNames.java
@@ -8,4 +8,5 @@ public class MetricNames {
     public static final String BUFFERED = "buffered";
     public static final String BUFFERED_MAX = "buffered_max";
     public static final String STATE = "state";
+    public static final String ELAPSED = "elapsed";
 }
--- a/resilience4j-documentation/src/docs/asciidoc/addon_guides/micrometer.adoc
+++ b/resilience4j-documentation/src/docs/asciidoc/addon_guides/micrometer.adoc
@@ -45,6 +45,7 @@ For each circuit breaker this registry will export:
 * `buffered` - current number of buffered calls `[int]`
 * `buffered_max` - maximum number of buffered calls `[int]`
 * `not_permitted` - current number of not permitted calls `[int]`
+* `elapsed` - timer metric measuring the elapsed time of the decorated calls `[Duration]`, tagged with the outcome of the operation: `result:success`, `result:error` or `result:ignoredError`.
 
 ===== RateLimiter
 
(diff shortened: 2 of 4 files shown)


berrueta:
CC: @madgnome

madgnome:
- [ ] Add a test to check that we are actually recording the time on events

We probably want to add the same metric to `resilience4j-metrics` (Dropwizard metrics) module and `resilience4j-prometheus`

Otherwise LGTM.

berrueta:
Thanks for the feedback. I've added the documentation. I've also added a couple of tests to check that the metric is being recorded and correctly tagged.

I have no experience with Dropwizard Metrics and Prometheus. The documentation suggests that users of Prometheus can use `CallMeter` to obtain information about execution time distribution. As for DW Metrics, version 3.x does not seem to support tagging, which IMO is necessary to ensure that the metric is useful.

storozhukBM:
Hi @berrueta,
Sorry for the late feedback.
Can we consider some option to make this set of metrics optional and not enabled by default?
There is actually performance reasons behind that. As you maybe have noticed, current metrics are gathered through dedicated `metrics` API of CircuitBreaker. This API was designed to have almost zero performance impact when you use it. Unfortunately it is not true about our event system. You won't pay anything when you not subscribe for events, but after subscription you will get some small, but noticeable latency impact and increased GC pressure.
I want to leave our users an option to avoid possible performance drawbacks, so there is few things we can do:
1) make this metrics optional
2) redesign this metrics in such way, where we can collect them inside CircuitBreaker without noticeable performance impact.

storozhukBM:
And after all I think almost all users that use metrics are actually gathering latency metrics from the directly from protected methods behind CircuitBreaker, so our elapsed metrics will duplicate almost the same measurements. That is another reason to make them optional.

berrueta:
Hi @storozhukBM ,
Thank you for the feedback. I see your points. What do you think would be the best way to make this behaviour optional? Perhaps a boolean flag in the constructor of `CircuitBreakerMetrics`?

storozhukBM:
maybe factory method `ofRegistryWithElapsedTimeMetrics` and also separate constructor

RobWin:
Any progress on this PR?

berrueta:
Sorry, this fell through the cracks. I'll try to figure out a way to add this metrics without compromising performance.

RobWin:
Please have a look at PR #335. We are refactoring metrics and want to add timer metrics as well.

voievodin:
Hey guys, it seems this one was hanging around for a while. I've drafted an implementation that takes into account previously mentioned points and the binding is optional, check 👉 [here](https://github.com/resilience4j/resilience4j/compare/master...voievodin:elapsed-cb). If you guys okay with the approach I will finish it @RobWin @berrueta please have a look.

RobWin:
I have the feeling that we actually shouldn't include this into the `TaggedCircuitBreakerMetrics `.
I agree to what @storozhukBM said:
>And after all I think almost all users that use metrics are actually gathering latency metrics directly from protected methods behind CircuitBreaker

Instead we should add a `Timer` decorator to resilience4j-micrometer comparable to resilience4j-metrics.
See: [TimerImpl](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-metrics/src/main/java/io/github/resilience4j/metrics/internal/TimerImpl.java)

You use the `Timer` decorator to decorate any function and then you get metrics for `TOTAL`, `FAILED` and  `SUCCESSFUL` calls. 
I think we need a call rate and call responses time percentiles for `TOTAL`,  `FAILED` and `SUCCESSFUL` calls. 

This would allow the user to use a `Timer` even without a CircuitBreaker or in combination with a Bulkhead.

RobWin:
Micrometer only has only support for [@Timed](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/aop/TimedAspect.java)

RobWin:
The metrics are added in my next PR and will be included in the upcoming v0.15.0

Issue #263, opened 2018-09-20
RxJava2 Bulkhead operator might release too much permit


danielimre:
`io.github.resilience4j.bulkhead.operator.BulkheadOperator` might release two permits in case a `Single<Iterable>` or `Maybe<Iterable>` is flatmapped into an `Observable` with `take`.
It should always release one permit no matter how much dispose or terminal event is triggered.

RobWin:
Would you like to create a PR?

PR #264, opened 2018-09-20
Issue 263 fix: Fix bulkhead on Single and Maybe 

danielimre:
Issue 263 fix: Fix bulkhead on Single and Maybe to always release permit only once.


storozhukBM:
@danielimre thank you very much for such a quick reaction

danielimre:
Thanks, this was a fix for #263 
Feel free to close that.

PR #281, opened 2018-10-24
Avoid creating unnecessary logging strings

hakonlo:
slf4j provides its own string formatting mechanism, which will avoid calling `toString()` on its parameters if the log message is below the current log level.

The previous approach generated a considerable amount of garbage, and caused a bit of GC pressure.

See https://www.slf4j.org/faq.html for more information.

### DIFF
--- a/resilience4j-cache/src/main/java/io/github/resilience4j/cache/internal/CacheImpl.java
+++ b/resilience4j-cache/src/main/java/io/github/resilience4j/cache/internal/CacheImpl.java
@@ -81,7 +81,7 @@ public class CacheImpl<K, V>  implements Cache<K,V> {
                 return result;
             }
         }catch (Exception exception){
-            LOG.warn(String.format("Failed to get a value from Cache %s", getName()), exception);
+            LOG.warn("Failed to get a value from Cache {}", getName(), exception);
             onError(exception);
             return Option.none();
         }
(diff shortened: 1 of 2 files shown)


resilience4j-cache/src/main/java/io/github/resilience4j/cache/internal/CacheImpl.java
```diff
@@ -93,7 +93,7 @@ private void putValueIntoCache(K cacheKey, V value) {
                 cache.put(cacheKey, value);
             }
         } catch (Exception exception){
-            LOG.warn(String.format("Failed to put a value into Cache %s", getName()), exception);
+            LOG.warn("Failed to put a value into Cache {}", getName(), exception);
```
RobWin: Hi,
sure that this works? Isn't the the exception treated as a param for the formatted string instead of anexception?

hakonlo: Yes, as of slf4j 1.6.0, a parameter will be handled as an exception if it is the last argument. https://www.slf4j.org/faq.html#paramException

RobWin: Thx 👍

Issue #283, opened 2018-11-07
InMemoryBulkheadRegistry calling bulkheadConfigSupplier nedlessly


hcf:
When using `BulkheadRegistry.bulkhead("name", supplier);` the supplier is being called every time, even though a Bulkhead is already registered with the given name, and the result of the supplier is never used. This should be avoided.

RobWin:
Thx

PR #284, opened 2018-11-07
#283 Avoiding calling bulkheadConfigSupplier nedlessly

hcf:
This is already being avoided in the other registries


RobWin:
Thx, I have a look at it.

Issue #288, opened 2018-11-26
Improve Kotlin support via Nullability annotations


checketts:
Kotlin supports nullable types as a first class citizen. Adding nullability annotations to methods and annotations improves the null checking and also streamlines Kotlin support.

Without these annotations Kotlin treats returned values as an unknown (aka platform) type.

Example of similiar change I did with Micrometer: https://github.com/micrometer-metrics/micrometer/pull/310

Would a PR adding these annotations be welcome?

storozhukBM:
This is a great idea @checketts
If you want to contribute such PR it will be more than welcome.

RobWin:
Merged. Will be part of 0.14.0

PR #289, opened 2018-11-26
Initial nullable API support

checketts:
Here are the initial @Nullable annotations for `core` and `circuitbreaker` modules.

The way these annotations work are via JSR-305 annotation support. To avoid adding it as a dependency for other projects, we add it as an optional dependency and create our own meta-annotations that use it. Then we can use the internal annotations throughout our code.

By adding `package-info.java` and annotation those with `@NonNullApi` and `@NonNullFields` we set the default for all parameters, return types, and fields and being `@NonNull` (so we don't have to annotate all the individual variables)

There aren't any checks in the code blowing up if user were to pass in a null on a method that doesn't allow nulls, but now compiler will warn the user they are mis-using the API.

Adresses #288


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -248,7 +250,9 @@ public CircuitBreakerConfig build() {
             config.failureRateThreshold = failureRateThreshold;
             config.ringBufferSizeInClosedState = ringBufferSizeInClosedState;
             config.ringBufferSizeInHalfOpenState = ringBufferSizeInHalfOpenState;
-            config.recordFailurePredicate = errorRecordingPredicate;
+            if(errorRecordingPredicate != null) {
```
checketts: This is the only functional change in this PR. Previously a null `errorRecordingPredicate` would allow being used, but that seem undesired from what I could see in the code.

checketts:
Let me know if you want me to continue annotating additional modules in this PR, or have me create a new PR after this is merged so as to avoid creating too large a PR.

storozhukBM:
@checketts thank you for this contribution.
This PR looks good except you should update copyright headers.
I think you can continue annotating other modules(i'd target all core modules) inside this PR.

checketts:
@storozhukBM What specifically should I update in the copyright headers? Just the date? Or are you asking that I put my name in them?

storozhukBM:
Your name and date would be great, but if you want to stay anonymous you can leave previous name.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/event/AbstractRetryEvent.java
```diff
@@ -18,16 +18,19 @@
...
 abstract class AbstractRetryEvent implements RetryEvent {
 
     private final String name;
     private final ZonedDateTime creationTime;
     private final int numberOfAttempts;
+    @Nullable
     private final Throwable lastThrowable;
 
-    AbstractRetryEvent(String name, int numberOfAttempts, Throwable lastThrowable) {
+    AbstractRetryEvent(String name, int numberOfAttempts, @Nullable Throwable lastThrowable) {
```
checketts: We have one instance in the code that passes a null into the `lastThrowable` value. So that forced this to be nullable. which means the `toString` implementation of its children also need to be updated to handle that. Should I just have the `toString` default to `null` when it isn't set?

checketts:
Please be sure to review https://github.com/resilience4j/resilience4j/pull/289/commits/21f4fc053ec9b5b33d2a733c9827d83a23ba9aec which is the commit I put all the fixes for the nullability findings

storozhukBM:
Everything seems to be fine. Please rebase and I'll merge it to master.

storozhukBM:
And please pay attention at Codacy report.

storozhukBM:
Hi @checketts,
Do you want to continue work on this issue, it is almost done, just few small changes required.

storozhukBM:
Great 👍  Such a wonderful news, my congratulations to you and your spouse. Please don't worry about this things, unless you want to distract yourself for a little bit. I'll be able to do everything necessary and this changes will be release with 0.14.0 version.

RobWin:
Any progess? I really would like to merge it.

checketts:
Wow, time flies. I've rebased the branch and it is ready for merging.

PR #293, opened 2018-11-29
WaitDurationInOpenState with millisecond support

hkwu008:
I think  millisecond is  necessary for "waitDurationInOpenState" in some situation.


storozhukBM:
Hmmm, I believe you're right, in some situations millisecond scale is necessary.

storozhukBM:
Thank you for contribution.

Issue #258, opened 2018-08-31
Support The Use Of @CircuitBreaker On Methods That Return A Mono Or Flux


ryanjbaxter:
Say I have this simple class in a Spring Boot 2 app

```
@CircuitBreaker(name = "httpbin")
public class WebClientHttpbinService {

	private WebClient client;

	public WebClientHttpbinService() {
		this.client = WebClient.create("http://httpbin.org");

	}

	public Mono<Map> delay(int seconds) {
		return client.get().uri("/delay/{seconds}", seconds).exchange().flatMap(r -> r.bodyToMono(Map.class)).timeout(Duration.ofSeconds(3));
	}
}
```

I am finding that even if the `delay` method times out (the`WebClient` request takes longer than 3 seconds) that the `httpbin` circuit breaker does not could that as a failure.  This makes sense as the method actually completed successfully in returning the `Mono<Map>`.  So should the `@CircuitBreaker` annotation not be used with methods that return reactive types?

ryanjbaxter:
@madgnome not sure if you have any thoughts on this one as well

madgnome:
Yeah unfortunately this annotation doesn't work when using a reactive type for the reason you mentioned. Could you raise an issue? Thanks

ryanjbaxter:
Can we just repurpose this one?

madgnome:
Sure

ryanjbaxter:
I edited the title, if someone wants to remove the question label and mark it as an enhancement that would be great

Romeh:
Small question , if we add response check in circuit breaker logic , would not cover all possible types for us including reactive types as you can check ur response to detect failure cases if no exceptions is thrown ?

Or just having custom logic in bodyToMono to check if mono map is empty throw an exception ?

RobWin:
Could you please rephrase your question? I'm not sure If I understood your question correctly.

Romeh:
I meant to support reactive types that could be returned despite there was an error in calling external API for example instead of exceptions : 
1- if we add response check of the service call in circuit breaker failure detection we could cover that need by doing so and maybe other future response types 
2- Or we should relay on the developer to add the right mapping for possible cases via bodyToMono where he can check the returned response object and throw exception if failure is detected which will trigger circuit breaker failure detection

ryanjbaxter:
Can you provide an example of what you are suggesting?

Romeh:
For point 2 i meant : 
```
	webClient.get("urServiceUrl")
[...]

PR #302, opened 2018-12-13
#258 Add webflux support circuitbreaker annotation

Romeh:
Hey @RobWin  ,

I have created a pull request for your review and check the implementation extension to the CircuitBreaker annotation AOP for Web Flux types and if it is a good idea or not .

Please let me know what do u think , you can find a working case in springboot2 module:
io.github.resilience4j.circuitbreaker.CircuitBreakerAutoConfigurationTest


Romeh:
will be creating another fresh one

PR #303, opened 2018-12-17
#258 Add webflux support circuitbreaker annotation

Romeh:
@RobWin 
I have created the pull request to make it easier for review the code changes needed for #258


Romeh:
@RobWin  any news about that :) ?

ryanjbaxter:
Were there some formatting changes that were applied here?  It looks like maybe spacing?

Romeh:
Yes there are code reformatting being done from intelliJ before committing.

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
```diff
@@ -38,89 +42,126 @@
...
+	private Object defaultWebFlux(ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.circuitbreaker.CircuitBreaker circuitBreaker, String methodName) throws Throwable {
+		CircuitBreakerUtils.isCallPermitted(circuitBreaker);
+		long start = System.nanoTime();
+		try {
+			Object returnValue = proceedingJoinPoint.proceed();
+			if (returnValue instanceof Flux) {
+				Flux fluxReturnValue = (Flux) returnValue;
+				return fluxReturnValue.transform(CircuitBreakerOperator.of(circuitBreaker));
+			} else if (returnValue instanceof Mono) {
+				Mono fluxReturnValue = (Mono) returnValue;
```
ryanjbaxter: variable name here says flux but is a mono

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
```diff
@@ -38,89 +42,126 @@
...
+				return fluxReturnValue.transform(CircuitBreakerOperator.of(circuitBreaker));
+			} else {
+				throw new IllegalArgumentException("Not Supported type for the circuit breaker in web flux :" + returnValue.getClass().getName());
+
+			}
+		} catch (Throwable throwable) {
+			long durationInNanos = System.nanoTime() - start;
```
ryanjbaxter: Why only do this when there is a failure?

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
Romeh: updated !

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
Romeh: This is only for any exception may be thrown during the operator invocation , meanwhile the whole logic of success case and other possible failure cases are covered into the operator logic itself : CircuitBreakerOperator.java

RobWin:
Could you please add the Apache License Headers to all new files

resilience4j-spring/build.gradle
```diff
@@ -1,10 +1,11 @@
...
-    compile ( libraries.aspectj )
-    compileOnly ( libraries.hibernate_validator )
-    compileOnly ( libraries.spring_4_core, libraries.spring_4_context )
+    compile(libraries.aspectj)
+    compileOnly(libraries.hibernate_validator)
+    compileOnly(libraries.spring_4_core, libraries.spring_4_context)
     compile project(':resilience4j-annotations')
     compile project(':resilience4j-consumer')
     compile project(':resilience4j-circuitbreaker')
+    compile project(':resilience4j-reactor')
```
RobWin: I'm not sure about the compile dependency to resilience4j-reactor.
resilience4j-spring is used by pure Spring applications and resilience4j-spring-boot1 and resilience4j-spring-boot2.
This would introduce Spring Reactor as a compile dependency to all applications using resilience4j-spring, even if they don't want to use it.
Maybe we can change it to `compilyOnly`

RobWin:
Any update on this PR?

Romeh:
@RobWin  i see your point , but unfortunately we can not define it as compilyOnly as it will lead to class def not found exception due to the spring Aspect use it to support the reactor flow for circuit-breaker , so if we need to add support to reactor circuit breaker in spring circuit breaker annotation , we will need it there as a compile dependency. so do we need to add the support for it or we should ignore it ? or we need to think more about a modular way to support extensions to the spring annotations or we can introduce new module to support react spring annotation AOP other than the default resilience4j spring module, so it could be up to the user if he want the reactor support , he will add that module ?

RobWin:
But you would only get a ClassDefNotFound if you use `ApiType.WEBFLUX` without adding Spring Reactor as a compile dependency to your App.
We could also add Reative Annotations 
@storozhukBM What do you think?

Romeh:
@storozhukBM  any updates regarding Rob question :) ?

storozhukBM:
@Romeh I agree with @RobWin  for now we can make it `compilyOnly`

Issue #307, opened 2018-12-21
Bulkhead Support in Prometheus


vwallden:
Monitoring bulkhead state is critical, and it would be great if the Prometheus Metrics Integration project supported exposing the amount of available calls, which is already tracked in Bulkhead.Metrics. Currently only metrics for RateLimiters and CircuitBreakers are supported for Prometheus. 

It also occurs to me that a helpful additional metric for bulkheads might be the amount of rejected attempts to acquire a permit. Could this be a metric that would be beneficial to add to Bulkhead.Metrics?

RobWin:
Yes, Feel free to create a PR

vwallden:
Will do.

RobWin:
Will be part of 0.14.0

PR #308, opened 2018-12-24
Add bulkhead metrics support for Prometheus

vwallden:
As referenced in issue #307


vwallden:
Codacy complains about the constructor test in BulkheadExportsTest, because they don't contain assert() or fail(). The test is equivalent to tests in CircuitBreakerExportsTest and RateLimiterExportsTest. Is the test ok as-is?

storozhukBM:
Hi @vwallden,
Thank you for this contribution. PR looks good for me. Let's wait few days, maybe @goldobin, will have some time for review, if not I'll merge your PR and it will be part of our 0.14.0 release.

PR #319, opened 2019-01-08
add getting current circuit state ratpack endpoint

fromanator:
Simplifies getting a snapshot of circuit breaker states for the application. Includes a snapshot of the metrics too.


fromanator:
@RobWin I've added in the apache headers

RobWin:
thx

PR #330, opened 2019-01-28
Refactor rxjava operators

danielimre:
I've refactored rxjava operators and removed all internal rxjava references so it is working with newer versions of rxjava as well. I left the commits as is so it's easier to see what and how has changed.
I also made all operators use the same approach. This required a slight modification in the behavior of the circuitbreaker operator. It won't register the call if it completes after the subscription was cancelled but this is an edge case anyway.
I haven't touched RetryTransformator as I saw another PR addressing it.

### DIFF
--- a/resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/AbstractBulkheadOperator.java
+++ b/resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/AbstractBulkheadOperator.java
@@ -1,0 +1,203 @@
+package io.github.resilience4j.bulkhead.operator;
+
+import io.github.resilience4j.adapter.Permit;
+import io.github.resilience4j.bulkhead.Bulkhead;
+import io.github.resilience4j.bulkhead.BulkheadFullException;
+import io.reactivex.exceptions.ProtocolViolationException;
+import io.reactivex.plugins.RxJavaPlugins;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A disposable bulkhead acting as a base class for bulkhead operators.
+ *
+ * @param <T>          the type of the emitted event
+ * @param <DISPOSABLE> the actual type of the disposable/subscription
+ */
+abstract class AbstractBulkheadOperator<T, DISPOSABLE> extends AtomicReference<DISPOSABLE> {
+    private final Bulkhead bulkhead;
+    private final AtomicReference<Permit> permitted = new AtomicReference<>(Permit.PENDING);
+
+    AbstractBulkheadOperator(Bulkhead bulkhead) {
+        this.bulkhead = requireNonNull(bulkhead);
+    }
+
+    /**
+     * Disposes this operator exactly once.
+     */
+    protected void dispose() {
+        if (disposeOnce()) {
+            releaseBulkhead();
+        }
+    }
+
+    /**
+     * Gets whether this operator was already disposed or not.
+     *
+     * @return true if the operator was disposed, otherwise false
+     */
+    protected boolean isDisposed() {
+        return get() == getDisposedDisposable();
+    }
+
+    /**
+     * Gets the current disposable.
+     *
+     * @return the disposable
+     */
+    protected abstract DISPOSABLE getDisposable();
+
+    /**
+     * Gets the reference of the one and only disposed disposable.
+     *
+     * @return the disposed disposable
+     */
+    protected abstract DISPOSABLE getDisposedDisposable();
+
+    /**
+     * Disposes a disposable.
+     *
+     * @param disposable the disposable to dispose
+     */
+    protected abstract void dispose(DISPOSABLE disposable);
+
+    /**
+     * onSuccess ensured to be called only when safe.
+     *
+     * @param disposable the disposable
+     */
+    protected void onSubscribeInner(DISPOSABLE disposable) {
+        //Override when needed.
+    }
+
+    protected final void onSubscribeWithPermit(DISPOSABLE disposable) {
+        if (setDisposableOnce(disposable)) {
+            if (acquireCallPermit()) {
+                onSubscribeInner(getDisposable());
+            } else {
+                dispose();
+                onSubscribeInner(getDisposable());
+                permittedOnError(bulkheadFullException());
+            }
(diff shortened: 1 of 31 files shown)


danielimre:
Is it ok to suppress PMD check `CompareObjectsWithEquals` or should I really use `Objects.equals` in the failing class?

RobWin:
Thx for your PR. We will have a look at it.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/AbstractBulkheadOperator.java
```diff
@@ -0,0 +1,46 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A disposable bulkhead acting as a base class for bulkhead operators.
+ *
+ * @param <T> the type of the emitted event
+ * @param <D> the actual type of the disposable/subscription
+ */
+abstract class AbstractBulkheadOperator<T, D> extends PermittedOperator<T, D> {
+    private final Bulkhead bulkhead;
```
Romeh: make bulkhead transient or seriazlizable as Fields in a Serializable class must themselves be either Serializable or transient even if the class is never explicitly serialized or deserialized. For instance, under load, most J2EE application frameworks flush objects to disk, and an allegedly Serializable object with non-transient, non-serializable data members could cause program crashes, and open the door to attackers. In general a Serializable class is expected to fulfil its contract and not have an unexpected behaviour when an instance is serialized.
This rule raises an issue on non-Serializable fields, and on collection fields when they are not private (because they could be assigned non-Serializable values externally), and when they are assigned non-Serializable types within the class.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/BulkheadCompletableObserver.java
```diff
@@ -1,15 +1,15 @@
...
 import io.reactivex.disposables.Disposable;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * A RxJava {@link CompletableObserver} to wrap another observer in a bulkhead.
  */
-final class BulkheadCompletableObserver extends DisposableBulkhead implements CompletableObserver {
+final class BulkheadCompletableObserver extends DisposableBulkhead<Object> implements CompletableObserver {
     private final CompletableObserver childObserver;
```
Romeh: same remark as above

Romeh: same remark as above

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/BulkheadSubscriber.java
```diff
@@ -1,65 +1,63 @@
...
+
 /**
  * A RxJava {@link Subscriber} to wrap another subscriber in a bulkhead.
  *
  * @param <T> the value type of the upstream and downstream
  */
-final class BulkheadSubscriber<T> extends AtomicReference<Subscription> implements Subscriber<T>, Subscription {
-    private final Bulkhead bulkhead;
+final class BulkheadSubscriber<T> extends AbstractBulkheadOperator<T, Subscription> implements Subscriber<T>, Subscription {
     private final Subscriber<? super T> childSubscriber;
```
Romeh: same remark as above

Romeh:
@danielimre  can u please re-base the code changes and maybe have a look about the serialization remark which is recurrent in the abstract operators , observer classes and the subscriber classes in (bulkhead , rate limiter , circuit breaker), thx a lot for your deep support .

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/BulkheadOperator.java
```diff
@@ -33,6 +33,7 @@
 
 /**
  * A RxJava operator which wraps a reactive type in a bulkhead.
+ * Please note that Bulkhead shouldn't use wait time as it may block while acquiring the permit.
```
Romeh: * Please note that Bulkhead shouldn't use wait time as it may block while acquiring the permit. 

can u please explain more what you mean by that ?

danielimre: If `io.github.resilience4j.bulkhead.BulkheadConfig#maxWaitTime` is set on `Bulkhead` then `io.github.resilience4j.bulkhead.internal.SemaphoreBulkhead#tryEnterBulkhead` might block the current executing thread for that long which can be problematic if that's one of the computational scheduler's threads. Would be nice to have a non-blocking way of doing this.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/AbstractBulkheadOperator.java
danielimre: Sorry for not getting back sooner. 

Not sure what should be done here..
I don't think making it transient would make any sense. It would break anyway...
Should I not extend an `AtomicReference` but rather move it to a field?

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/BulkheadOperator.java
RobWin: Currently that's only possible by using maxWaitTime = 0

danielimre: Yes, I meant that. Should I rephrase this comment?

RobWin: Yes please

resilience4j-rxjava2/src/main/java/io/github/resilience4j/bulkhead/operator/AbstractBulkheadOperator.java
Romeh: Yes i guess it would be better

PR #342, opened 2019-02-20
Expose bulkhead max allowed concurrent calls metric

voievodin:
Hi there :wave:

The idea behind these changes is to expose upper configured bound for bulkhead max concurrent calls, since bulkhead configuration might be dynamically changed the visibility of the value seems to be essential. Also it gives _a simple way_ to visualize data in other (meaning different than current available number of concurrent calls) manner, a few examples (prometheus + grafana):

```
[Current concurrent calls]
resilience4j_bulkhead_Backend_max_allowed_concurrent_calls - resilience4j_bulkhead_Backend_available_concurrent_calls

[Max allowed concurrent calls]
resilience4j_bulkhead_Backend_max_allowed_concurrent_calls 
```
image

```
[Used concurrent calls in percent]
100 - 100 * resilience4j_bulkhead_Backend_available_concurrent_calls / resilience4j_bulkhead_Backend_max_allowed_concurrent_calls

80% Alert
```
image

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/Bulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/Bulkhead.java
@@ -414,6 +414,15 @@ public interface Bulkhead {
          * @return remaining bulkhead depth
          */
         int getAvailableConcurrentCalls();
+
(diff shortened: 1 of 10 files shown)


RobWin:
Thank you. LGTM

Issue #354, opened 2019-03-06
Proposal: resilience4j-micrometer integration should use tags


noroute:
# Context
Micrometer regards dimensions as a first-class citizen of time series and advises to not include them in the time series name (see https://micrometer.io/docs/concepts#_meters).

resilience4j-micrometer adds the name of CircuitBreakers (or Bulkheads, Retries and RateLimits for that matter) to the metric time series. This leads to limitations for querying (e.g. "find me all the open circuit breakers" would not be possible) and a large number of series being published (e.g. 6x #circuitbreakers) which is not optimal for most time series databases.

# Proposal
With this issue I propose to use tags for dimensions for all the Metrics currently supported (CircuitBreaker, RateLimit, Bulkhead, Retry).

# Consequences
The implementation changes would be trivial (see [Gauge Fluent Builder](https://micrometer.io/docs/concepts#_gauge_fluent_builder)) but would result in *backward incompatible change* in structure and names of time series. I am not sure if this is ok or if we should add this as a non-default option. I would be able to contribute a PR if we get agreement on an option.

# Example
I use CircuitBreaker as an example here as it has the most complex structure of the four.

## Current state
In Prometheus, this is what you get by default (`cb1` being the circuitbreaker name, only showing one metric):
```
# HELP resilience4j_circuitbreaker_cb1_successful  
# TYPE resilience4j_circuitbreaker_cb1_successful gauge
resilience4j_circuitbreaker_cb1_successful 0.0
```

## Proposed state: tags for names and metrics
To make discovery and drilldown easier, `name` and `metric` should be added as a dimension. This would allow drilldown on one specific circuitbreaker:
```
# HELP resilience4j_circuitbreaker  
# TYPE resilience4j_circuitbreaker gauge
resilience4j_circuitbreaker_stats{name="cb1",metric="successful"} 0.0
resilience4j_circuitbreaker_state{name="cb1"} 3.0
```

This version would support queries like:
* list of all open CircuitBreakers
* overview of buffered calls across all CircuitBreakers
* CircuitBreakers with more than 300 failed calls

RobWin:
Thx for this well structured issue :1st_place_medal: 
I totally agree.  But as you said it's a backward incompatible change and the resilience4j-prometheus module should get a new major version, if we do semantic versioning. 
We could start to give every module it's own version? Or should we start to separate add-on modules from the main project? But then the release process would be really cumbersome.

[...]

PR #355, opened 2019-03-08
Add tag based micrometer metric binders

voievodin:
Motivation
--- 
The current implementation of micrometer binders exposes all the needed metrics to build monitoring around it, however it doesn't suit well for some of the provider implementations, such as Prometheus. The main inconvenience here is that values are inlined into the metric names, while tags would be more suitable. The tags basically mapped to the Prometheus [labels](https://prometheus.io/docs/practices/naming/#labels) and allow to build generic graphs and take advantage of monitoring tools features, like  Grafana variables.

Simply changing the existing binders to export tags doesn't work well because of a few reasons:
1. Most likely we want to keep things backward compatible, thus those who have graphs built around existing approach shouldn't suffer after they update the library.
2. I guess the current approach is perfectly fine for dropwizard (and maybe other) users, since it doesn't have tags concept at all [yet](https://github.com/dropwizard/metrics/issues/1175).
3. We could simply expose twice as many metrics from the existing binders, but the clients most likely need only one subset of those (with tags or without) thus it would be odd and polluting the monitoring system.

Naming
---
### Bulkhead
```
resilience4j_bulkhead_available_concurrent_calls{name="Backend1"}
resilience4j_bulkhead_max_allowed_concurrent_calls{name="Backend1"}
```

### Circuit Breaker
```
resilience4j_circuitbreaker_state{name="Backend1"}

resilience4j_circuitbreaker_calls{name="Backend1", kind="successful"}
resilience4j_circuitbreaker_calls{name="Backend1", kind="failed"}
resilience4j_circuitbreaker_calls{name="Backend1", kind="not_permitted"}

resilience4j_circuitbreaker_buffered_calls{name="Backend1"}
resilience4j_circuitbreaker_max_buffered_calls{name="Backend1"}
```

### Retry
```
resilience4j_retry_calls{name="Backend1", kind="successful_with_retry"}
resilience4j_retry_calls{name="Backend1", kind="successful_without_retry"}
resilience4j_retry_calls{name="Backend1", kind="failed_with_retry"}
resilience4j_retry_calls{name="Backend1", kind="failed_without_retry"}
```

### Async Retry
```
resilience4j_async_retry_calls{name="Backend1", kind="successful_with_retry"}
resilience4j_async_retry_calls{name="Backend1", kind="successful_without_retry"}
resilience4j_async_retry_calls{name="Backend1", kind="failed_with_retry"}
resilience4j_async_retry_calls{name="Backend1", kind="failed_without_retry"}
```

### Rate Limiter
```
resilience4j_ratelimiter_available_permissions{name="Backend1"}
resilience4j_ratelimiter_waiting_threads{name="Backend1"}
```

Usage example
---
Taking advantage of grafana variables:
```
// backend is defined as 
label_values(resilience4j_bulkhead_available_concurrent_calls, name)

// the actual query
resilience4j_bulkhead_available_concurrent_calls{name=~"[[backend]]"}
```
image

### DIFF
--- a/resilience4j-documentation/src/docs/asciidoc/addon_guides/micrometer.adoc
+++ b/resilience4j-documentation/src/docs/asciidoc/addon_guides/micrometer.adoc
@@ -18,12 +18,20 @@ final Bulkhead boo = bulkheadRegistry.bulkhead("boo");
 // Register all bulkheads at once
 BulkheadMetrics bulkheadMetrics = BulkheadMetrics.ofBulkheadRegistry(bulkheadRegistry);
 bulkheadMetrics.bindTo(meterRegistry);
+
+// Or for tag based metrics
+TaggedBulkheadMetrics.ofBulkheadRegistry(bulkheadRegistry).bindTo(meterRegistry);
 --
 
 For each bulkhead this registry will export:
 
 * `available_concurrent_calls` - instantaneous read of the number of currently available concurrent calls `[int]`
 * `max_allowed_concurrent_calls` - maximum number of allowed concurrent calls `[int]`
+
+Tag based metrics:
+
+* `resilience4j_bulkhead_available_concurrent_calls{name="name of bulkhead"}` - number of available concurrent calls
+* `resilience4j_bulkhead_max_allowed_concurrent_calls{name="name of bulkhead"}` - maximum number of allowed concurrent calls
 
 ===== CircuitBreaker
 
@@ -36,6 +44,9 @@ final CircuitBreaker boo = circuitBreakerRegistry.circuitBreaker("boo");
 // Register all circuit breakers at once
 CircuitBreakerMetrics circuitBreakerMetrics = CircuitBreakerMetrics.ofCircuitBreakerRegistry(circuitBreakerRegistry);
 circuitBreakerMetrics.bindTo(meterRegistry);
+
+// Or for tag based metrics
+TaggedCircuitBreakerMetrics.ofCircuitBreakerRegistry(circuitBreakerRegistry).bindTo(meterRegistry);
 --
 
 For each circuit breaker this registry will export:
@@ -47,6 +58,16 @@ For each circuit breaker this registry will export:
 * `buffered_max` - maximum number of buffered calls `[int]`
 * `not_permitted` - current number of not permitted calls `[int]`
 
+Tag based metrics:
+
+* `resilience4j_circuitbreaker_state{name="name of circuitbreaker"}` - the state of the circuit breaker where 0-CLOSED, 1-OPEN, 2-HALF-OPEN
+* `resilience4j_circuitbreaker_calls{name="name of circuitbreaker", kind="one of the values below"}` - the number of calls of a certain kind
+  - `successful` - indicates successful calls
+  - `failed` - indicates failed calls
+  - `not_permitted` - indicates not permitted calls
+* `resilience4j_circuitbreaker_buffered_calls{name="name of circuitbreaker"}` - the number of buffered calls
+* `resilience4j_circuitbreaker_max_buffered_calls{name="name of circuitbreaker"}` - the maximum number of buffered calls
+
 ===== RateLimiter
 
 [source,java]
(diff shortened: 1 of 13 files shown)


RobWin:
Thx. LGTM.

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/TagNames.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.resilience4j.micrometer.tagged;
+
+/** Common constants for metric binder implementations based on tags. */
+public final class TagNames {
+    public static final String NAME = "name";
+    public static final String KIND = "kind";
```
RobWin: Just out of interest. Why did you call it `kind`?

voievodin: The hardest part :fire: 

For example, for `resilience4j_circuitbreaker_calls` i read it like:
> The kind of circuit breaker calls measured by this meter

The alternatives that were also considered:
1. `characteristic`, if we name the metric `resilience4j_circuitbreaker_ring_buffer` then it would make sense, since it would be characteristic of the ring buffer, and I like this name, except the fact that it's too long/errorprone and the ring_buffer part relies on circuit breaker implementation details
2. `metric`/`measure`/`meter` wasn't sure about these ones because the actual meter name + tags is a  **metric**, and having "metric" word inside it looks weird
3. `type`/`category` basically the same as kind

Would you suggest naming it differently ?

voievodin:
Thanks for reviewing @RobWin. 
If you're okay with the approach and we agree on naming, I can also cover the left bits
- tagged implementations for rate limiter and retries, since we have those in place for non tag based binder
- documentation

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/TagNames.java
fitzoh: It would probably make sense to mirror the naming of the [Prometheus integration](http://resilience4j.github.io/resilience4j/#_prometheus_metrics_integration):
Screen Shot 2019-03-15 at 7 34 17 PM

voievodin: Can be an option, yes. I did it this way to make everything consistent, I mean that would be weird to have this one for bulkheads
```
resilience4j_bulkhead_concurrent_calls{name="Backend1", call_result="max_allowed"}
```
But maybe it should not be that consistent.

RobWin: Let's keep the discussion in the issue https://github.com/resilience4j/resilience4j/issues/354
I already made some suggestion to rename it.
I don't want to keep the naming of the Prometheus Integration. We should improve it.

voievodin:
Updated according to the discussion in the #354

RobWin:
Are you done? Can I start reviewing? 
Would you like to add the changes from PR #255 as well?
And also incooperate the review findings of @storozhukBM ?

voievodin:
Hello, yes the PR is ready for the review, sorry for not mentioning that.

> Would you like to add the changes from PR #255 as well?

I would avoid adding the new metrics within this PR to make it simpler. I assume that the next logical thing to change would be the existing, dedicated prometheus metrics, so that they are aligned with the naming introduced here. I can do that as well, if don't have other plans for it.

RobWin:
Yes, please adapt prometheus metrics as well. 
I will merge it then.

voievodin:
Okay, will do it this week

voievodin:
Here we go https://github.com/resilience4j/resilience4j/pull/375, I decided to do a separate PR to keep things clearer and simplify the review, even though the topic is the same.

resilience4j-documentation/src/docs/asciidoc/addon_guides/micrometer.adoc
```diff
@@ -75,6 +104,9 @@ final Retry retry = retryRegistry.retry("testLimit");
 // Register all retries at once
 RetryMetrics retryMetrics = RetryMetrics.ofRetryRegistry(retryRegistry);
 retryMetrics.bindTo(meterRegistry);
+
+// Or for tag baed metrics
```
RobWin: Typo: baed -> based

Issue #301, opened 2018-12-11
RxJava 2.2.4 broke operator implementations


sfitts:
Release [2.2.4](https://github.com/ReactiveX/RxJava/releases/tag/v2.2.4) of RxJava removed `SubscriptionHelper.isCancelled` which is currently used by the operator implementations.  Honestly I'm tempted to log this as an RxJava bug since this kind of breaking change in a patch release is unfriendly.  However, the class in question is internal, so...

Anyway, until this is fixed moving beyond 2.2.3 will be impossible.

danielimre:
While making `BulkheadSubscriber`, `CircuitBreakerSubscriber` and `RateLimiterSubscriber` work both with former and latest RxJava is simple, I found that 2.2.4 also broke `RetryTransformer` which needs to be redesigned if we want to support both versions of RxJava. The problem is with `SubscriptionArbiter` as its constructor has changed.
Looking at the implementation of e.g. `Observable#retry` and `RetryTransformer`, they are pretty close to each other. I assume `RetryTransformer` was originated from RxJava code directly. The question is whether we want to do the same again as it would still require internal classes to be used. (I haven't checked if it can be done in a backward-forward compatible way at all.) It would be good to avoid similar issues in the future. 
@RobWin what's your thought on this?

Romeh:
@danielimre  i guess this will be covered by https://github.com/resilience4j/resilience4j/pull/356

Romeh:
@sniff 0.14.0 is released , can u please test it out again ?

PR #356, opened 2019-03-08
Retry RxJava2 enhancements

adik993:
* RxJava2 was upgraded to latest `2.2.7` version
 * Support for `Completable` reactive type
 * Support for `retryOnResult`

### DIFF
--- a/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
+++ b/resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
@@ -18,22 +18,10 @@ package io.github.resilience4j.retry.transformer;
 
 import io.github.resilience4j.retry.Retry;
 import io.reactivex.*;
-import io.reactivex.internal.subscriptions.SubscriptionArbiter;
 import org.reactivestreams.Publisher;
-import org.reactivestreams.Subscriber;
-import org.reactivestreams.Subscription;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.util.concurrent.atomic.AtomicInteger;
 
 public class RetryTransformer<T> implements FlowableTransformer<T, T>, ObservableTransformer<T, T>,
-        SingleTransformer<T, T>, CompletableTransformer {
-
-    private static final Logger LOG = LoggerFactory.getLogger(RetryTransformer.class);
-
-    private static final RuntimeException RESULT_EXCEPTION = new RuntimeException("retry due to retryOnResult predicate");
-
+        SingleTransformer<T, T>, CompletableTransformer, MaybeTransformer<T, T> {
     private final Retry retry;
 
     private RetryTransformer(Retry retry) {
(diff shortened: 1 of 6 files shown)


RobWin:
Thanks for your PR. Could you pelase create an issue for this PR.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
```diff
@@ -58,51 +53,67 @@ private RetryTransformer(Retry retry) {
...
+            RetrySubscriber<T> retrySubscriber = new RetrySubscriber<>(downstream,
+                    retry.getRetryConfig().getMaxAttempts(), sa, flowable, retry.context(), true);
             flowable.subscribe(retrySubscriber);
         });
     }
 
+    @SuppressWarnings("unchecked")
+    private void applyRetrySubscriber(Subscriber<? super T> downstream, Flowable<T> flowable) {
+        Retry.Context<T> context = retry.context();
+        flowable = flowable.doOnNext(value -> throwExceptionToForceRetryOnResult(context, value));
```
RobWin: This change uses an Exception to change the flow control.
Could we instead invoke `context.onResult(...)` in  `RetrySubscriber.onNext(...)` and invoke `subscribeNext` when onResult returns true?
Or doesn't this work?

adik993: Hi @RobWin,
I even have that in my local history :smile:. In short it does not work, but let me elaborate on that a little bit.

When we do that, this is what happens:
1. `RetrySubscriber.onSubscribe()` is called for the first time
2. `RetrySubscriber.onNext()` receives the value, checks for the `context.onResult(t)` and proceeds to `subscribeNext()`
3. `subscribeNext()` gets called and passes the trampolining check since it's our first "recurrent" call
    1.`RetrySubscriber.onSubscribe()` is called for the second time (our retry)
    2. `SubscriptionArbiter.setSubscription` is called and our previous subscription is cancelled, so when it finishes at some point in time it won't call `onComplete()`. Lastly our new stream starts
    3. `RetrySubscriber.onNext` is called, but this time since we are in recurrent call trampolining won't allow doing subscribe again in this call stack
    4. `RetrySubscriber.onNext` will finish, but since we didn't call the subscribe immediately the stream was not closed and RxJava will notify the subscriber that the stream completed. At this point we are already broken since there will be at least few more `onComplete` calls for the subscriptions scheduled via trampolining.

By the way I came up with another approach that seems to work and we do not have to user RxJava's internal api's. I'll push that any minute

adik993:
> Thanks for your PR. Could you pelase create an issue for this PR.
> 
> Or is it #301

To be honest I haven't looked through the issues, but it should fix the one you mentioned and add some functionality as well

adik993:
I've added an issue  #359 for the stuff not covered by the aforementioned #301

resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
RobWin: Thanks for your answer

resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
```diff
@@ -58,118 +41,78 @@ private RetryTransformer(Retry retry) {
...
+        //noinspection unchecked
+        Context<T> context = new Context<T>(retry.context());
+        return upstream.doOnSuccess(context::throwExceptionToForceRetryOnResult)
+                .retryWhen(errors -> errors.doOnNext(context::onError))
+                .doOnSuccess(t -> context.onComplete())
+                .doOnComplete(context::onComplete);
+    }
+
+    private static class Context<T> {
+        private static final RuntimeException RESULT_EXCEPTION = new RuntimeException("retry due to retryOnResult predicate");
```
RobWin: If we have to use exceptions for flow control, we should introduce a custom runtime exception like
`RetryDueToResultException`. What do you think?

resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
```diff
@@ -58,118 +41,78 @@ private RetryTransformer(Retry retry) {
...
+            if (context.onResult(value))
+                throw RESULT_EXCEPTION;
         }
 
-        @Override
-        public void onComplete() {
-            if (LOG.isDebugEnabled()) {
-                LOG.info("onComplete");
+        void onError(Throwable throwable) throws Exception {
+            if (throwable == RESULT_EXCEPTION) return;
```
RobWin: We could replace it by an `throwable instanceof RetryDueToResultException`

RobWin:
Btw. Thank you for your effort.

resilience4j-rxjava2/src/main/java/io/github/resilience4j/retry/transformer/RetryTransformer.java
adik993: Sounds good will do that

PR #377, opened 2019-03-27
Support recovery function on spring

### LINKED ISSUE #309: [question] Spring AOP and recovery function?
Hello,

I'm currently trying to migrate a spring boot project that's using hystrix to resilience4j.
Currently I'm using this simple hystrix command:
```JAVA
 @HystrixCommand(fallbackMethod = "myrecoverymethod")
```

In the docs, I saw that ratpack users can use 
```JAVA
@CircuitBreaker(name = "backendA", recovery = MyRecoveryFunction.class)
```
 to call a recovery function.
How can I achieve something similar with Spring AOP?  
Looks like the Spring AOP annotation currently only supports setting a name.

lespinsideg:
resolves #309

- support recovery function in annotations for spring.


resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java
```diff
@@ -61,7 +62,9 @@ public Object bulkheadAroundAdvice(ProceedingJoinPoint proceedingJoinPoint, Bulk
 		}
 		String backend = backendMonitored.name();
 		io.github.resilience4j.bulkhead.Bulkhead bulkhead = getOrCreateBulkhead(methodName, backend);
-		return handleJoinPoint(proceedingJoinPoint, bulkhead, methodName);
+		RecoveryFunction recovery = backendMonitored.recovery().newInstance();
```
RobWin: Maybe we can create an instance of the RecoveryFunction only when needed. What do you think?

Romeh: I agree with @RobWin  , let us make on need basis and plus can u please update the documentation as well for the same ?

Romeh:
@lespinsideg  Thx for your PR :) , i have some questions : 
The recovery type is forced the follow the recovery function type , now for the following cases : 

1. if the called logic is void ? how this will be optimally covered ?
2- for Reactive flow , in our case with Reactor types , the recovery logic most of the cases need to be properly handled by using for example onErrorResume , can u please add test cases for Spring web flux API with circuit breaker and see how it will be handled ?

Thx a lot in advance

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
```diff
@@ -123,14 +127,16 @@ private Object defaultWebFlux(ProceedingJoinPoint proceedingJoinPoint, io.github
 			if (logger.isDebugEnabled()) {
 				logger.debug("Invocation of method '" + methodName + "' failed!", throwable);
 			}
-			throw throwable;
+
+			return recovery.apply(throwable);
```
RobWin: I think recovery should apply, when the Flux/Mono emits and error. We should use `onErrorResume`
Line 131 does not cover that.

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java
```diff
@@ -61,7 +63,9 @@ public Object bulkheadAroundAdvice(ProceedingJoinPoint proceedingJoinPoint, Bulk
 		}
 		String backend = backendMonitored.name();
 		io.github.resilience4j.bulkhead.Bulkhead bulkhead = getOrCreateBulkhead(methodName, backend);
-		return handleJoinPoint(proceedingJoinPoint, bulkhead, methodName);
+		RecoveryFunction recovery = RecoveryFunctionUtils.getInstance(backendMonitored.recovery());
```
RobWin: Can we move the creation of a RecoveryFunction into the `handleJoinPoint` method and only create an instance when an exception is catched?

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
lespinsideg: I fixed it.

> @lespinsideg Thx for your PR :) , i have some questions :
> The recovery type is forced to follow the recovery function type which has return type by default, now for the following cases :
> 
> * if the called logic is void ? how this will be optimally covered ?
> * for Reactive flow , in our case with Reactor types , the recovery logic most of the cases need to be properly handled by using for example onErrorResume , can u please add test cases for Spring web flux API with circuit breaker and see how it will be handled ?
> 
> Thx a lot in advance

Do you have any good idea for handling void return? In my opinion, `RecoveryFunction<Void>` could be used there.

lespinsideg:
@Romeh 

> @lespinsideg Thx for your PR :) , i have some questions :
> The recovery type is forced to follow the recovery function type which has return type by default, now for the following cases :
> 
> * if the called logic is void ? how this will be optimally covered ?
> * for Reactive flow , in our case with Reactor types , the recovery logic most of the cases need to be properly handled by using for example onErrorResume , can u please add test cases for Spring web flux API with circuit breaker and see how it will be handled ?
> 
> Thx a lot in advance

Do you have any good idea for handling void return? In my opinion, `RecoveryFunction<Void>` could be used there.

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java
lespinsideg: That is a good idea. I moved it to the catch block.

resilience4j-spring/src/main/java/io/github/resilience4j/retry/configure/RetryAspect.java
```diff
@@ -141,15 +143,21 @@ private Object handleSyncJoinPoint(ProceedingJoinPoint proceedingJoinPoint, io.g
...
+	private Object handleAsyncJoinPoint(ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.retry.Retry retry, Class<? extends RecoveryFunction> recoveryFunctionClass, String methodName) throws Throwable {
 		if (logger.isDebugEnabled()) {
 			logger.debug("async retry invocation of method {} ", methodName);
 		}
 		return io.github.resilience4j.retry.Retry.decorateCompletionStage(retry, retryExecutorService, () -> {
 			try {
 				return (CompletionStage<Object>) proceedingJoinPoint.proceed();
 			} catch (Throwable throwable) {
-				throw new CompletionException(throwable);
+				try {
```
RobWin: I think you also want to apply the recovery function when the CompletationStage completes exceptionally, or?
I guess you have to create a `CompletableFuture<T> promise = new CompletableFuture<>();`

and then 

```
decoratedCompletationStage.whenComplete((result, t) -> {
				if (result != null) {
					promise.complete(result);
				} else if (t != null) {
					promise.complete(recovery.apply(t));
				}
			});
return promise;
```

RobWin: Maybe it would be even better to add the recoveryFunction logic directly into the CircuitBreaker, Retry, Bulkhead.

RobWin: Maybe we should create a Recovery Decorator where we can implement it once and use it for Bulkhead, CircuitBreaker, RateLimiter and Retry,

RobWin: For example `Recovery.decorateCompletionStage(recoveryFunction, completationStage)`

lespinsideg: @Romeh I wrote this `RecoveryFunction` to maintain consistency with `RecoveryFunction` in ratpack module. But there is no problem to change it if needed.

Is this what you mean?
```java
@Recover("recoveryMethod")
public void serviceMethod(String paramter1, Integer paratemre2) {}
public void recoveryMethod(String parameter1, Integer parameter2) {}
```

Romeh: I would say just add the annotation over the recovery method itself and in that case u do not need to define method name .
public void serviceMethod(String paramter1, Integer paratemre2) {}
@Recover
public void recoveryMethod(String parameter1, Integer parameter2) {}

@RobWin what do u think ?

RobWin: What happens when someone adds `@Retry` and `@CircuitBreaker` and `@Recover` to a method.
When should the recovery function be applied? 
I think the RecoveryFunction must be attached to a concrete annotation.

RobWin: I'm also still not sure if it's an overuse of annotations.
Why not use plain old java code ;)

123Haynes: imho a seperate annotation is not needed.    
The goal is to make it easier for ex-hystrix users to switch over and to simplify the code users need to write.  
So something like `@CircuitBreaker(name = "backendA", recovery = myRecoveryFunction)`
is more than enough.

Romeh: Yes it is better to attached it to the target annotation by method name and use reflection to execute the method but do not forget to validate that it has the same return type and same list of params as the retryable method

123Haynes: @Romeh I think the recovery method should have one additional parameter.  
The exception that leads to the recovery function being called needs to be passed somehow, so developers can define different behaviors depending on the exception.

Romeh: one last remark , if we plan to add on result predicate check for circuit breaker into the future like the one we have on Retry right now , we need to consider that in our recovery process or not ? as there is no exception being thrown there in that case !

lespinsideg: Good point. I think the recovery function should be applied in that case too. Since that was a failing operation anyway. In that case, the throwable paramter could be null.
@Romeh @RobWin @123Haynes So do you think it is better to implement the `recovery` paramter as a `String` in the annotations and use reflection to call the method? And the recovery method has same signature with the service method except a throwable paratemer? Like

```java
@CircuitBreaker(name = "backendA", recovery = "recovery")
public void service(String param1, Integer param2);
public void recovery(String param1, Integer param2, Throwable throwable);
```

123Haynes: @lespinsideg yeah. I think that would be great 😃

RobWin: Yes, looks good.
I still don't like this overuse of annotations and reflection magic. I would prefer a functional-style solution which can be checked by a compiler.

But we should implement it, if users of R4j request it.

123Haynes: @RobWin I think aspectj is already used for all annotations, so I think the annotations can be checked by the compiler if aspectj is configured to weave things at compile time. 😃

RobWin: Right now we don't weave at compile time.

RobWin: As far as I remember we use Spring AOP and Spring AOP is proxy-based. But it's long time ago when I implemented the CircuitBreakerAspect.

See:
https://docs.spring.io/spring/docs/4.0.1.RELEASE/spring-framework-reference/htmlsingle/#aop-understanding-aop-proxies

123Haynes: @RobWin I've created https://github.com/resilience4j/resilience4j/issues/391 to discuss this further.

lespinsideg:
@RobWin @Romeh @123Haynes 
I implemented it. Would you review the code?

resilience4j-spring/src/test/java/io/github/resilience4j/bulkhead/configure/BulkheadRecoveryTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertThat(testDummyService.observable().blockingFirst()).isEqualTo("recovered");
+    }
+
+    @Test
+    public void testSingleRecovery() {
+        assertThat(testDummyService.single().blockingGet()).isEqualTo("recovered");
+    }
+
+    @Test
+    public void testCompletableRecovery() {
```
Romeh: what is the recovery handling here for Compeltable ? why it should be null ?

resilience4j-spring/src/main/java/io/github/resilience4j/recovery/RecoveryMethod.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class RecoveryMethod {
+    private final Method recovery;
+    private final Object[] args;
+    private final Object target;
+
+    public RecoveryMethod(String recoveryMethodName, Method originalMethod, Object[] args, Object target) throws NoSuchMethodException, ClassCastException {
+        Class[] params = originalMethod.getParameterTypes();
+        Class[] recoveryParams = Arrays.copyOf(params, params.length + 1);
+        recoveryParams[params.length] = Throwable.class;
```
Romeh: Here I think you need to use Class<? extends Throwable> instead as this way you are forcing the user to specify the recovery method with Throwbale type parameter not specific exception type which is bad practice 
So you need to use do the following :
1- use instead Class<? extends Throwable>
2- check if the casue configured into recover method is assignable from it 

Second remark : 
you do not check the accessibility of the recovery method so you are assuming it is public which can be not the case all the time ,so you need to check if it is accessible otherwise make it accessible  :

// check isAccessible
 method.isAccessible();
// No then make it accessible 
ReflectionUtils.makeAccessible(method);

Romeh:
@lespinsideg  Nice PR man ;) . please check my review comments above , happy coding  !

resilience4j-spring/src/main/java/io/github/resilience4j/recovery/RecoveryMethod.java
```diff
@@ -0,0 +1,63 @@
...
+        Method recovery = ReflectionUtils.findMethod(target.getClass(), recoveryMethodName, recoveryParams);
```
Romeh: rename the Method recovery as it is hiding the class property one .

resilience4j-spring/src/main/java/io/github/resilience4j/recovery/RecoveryDecorators.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public RecoveryDecorators(List<RecoveryDecorator> recoveryDecorator) {
+        this.recoveryDecorator = recoveryDecorator;
+    }
+
+    public CheckedFunction0<Object> decorate(RecoveryMethod recoveryMethod, CheckedFunction0<Object> supplier) {
+        return get(recoveryMethod.getReturnType())
+                .decorate(recoveryMethod, supplier);
+    }
+    public RecoveryDecorator get(Class<?> returnType) {
```
Romeh: this can be private i guess

resilience4j-spring/src/test/java/io/github/resilience4j/bulkhead/configure/BulkheadRecoveryTest.java
lespinsideg: Since `Completable.blockingGet()` returns `Throwable` or `null` depends on the terminated result, I wanted to assert that the exception thrown in the method is recovered.

lespinsideg:
@Romeh Thank you for the review. I fixed issues you mentioned and tried to write javadoc. I'm not really good at writing in English, so don't hesitate to correct them.

RobWin:
Your english is as good as ours ;)
We are also no native english speakers.

resilience4j-spring/src/main/java/io/github/resilience4j/recovery/RecoveryMethod.java
```diff
@@ -0,0 +1,151 @@
...
+     *
+     * @return return type of recovery method
+     */
+    public Class<?> getReturnType() {
+        return returnType;
+    }
+
+    private Object invoke(Method recovery, Throwable throwable) throws IllegalAccessException, InvocationTargetException {
+        if (!recovery.isAccessible()) {
+            ReflectionUtils.makeAccessible(recovery);
```
Romeh: here you need to undo the make accessible in finally block something like :
```
boolean methodAccessible = method.isAccessible();
try {
			ReflectionUtils.makeAccessible(method);
			// ur recovery logic 
		}
		finally {
			if (methodAccessible != method.isAccessible()) {
				method.setAccessible(methodAccessible);
			}
}
```

Romeh:
Hey @lespinsideg ,

i have added one more remark about accessible action above  .

Can you please also do the following :

- add documentation about the recovery usage in spring boot section so the users will be aware of it

-  can add the package info with null API support with the @Nullable annotation for recovery classes if possible , check maybe retryConifg for an example 

- for recovery methods extraction , do u think it would be optimal to cache rather than doing the same calculation every-time it is invoked from the aspect  ? it will be more efficient i guess 


Thanks a lot for your nice PR .

lespinsideg:
> * for recovery methods extraction , do u think it would be optimal to cache rather than doing the same calculation every-time it is invoked from the aspect  ? it will be more efficient i guess

@Romeh Do you have any good idea for caching the recovery methods effectively? I think we can only skip filtering methods by its method name by caching the recovery methods.

```java
Map<Class<?>, Method> extractMethods(String recoveryMethodName, Class<?>[] params, Class<?> originalReturnType, Class<?> targetClass)
```

This is a method signature for extracting recovery methods from `targetClass`. There are quite many paramters for extracting specific recovery method. I am not sure we can effectively cache the recovery methods.  Since `ReflectionUtils#declaredMethodsCache` caches all declared methods, I thought extracting recovery methods from that is not so bad. Please let me know if you have any good idea for this.

lespinsideg:
@Romeh Can i make a separate PR for writing document? And always thank you for your detailed review :)

Romeh:
> > * for recovery methods extraction , do u think it would be optimal to cache rather than doing the same calculation every-time it is invoked from the aspect  ? it will be more efficient i guess
> 
> @Romeh Do you have any good idea for caching the recovery methods effectively? I think we can only skip filtering methods by its method name by caching the recovery methods.
> 
> ```java
> Map<Class<?>, Method> extractMethods(String recoveryMethodName, Class<?>[] params, Class<?> originalReturnType, Class<?> targetClass)
> ```
> 
> This is a method signature for extracting recovery methods from `targetClass`. There are quite many paramters for extracting specific recovery method. I am not sure we can effectively cache the recovery methods. Since `ReflectionUtils#declaredMethodsCache` caches all declared methods, I thought extracting recovery methods from that is not so bad. Please let me know if you have any good idea for this.

Hey @lespinsideg  ,

For caching implementation , you can check the following :https://github.com/spring-projects/spring-retry/blob/master/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java

It will explain to you how to do caching by having Method meta data object lookup , if u have any further questions just ping me .

Thx a lot

RobWin:
As soon as caching is solved, I would like to merge this PR.

Romeh:
@lespinsideg  if you are busy those days :)  just inform us i can pick that PR next week and cover the remaining point plus merge conflict resolving . have a nice week end

lespinsideg:
@Romeh i was little bit busy these days. But I can make it tomorrow :)

lespinsideg:
@Romeh I implemented `MethodMeta` and used it as the key of the recovery method cache. And I referenced `java.lang.reflect.Method.hashCode()` for writing `MethodMeta.hashCode()`

resilience4j-spring-boot2/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/BulkheadAutoConfiguration.java
```diff
@@ -38,7 +39,7 @@
 @Configuration
 @ConditionalOnClass(Bulkhead.class)
 @EnableConfigurationProperties(BulkheadProperties.class)
-@Import(BulkheadConfigurationOnMissingBean.class)
+@Import({BulkheadConfigurationOnMissingBean.class,  RecoveryConfiguration.class})
```
Romeh: Here you need to follow the same pattern of condition of missing beans to allow user to override the recovery configuration if he needs so .

resilience4j-spring/src/main/java/io/github/resilience4j/recovery/RecoveryMethod.java
```diff
@@ -0,0 +1,201 @@
...
+ * For example, there are two recovery methods
+ * {@code
+ * String recovery(String parameter, RuntimeException exception)
+ * String recovery(String parameter, IllegalArgumentException exception)
+ * }
+ * and if try to recover from {@link NumberFormatException}, {@code String recovery(String parameter, IllegalArgumentException exception)} will be invoked.
+ * </pre>
+ */
+public class RecoveryMethod {
+    private static final Map<MethodMeta, Map<Class<?>, Method>> RECOVERY_METHODS_CACHE = new ConcurrentReferenceHashMap<>();
```
Romeh: may I ask why ConcurrentReferenceHashMap ? FYI it is not 100% thread safe as i remember !

lespinsideg: I found `ConcurrentReferenceHashMap` is used as a cache in [org.springframework.util.ReflectionUtils](
https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/ReflectionUtils.java). And I thought cached values could be removed if there is no memory space, so I chose `SoftReference`. Since there is no such warning in the document, I didn't know `ConcurrentReferenceHashMap ` is not 100% thread-safe. But should it be a thread-safe guaranteed? Recovery methods are not changing in run-time.

resilience4j-spring-boot2/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/BulkheadAutoConfiguration.java
lespinsideg: I wrote `RecoveryConfigurationOnMissingBean` and added `Import` to each configuration it needed

resilience4j-spring/src/main/java/io/github/resilience4j/recovery/configure/RecoveryConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+		return new ReactorRecoveryDecorator();
+	}
+
+	@Bean
+	public RecoveryDecorator completionStageRecoveryDecorator() {
+		return new CompletionStageRecoveryDecorator();
+	}
+
+	@Bean
+	public RecoveryDecorators recovery(List<RecoveryDecorator> recoveryDecorator) {
```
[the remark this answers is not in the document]
Romeh: @RobWin  it is spring bean configuration , is the new name here will not be a little bit misleading, what do u think  ?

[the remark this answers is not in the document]
RobWin: The comment was a mistake. Forget it.

PR #375, opened 2019-03-25
Adapt prometheus metrics to the new format

voievodin:
The motivation and new naming for micrometer is introduced within this PR https://github.com/resilience4j/resilience4j/pull/355, while here we adapt to the new naming.

During the conversation within the issue #354 it was mentioned that backward compatibility wasn't the main concern, however in this PR we introduce a new set of collectors while deprecating the existing ones, so that library users will still have a chance to update smoothly. I believe this is one of the important principles that we shouldn't be breaking while changing open source libraries.

PR #382, opened 2019-03-28
Use new metrics collectors for spring-boot and spring-boot2

voievodin:
Basically adapting the spring auto configuration to cover the changes introduced in #375 and #355.

The changes within this PR are not backward compatible in a sense that default metrics format is different as described within #355. But there is a back port for those who want to keep using the old naming, it can be controlled by the following properties:

For spring boot 1 (prometheus)
```properties
resilience4j.circuitbreaker.metrics.use_legacy_collector=true
resilience4j.bulkhead.metrics.use_legacy_collector=true
resilience4j.ratelimiter.metrics.use_legacy_collector=true
```

For spring boot 2 (micrometer)
```properties
resilience4j.circuitbreaker.metrics.use_legacy_binder=true
resilience4j.bulkhead.metrics.use_legacy_binder=true
resilience4j.ratelimiter.metrics.use_legacy_binder=true
resilience4j.retry.metrics.use_legacy_binder=true
resilience4j.async_retry.metrics.use_legacy_binder=true
```


RobWin:
Awesome. Thanks

PR #392, opened 2019-04-03
Update Gradle to 5.3.1

123Haynes:
This commit updates gradle to 5.3.1.   
To make the update possible all plugins were updated to the latest version as well.

PR #393, opened 2019-04-03
Update dependencies

123Haynes:
This PR updates the dependencies to the latest versions.  
It is based on https://github.com/resilience4j/resilience4j/pull/392.   
I'll rebase this PR when https://github.com/resilience4j/resilience4j/pull/392 is merged.


Romeh:
@RobWin  also here we need to mention the min version of libraries where resilience4j is tested with in the documentation if we will go with it i guess

123Haynes:
@Romeh Is this currently mentioned somewhere in the docs already? I can update it in that case 😃

RobWin:
Unfortunately no

PR #396, opened 2019-04-03
Changed all external dependencies to compileOnly so that Resilience4j…

RobWin:
… does not introduce too many transitive dependencies.


Romeh:
we will make a new one once fixing all compilation issues

Issue #343, opened 2019-02-22
Behavior in half_open state


hraab:
My primary use case is to open the circuit when backend timeouts occur.

When I use a time limiter chained with a circuit breaker (like in examples) I found that the behavior in HALF_OPEN state is not as I would exptect. When HALF_OPEN, all subsequent requests will go to the backend (and time out) until the circuit breaker switches to OPEN again. I would expect that only 1 request (or configurable amount) will be done in HALF_OPEN state for probing the backend. This is the way Hystrix works. But resilience4j always permits calls in HALF_OPEN state. Therefore all calls will reach reach the time limiter.

Did I miss something? Or is it not possible to get a Hystrik-like behavior? I'm afraid that the latter is true due to the way how resilience4j works (chained decorators which are independend from another).

RobWin:
Hi, thanks for using Resilience4j.

> Did I miss something?
Yes, you missed to read the [implementation details](http://resilience4j.github.io/resilience4j/#_implementation_details) :) 
Resilience4j uses a ring bit buffer to evaluate the failure rate in the HALF_OPEN state. The size of the ring buffer is configurable.  You can configure it to 1, but I would not advice to do so. Imagine you have a backend with multiple instances behind a round-robin load balancer. Just 1 instance could be unreliable which unfortunately answers your only request for probing the backend. The CircuitBreaker switches to OPEN and blocks all calls, even if all other backend instances would respond successfully.

If that's your set-up, better use a `ringBufferSizeInHalfOpenState` >1 and configure your failure rate threshold accordingly.

hraab:
Hi Robert, thanks for your answer.

I set the ring buffer size for HALF_OPEN to 1 (for testing). The problem is: Before the circuit breaker gets a failure, the timeout must occur. The timeout is the "failure" to which the circuit breaker reacts (going from HALF_OPEN to OPEN again). During this period of time all requests will go to the not responding backend (and time out as well).

RobWin:
Ok, I got your point.
That's true. The CircuitBreaker does not reject any further calls yet, when all of the configured probing request have been initiated. That would be an interesting new feature which should be configurable. But it would require to track initiated requests when in HALF_OPEN state.

Would you like to create a PR? But you would have to dip into the internals of the CircuitBreaker.

hraab:
Cannot promise anything. I'll have look into the code.

leonard84:
[...]

PR #402, opened 2019-04-04
WIP:Issue #343: CircuitBreaker should only allow a certain number of test…

RobWin:
… requests in HALF_OPEN and reject calls when maximum number is reached. The number of requests is equal to ringBufferSizeInHalfOpenState.


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerState.java
```diff
@@ -32,7 +32,9 @@
         this.stateMachine = stateMachine;
     }
 
-    abstract boolean isCallPermitted();
+    abstract boolean obtainPermission();
```
leonard84: Having done some C# programming I would actually switch how you implemented them, `tryObtainPermission()` should return `boolean` and `obtainPermission()` should throw an exception.

See https://stackoverflow.com/questions/467613/parse-v-tryparse

RobWin: Yes makes sense.

RobWin: Be aware that the PR is still work in progress. But feel free to review  it

Romeh:
@RobWin  we do not need to follow the same naming for Fixed thread pool bulkhead or this will be done on another PR ?

RobWin:
Same naming, should be done in this PR

PR #408, opened 2019-04-04
Change dependencies to compile only

RobWin:


Romeh:
@RobWin  can u please mention the same on the documentation so the users are aware of that so they can apply the same when the use it ?

Issue #381, opened 2019-03-28
Count long running executions as failure


danielrohe:
Is it possible to count long running tasks (calls to other services) as failure in the circuit breaker. So that if a backend service encounters too many long running tasks the circuit breaker opens.

The long running task should not be cancelled and also not be executed in another thread because then I would have to manage the thread pool and in worst case the limiting factor would be the thread pool. It should only count as failure in the circuit breaker.

RobWin:
Hi,
you mean, the long running tasks finishes successfully, but only runs longer than a certain threshold and hence should count as a failure?

danielrohe:
Yes, that's what I have in mind.

The idea is that if the long running tasks finishes successfully after a certain threshold. This is a sign of service degradation and in the end should result in opening the circuit breaker if it happens too often.

RobWin:
That's not build in, but you measure latency your own and record a failure whenever its longer than a certain treshold.

But we are working on an Adaptive Bulkhead which implements an algorithm comparable to the congestion control algorithm TCP Vegas. The bulkhead would reduce the number of allowed concurrent calls when the latency increases too much or timeouts occur.

Vincent-Zeng:
Hi. I think that the response time threshold above which the CircuitBreaker should trip open would be helpful to me, cause the client expect faster response.

RobWin:
See issue #384 for a possible solution.

RobWin:
I think we will add a `Timer` decorator which can be used to measure the response time of calls.
Additionally it will report call rates and number of successful, failed and total calls.
We will use Micrometer for it.

You can configure a 95th and 98th percentile threshold, and if a percentile threshold is reached you can trigger an action. For example you can open a circuitBreaker or decrease the semaphore count/thread pool of a bulkhead.

RobWin:
The upcoming version of the CircuitBreaker will allow you to configure a duration threshold which will be used to check if a call will be considered as slow.
And you can configure a threshold in percentage which will be used to check if too many calls are slow.

For example: 
slowCallDuration = 5s
slowCallRateThreshold = 75

If 75% or more calls take longer than 5s, the CircuitBreaker will transition to open.
It can be configured as follows:

```
CircuitBreakerConfig.custom()
    .slowCallDurationThreshold(Duration.ofSeconds(4))
    .slowCallRateThreshold(75)
    .build().
```

PR #414, opened 2019-04-07
Result predicate function for circuitbreaker

### LINKED ISSUE #384: CircuitBreaker record a failure when a certain result is returned
`Retry` allows check the result of a call and retry a call when a configurable predicate is true.
It would be nice if the CircuitBreaker allows it as well.
`CircuitBreakerConfig` should allow to configure a result predicate.

lespinsideg:
Resolves #384 

- added result predicate function in execute and decorate methods.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
@@ -181,16 +181,40 @@ public interface CircuitBreaker {
     }
 
     /**
+     * Decorates and executes the decorated Supplier.
+     *
+     * @param supplier the original Supplier
+     * @param resultPredicate result predicate by duration and return object function.
+     * @param <T> the type of results supplied by this supplier
+     * @return the result of the decorated Supplier.
+     */
+    default <T> T executeSupplier(Supplier<T> supplier, Function2<Long, T, Boolean> resultPredicate){
+        return decorateSupplier(this, supplier, resultPredicate).get();
+    }
+
+    /**
      * Decorates and executes the decorated Callable.
      *
      * @param callable the original Callable
-     *
-     * @return the result of the decorated Callable.
      * @param <T> the result type of callable
+     * @return the result of the decorated Callable.
      * @throws Exception if unable to compute a result
      */
     default <T> T executeCallable(Callable<T> callable) throws Exception{
         return decorateCallable(this, callable).call();
+    }
+
+    /**
+     * Decorates and executes the decorated Callable.
+     *
+     * @param callable the original Callable
+     * @param resultPredicate result predicate by duration and return object function.
+     * @param <T> the result type of callable
+     * @return the result of the decorated Callable.
+     * @throws Exception if unable to compute a result
+     */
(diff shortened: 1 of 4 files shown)


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -556,13 +709,32 @@ static CheckedRunnable decorateCheckedRunnable(CircuitBreaker circuitBreaker, Ch
...
+    static <T> Consumer<T> decorateConsumer(CircuitBreaker circuitBreaker, Consumer<T> consumer, Function1<Long, Boolean> resultPredicate){
         return (t) -> {
             CircuitBreakerUtils.isCallPermitted(circuitBreaker);
             long start = System.nanoTime();
             try {
                 consumer.accept(t);
                 long durationInNanos = System.nanoTime() - start;
-                circuitBreaker.onSuccess(durationInNanos);
+
+                if (resultPredicate.apply(durationInNanos)) {
```
Romeh: @lespinsideg  the result predicate take the start time any reason why ? as u can simply do the calculation inside the if/else statement blocks .

RobWin:
Thank you for your PR.
But I'm not sure if we should add a `resultPredicate` to all methods.
In `RetryConfig` we added an option to configure a global `resultPredicate`.  But I actually prefer higher-order functions.

If someone wants to be able to have a different resultPredicate per decorated function, he could easily use a higher-order function.

```
static <T> Supplier<T> checkResult(Supplier<T> supplier, Function<T, T> exceptionMapper){
        return () -> {
            T result = supplier.get();
            return exceptionMapper.apply(result);
        };
}
```

and then something like

```
        Supplier<HttpResponse> httpResponseSupplier = checkResult(helloWorldService::returnResult, result -> {
            if (result.getStatusCode() == 400) {
                throw new ClientException();
            } else if (result.getStatusCode() == 500) {
                throw new ServerException();
            }
            return result;
        });
        //When
        HttpResponse httpResponse = circuitBreaker.executeSupplier(httpResponseSupplier);
```

storozhukBM:
Agree with @RobWin

storozhukBM:
Sorry, closed by mistake.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
lespinsideg: @Romeh Sorry but I didn't get your point. The result predicate doesn't take the start time. I thought duration could be a parameter for deciding the result.
https://github.com/resilience4j/resilience4j/issues/381

lespinsideg:
@RobWin @Romeh @storozhukBM 
I will check `RetryConfig` and update the code soon. Thanks for the review.

RobWin:
I really think it was a mistake to add a onResult predicate to Retry. A decorator would have been the better choice. A decorator could have been reused for CircuitBreaker as well.

RobWin:
We added it to Retry, because we didn't want to throw an exception every time we retry. But in our CirucirBreaker we need an exception to be thrown.

PR #415, opened 2019-04-08
Issue #384: Added higher-order functions to chain resultHandler and e…

RobWin:
…xceptionHandler to supplier and callable.

### DIFF
--- a/resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
+++ b/resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
@@ -25,8 +25,8 @@ import java.util.concurrent.ConcurrentMap;
 
 public class EventProcessor<T> implements EventPublisher<T> {
 
-    protected volatile boolean consumerRegistered;
-    @Nullable private volatile EventConsumer<T> onEventConsumer;
+    private boolean consumerRegistered;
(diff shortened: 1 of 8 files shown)


Romeh:
Maybe we need to add some documentation ? :)

resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
```diff
@@ -25,16 +25,16 @@
...
+    @Nullable private EventConsumer<T> onEventConsumer;
     private ConcurrentMap<Class<? extends T>, EventConsumer<Object>> eventConsumers = new ConcurrentHashMap<>();
 
     public boolean hasConsumers(){
         return consumerRegistered;
     }
 
     @SuppressWarnings("unchecked")
-    public <E extends T> void registerConsumer(Class<? extends E> eventType, EventConsumer<E> eventConsumer){
+    public synchronized <E extends T> void registerConsumer(Class<? extends E> eventType, EventConsumer<E> eventConsumer){
```
Romeh: why u need it synchronized ?

RobWin: It was a SonarCloud warning. 
The method `onEvent` changes two fields, but both fields are only volatile. This can cause problems.
The method `registerConsumer` registerConstumer also updates two fields.

PR #416, opened 2019-04-08
WIP: Use tags instead of gauge values to indicate circuit breaker state

leonard84:
Implement change as discussed in #354


leonard84:
I don't know, but this doesn't really seem to be related to my change.
```
$ sonar-scanner
sonar-scanner: command not found
The command "sonar-scanner" exited with 127.
```

It also executes and fails on my fork since the secure variables are not available. The execution should run in a preview mode or something similar if no credentials are available.

Romeh:
@leonard84  you can ignore that for now we are working right now on upgrading our code coverage tooling .

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/CircuitBreakerMetrics.java
```diff
@@ -71,10 +71,15 @@ public static CircuitBreakerMetrics ofIterable(String prefix, Iterable<CircuitBr
 
     @Override
     public void bindTo(MeterRegistry registry) {
+        final CircuitBreaker.State[] states = CircuitBreaker.State.values();
         for (CircuitBreaker circuitBreaker : circuitBreakers) {
             final String name = circuitBreaker.getName();
-            Gauge.builder(getName(prefix, name, STATE), circuitBreaker, (cb) -> cb.getState().getOrder())
-                    .register(registry);
+            for (CircuitBreaker.State state : states) {
+                Gauge.builder(getName(prefix, name, STATE), circuitBreaker, (cb) -> cb.getState() == state ? 1 : 0)
```
RobWin: Why do we need a gauge metric per state?
Isn't the current state enough? The metric value could still be `cb.getState().getOrder()`
The query which counts the number of CircuitBreakers with label state==open should still be possible or?

RobWin: It would be backward compatible but add more query possibilities.

RobWin:
Could you rebase your changes?
I would like to merge it.

PR #424, opened 2019-04-12
Jdk 11 support

RobWin:


123Haynes:
@RobWin you need to rebase this on top of the current master and change
`spockVersion = '1.3-RC1-groovy-2.4'`
to
`spockVersion = '1.3-groovy-2.5'`

after that that java 11 build should work 😃

RobWin:
I try ;)

123Haynes:
@RobWin That worked 👍   
But the travis config seems to be a bit weird. Why are there 2 java 8 jobs in the test stage and one is failing while the other one is passing?

RobWin:
Yes, I'm also reading TravisCI docs at the moment. I guess it's the default TravisCI test stage `gradlew check` which is running twice. And my defined test stage is running as a third job.

PR #425, opened 2019-04-12
Fixed SonarCloud findings

RobWin:


resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java
```diff
@@ -59,12 +59,15 @@ public void matchAnnotatedClassOrMethod(Bulkhead Bulkhead) {
...
 
 	@Around(value = "matchAnnotatedClassOrMethod(backendMonitored)", argNames = "proceedingJoinPoint, backendMonitored")
-	public Object bulkheadAroundAdvice(ProceedingJoinPoint proceedingJoinPoint, Bulkhead backendMonitored) throws Throwable {
+	public Object bulkheadAroundAdvice(ProceedingJoinPoint proceedingJoinPoint, @Nullable Bulkhead backendMonitored) throws Throwable {
 		Method method = ((MethodSignature) proceedingJoinPoint.getSignature()).getMethod();
 		String methodName = method.getDeclaringClass().getName() + "#" + method.getName();
 		if (backendMonitored == null) {
 			backendMonitored = getBackendMonitoredAnnotation(proceedingJoinPoint);
 		}
+		if(backendMonitored == null) { //because annotations wasn't found
```
Romeh: is that even possible to happen , to have the aspect invoked despite no annotation ?

RobWin: No don't think so.

RobWin: But the parameter is null... always... don't understand why.

RobWin: That's why we have to invoke `getBackendMonitoredAnnotation`

PR #426, opened 2019-04-12
Add the Automatic-Module-Name for each project.

123Haynes:
Until proper support for modules is introduced, this PR adds the appropriate Automatic-Module-Name to every produced .jar file.  
For Java 8 this does not matter, but it helps people who are already using the JPMS introduced in Java 9.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,6 +106,17 @@ subprojects {
             property "sonar.jacoco.reportPaths", allTestCoverageFile
         }
     }
+    afterEvaluate {
+        // exclude subprojects that don't produce a jar file.
+        if(!project.name.equals('resilience4j-bom') && !project.name.equals('resilience4j-documentation') && !project.name.equals('resilience4j-test')) {
+            jar {
+                inputs.property('moduleName', moduleName)
+                manifest.attributes(
+                    'Automatic-Module-Name': moduleName
+                )
+            }
+        }
+    }
 }
 
 tasks.check.dependsOn tasks.jacocoTestReport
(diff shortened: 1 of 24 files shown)


RobWin:
I started a while ago to create manifest files in a separate feature branch. We can use it as a start to introduce proper modules.

RobWin:
https://github.com/resilience4j/resilience4j/commit/4c9bfe43680dd3a636392e525746288f52f28bcf

123Haynes:
@RobWin I saw that. But in order to keep supporting java 8, we'd need to create multi release jar files and at the moment not all dependencies of resilience4j support the module system already.  
So proper support will likely take a while to implement.  
But the changes in this PR allow people to at least use resilience4j with the module system without changing anythign for java 8.

RobWin:
Yes I merge it also without module info files. I still think, since Java8 is soon out of supports, it's time to add java modules.

123Haynes:
@RobWin I created https://github.com/resilience4j/resilience4j/issues/427 to track what's necessary for full module support.

RobWin:
I would prefer to have a shorter module name like `resilience4j.`. Please see my commit.

123Haynes:
@RobWin I chose the longer name because that's how the [JLS §6.1, Declarations](https://docs.oracle.com/javase/specs/jls/se10/html/jls-6.html#jls-6-170) defines it:

>The name of a module should correspond to the name of its principal exported package. If a module does not have such a package, or if for legacy reasons it must have a name that does not correspond to one of its exported packages, then its name should still start with the reversed form of an Internet domain with which its author is associated.

>**Example 6.1-2. Unique Module Names**
```
com.nighthacks.scrabble
org.openjdk.compiler
net.jcip.annotations
```

See also https://stackoverflow.com/questions/50736107/naming-convention-for-java-9-modules

with the longer name it's also more clear which module to choose if someone forks resilience4j and releases their own version.

PR #432, opened 2019-04-24
Common registry functionality update

Romeh:
Still need to add documentation update .

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/InMemoryBulkheadRegistry.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/InMemoryBulkheadRegistry.java
@@ -49,6 +49,7 @@ public final class InMemoryBulkheadRegistry extends AbstractRegistry<Bulkhead, B
 	 * @param bulkheadConfig custom bulkhead config to use
 	 */
 	public InMemoryBulkheadRegistry(BulkheadConfig bulkheadConfig) {
+		super();
 		this.defaultBulkheadConfig = bulkheadConfig;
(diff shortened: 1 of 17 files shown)


resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/InMemoryBulkheadRegistry.java
```diff
@@ -18,68 +18,69 @@
...
+				k -> notifyPostCreationConsumers(Bulkhead.of(name, bulkheadConfigSupplier.get()))
+		);
+	}
 
-    @Override
-    public BulkheadConfig getDefaultBulkheadConfig() {
-        return defaultBulkheadConfig;
-    }
+	@Override
+	public BulkheadConfig getDefaultBulkheadConfig() {
```
RobWin: Can we deprecate this method and add a `getDefaultConfig` into `AbstractRegistry`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfigurationNotFound.java
```diff
@@ -0,0 +1,33 @@
...
+ *  limitations under the License.
+ *
+ *
+ */
+package io.github.resilience4j.circuitbreaker;
+
+/**
+ * configuration not found exception
+ */
+public class CircuitBreakerConfigurationNotFound extends RuntimeException {
```
RobWin: Can we reanme it to `ConfigurationNotFoundException` and move it into resilience4j-core.
We can reuse it for other Registries as well.

Issue #417, opened 2019-04-09
Allow fully closing a bulkhead


jguerra:
BulkheadConfig.Builder currently guards against setting maxConcurrentCalls to 0. It would be useful to be able to completely close a bulkhead by setting maxConcurrentCalls to 0

RobWin:
Hi,

can you please explain when do you need it? Maybe a CircuitBreaker is better suited in front of a Bulkhead?

RobWin:
Will be part of v0.15.0

PR #433, opened 2019-04-24
Issue #417: Allow to set maxConcurrentCalls to zero in order to close…

RobWin:
… a bulkhead fully.


Romeh:
maybe one unit test for changing the configuration from 1 to 0 and see if it is well respected as it should be

PR #323, opened 2019-01-17
Created resilience4j-kotlin module with circuitbreaker

gmariotti:
This is an example for #322. If is it okay for you, I'll add proper documentation and copyright and look into other cases.

Issue #322, opened 2019-01-17
Support suspending functions


gmariotti:
Hi team, AFAIK, the current decorator functions don't support Kotlin suspending functions. I'd like to contribute with a `resilence4j-kotlin` module, where these small cases are handled. Would you be interested in it?

storozhukBM:
Hi @gmariotti 
Yes, I think such module would be a nice addition to our library.

RobWin:
Any progress on this issue?

gmariotti:
Hi @RobWin, my MR is still blocked by the Gradle version, that is why I didn't do any more work on it

RobWin:
I think we should try to update gradle and all of this plugins.
I think the last time I tried it, I just had troubles with `gradle-git` which we use to deploy to GitHub Pages.
We should be able to fix it.

RobWin:
I give gradle  5.3 a try.

gmariotti:
5.3 might be too new for some plugins, if it doesn't work, 4.10.3 might be a good enough starting point

RobWin:
Alright

RobWin:
Merged

PR #441, opened 2019-04-29
Add resilience4j-kotlin module

bradnewman:
At least partially addresses #322: adds support for suspend functions for CircuitBreaker, Retry, TimeLimiter, and RateLimiter.  Based on #323 - thanks to @gmariotti.  I haven't (yet?) added support for Caches or Bulkheads.  

I haven't used Resilience4j's caches before, and I'm honestly slightly puzzled by them:
 * Is there no guarantee that multiple concurrent calls to `computeIfAbsent` for the same key only result in a single successful call to the `Supplier`?
 * Shouldn't the decorate series of functions operate on `Function<K, R>` rather than `Supplier<R>` so that the work to compute the value can be based on the key?

And I don't think Bulkheads can be supported well, at least without changes to the bulkhead core module:
 * The concept of a `ThreadPoolBulkhead` that limits concurrent operations via a thread pool with a maximum size is incompatible with the idea of suspending functions - when the coroutine suspends, the thread is freed to execute another coroutine (returned to the pool), even though the suspend function hasn't completed yet.  So at best you could limit the number of actively executing coroutines, rather than the number that have been started and haven't finished yet.  I think that's probably too confusing and not useful enough to warrant supporting.
 * The semaphore-based bulkheads do not offer any asynchronous way to wait for a permission, and I don't think a way can practically be added while using a `j.u.concurrent.Semaphore`.  So we could block (rather than suspend) the coroutine while we wait for permission, but that rather defeats the purpose of using coroutines in the first place.


RobWin:
Hi,
thank you for your PR.
I agree `ThreadPoolBulkhead` does not make sense for coroutines.

The Semaphore-based Bulkhead can be configured with `maxWaitTime=0` which means it is non-blocking. This is how we use the Bulkhead in RxJava2 or Reactor.

The cache decorator works as follows:

You can decorate any `Supplier` function as follows:

`Function<String, Result> cachedFunction = Cache.decorateSupplier(cacheContext, () -> service.doSomething(param1, param2));`

The decorator returns a function. You can invoke the function and provide the key which should be used.

`Result result = cachedFunction.apply("key");`

The result is stored in the cache and a subsequent call returns the result from a cache.

bradnewman:
Okay, thanks for the clarification.  Cache and non-blocking semaphore
bulkheads should be straight forward - I'll add them.  It looks like
RxJava2 and Reactor just ignore any configured max wait time (rather than
enforcing it must be set to 0)?  I will make Kotlin consistent in that
regard.

bradnewman:
I've added commits with support for Bulkheads.  But upon further review of Cache, I don't think it's feasible to add support for suspend functions without changes to the core module.  

Specifically, we would need to expose `getValueFromCache()` and `putValueIntoCache()` (or equivalent methods) and essentially re-implement `computeIfAbsent()` in the context of the suspending extension function.

fvigotti:
this post is stale since 28 days.. can I say +1 to fix & merge ? :) :+1:

RobWin:
Yes, we can merge it after the conflict has been solved.

PR #442, opened 2019-04-29
Moved duplicated files into Spring Boot Commons.

RobWin:
Refactored CircuitBreakerConfigurationProperties and CircuitBreakerConfiguration


resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerConfiguration.java
```diff
@@ -90,19 +85,29 @@ public ReactorCircuitBreakerAspectExt reactorCircuitBreakerAspect() {
...
+
+		return CircuitBreakerRegistry.of(configs);
+	}
 
+	/**
+	 * Initializes the CircuitBreaker registry.
+	 *
+	 * @param circuitBreakerRegistry The circuit breaker registry.
+	 */
+	public void initCircuitBreakerRegistry(CircuitBreakerRegistry circuitBreakerRegistry) {
```
Romeh: we can not the back-ends the same away as above in  CircuitBreakerRegistry bean init ?

PR #443, opened 2019-05-01
Mark success on cancel for Reactor

making:
This commit marks success for circuit breaker when cancel of Reactor stream is called.

My use case is Spring WebFlux + resilience4j. When a `Mono` is successfully returned from a controller, `cancel` is called instead of `onComplete`.

Let's see the easiest example:

```java
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;
import reactor.core.publisher.Mono;

@RestController
public class HelloController {

    @GetMapping
    public Mono<String> hello() {
        return Mono.just("Hello World!").log("hello");
    }
}
```

Log is following:

```
2019-05-02 03:28:56.032  INFO 46255 --- [ctor-http-nio-6] hello : | request(unbounded)
2019-05-02 03:28:56.032  INFO 46255 --- [ctor-http-nio-6] hello : | onNext(Hello World!)
2019-05-02 03:28:56.035  INFO 46255 --- [ctor-http-nio-6] hello : | cancel()
```

`onComplete` is called when a `Flux` is successfully returned:

```java
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RestController;
import reactor.core.publisher.Flux;

@RestController
public class HelloController {

    @GetMapping
    public Flux<String> hello() {
        return Flux.just("Hello World!").log("hello");
    }
}
```

```
2019-05-02 03:38:17.684  INFO 48014 --- [ctor-http-nio-6] hello : | request(1)
2019-05-02 03:38:17.684  INFO 48014 --- [ctor-http-nio-6] hello : | onNext(Hello World!)
2019-05-02 03:38:17.691  INFO 48014 --- [ctor-http-nio-6] hello : | request(32)
2019-05-02 03:38:17.692  INFO 48014 --- [ctor-http-nio-6] hello : | onComplete()
```

This means, if `Mono` is returned with `CircuitBreakerOperator` (as shown in [the doc](https://resilience4j.github.io/resilience4j/#_circuitbreaker_and_reactor)) from a WebFlux controller, the state never get back to `CLOSED` once it goes to `OPEN`.
A current workaround to get it working is using `Flux` instread of `Mono`.

(I've never tried bulkhead, but `BulkheadSubscriber` seems to have same problem.)

### DIFF
--- a/resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/circuitbreaker/operator/CircuitBreakerSubscriber.java
+++ b/resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/circuitbreaker/operator/CircuitBreakerSubscriber.java
@@ -21,6 +21,7 @@ import io.github.resilience4j.core.StopWatch;
 import io.github.resilience4j.core.lang.Nullable;
 import io.github.resilience4j.reactor.ResilienceBaseSubscriber;
 import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
 import reactor.core.CoreSubscriber;
 
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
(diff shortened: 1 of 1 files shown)


RobWin:
Hi, could you create an issue as well.
I personally think it's an issue in WebFlux. Mono should complete with `onComplete`.

RobWin:
I actually can't reproduce it in our Spring Boot 2 demo.

making:
Thank you for confirmation.
Maybe my sample's issue. I'll dig into the difference.

RobWin:
Not sure, because you are right that cancel is invoked which is strange.
We should ask @smaldini

Btw, your yavi library looks very nice.

PR #446, opened 2019-05-03
Circuitbreaker predicate config

Romeh:


RobWin:
I solved this differently now. In my upcoming PR, you arw able to overwrite any part (ignoredExceptions, recordExceptions or failurePredicate) and the final predicate is build when the config build . I show it to you in my upcoming PR.

PR #447, opened 2019-05-07
add fallbackMethod for ratpack annotations

drmaas:
@RobWin circuitbreaker example - once good can be extended to other components like retry etc.


RobWin:
The Spring Boot solution is a little bit more powerful. You can define multiple methods with the same `fallbackMethod` name, but with different signatures. The new class `RecoveryMethod` can find and use the method with the best signature match at runtime.

```
@CircuitBreaker(name = BACKEND, fallbackMethod = "recovery")
public String method(String parameter) {
        throw new NumberFormatException();
}

private String recovery(String parameter, RuntimeException exception) {
        return "recovered-RuntimeException";
}

private String recovery(String parameter, IllegalArgumentException exception) {
        return "recovered-IllegalArgumentException";
}
```

The following would use `recovery` with `IllegalArgumentException exception`, since `NumberFormatException` extends `IllegalArgumentException`

RobWin:
We can move `RecoveryMethod`, `RecoveryDecorators` and all implementations like `CompletionStageRecoveryDecorator` into a common module so that we can reuse it.
(As soon as the Spring Boot PR is merged)

drmaas:
@RobWin it should match the Spring Boot solution now.

drmaas:
@RobWin PTAL

Issue #311, opened 2019-01-01
Rate limiting implementation for reactor is broken


thekalinga:
Please ignore this completey, Just look at below comment as it makes more sense than this

~~Rate limiter is supposed to protect a resource against unexpected number of calls like lets say number of times an external API can called/sec~~

~~Meaning, we guard the actual calls, not the results from external call~~

~~In our current Reactor implementation (I assume the same is true for RxJava2 aswell), we are rate limiting the downstream results, not the actual upstream subscription (Refer to `MonoCallable#subscribe`)~~

~~Rate limiter should limit the number of time a method is called, so, for `Mono.fromCallable` the upstream subscription should be delayed (using `delaySubscription`) so that the actual method under rate limit never gets executed till a rate limit token is available~~

~~In the current implementation~~

~~`ResilienceBaseSubscriber#onSubscribe` - is a blind request to upstream without any rate limit
`RateLimiterSubscriber#hookOnNext` - we are rate limiting downstream elements aswell, which is wrong completely, as we should never care about how many elements upstream is producing, we are interested only in rate limiting the upstream calls~~

~~For rate limiting downstream elements within a `Flux`, we have `Flux.limitRate` & `Flux.limitRequest`~~

~~So the current limiter is completely broken~~

thekalinga:
Looks like I'm wrong about half of my analysis. i.e about rate limiting subscription

Looks like the current implementation of `ResilienceBaseSubscriber#isCallPermitted()` is a blocking operation. So all the calls(+ threads) if any of the threads use the same rate limiter will be blocked (blocking thread is not a good idea to begin with) when rate limiter does not have enough tokens to issue

Also, no one asked for limiting downstream events aswell along with upstream subscription

So, the downstream elements are also rate limited for some unknown reason

There are a total of three aspects to rate limit in a flux

1. Intercepting upstream subscription (applicable both to Flux & Mono)
2. Intercepting upstream requests (applicable only to Flux)
3. Intercepting downstream elements (applicable only to Flux)

And each of them should be opt-in. Currently (2) is not at all considered as part of rate limiting, (1) & (3) come in a single package (they are no opt-in), that too by blocking thread

So, the current implementation handles none of the three scenarios correctly

storozhukBM:
Hi @thekalinga,
Thank you for bringing up this conversation. I think we should wait for @madgnome for comment.
[...]

Issue #361, opened 2019-03-14
CircuitBreaker does not prevent execution of upstream Mono if flatMap is used


leonard84:
### Expected behavior
Both tests `simple`, `withMap`, and `withFlatMap` behave the same.

### Actual behavior
In `simple` and `withMap` `getNext()` is called twice (as expected) and `withFlatMap` `getNext()` is called four times.

Log for `simple`/`withMap` test
```
Requested Counter: 1
16:03:19.461 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event SUCCESS not published
Ctr: 1
Requested Counter: 2
16:03:19.476 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - CircuitBreaker 'test' recorded a failure:
java.lang.RuntimeException: Error
	...
16:03:19.477 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event ERROR not published
16:03:19.494 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event STATE_TRANSITION not published
RuntimeException
16:03:19.496 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event NOT_PERMITTED not published
CircuitBreakerOpenException
16:03:19.497 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event NOT_PERMITTED not published
CircuitBreakerOpenException
```

Log for `withFlatMap` test
```
Requested Counter: 1
16:03:19.601 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event SUCCESS not published
Ctr: 1
Requested Counter: 2
16:03:19.603 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - CircuitBreaker 'test' recorded a failure:
java.lang.RuntimeException: Error
	...
16:03:19.603 [main] DEBUG io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine - No Consumers: Event ERROR not published
[...]

Issue #357, opened 2019-03-11
More than one circuit breaker for same backend results in "Operator called default onErrorDropped"


nischit7:
First of all, this is not a straight forward issue.
It can occur under specific conditions.

I do have a business requirement where I have to define 2 circuit breakers config for the same downstream API. This back end is very unstable and a specific endpoint is error prone. Hence we are having 2 different circuit breaker level to abort further class.

From the below test case, we have a requirement where in for a collection of targets, we call this unstable API endpoint.

```
import java.time.Duration;
import java.util.Arrays;
import java.util.List;
import java.util.function.Function;

import org.reactivestreams.Publisher;
import org.testng.annotations.BeforeClass;
import org.testng.annotations.Test;

import io.github.resilience4j.circuitbreaker.CircuitBreaker;
import io.github.resilience4j.circuitbreaker.CircuitBreakerConfig;
import io.github.resilience4j.circuitbreaker.CircuitBreakerOpenException;
import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
import io.github.resilience4j.reactor.circuitbreaker.operator.CircuitBreakerOperator;

import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;

public class CircuitBreakerSubscriberTest {

    private CircuitBreakerRegistry  circuitBreakerRegistry;

    @BeforeClass(alwaysRun = true)
    public void before() {
        this.circuitBreakerRegistry = CircuitBreakerRegistry.ofDefaults();
    }

   // With two circuit breakers
    @Test
    public void whenSubscriptionCancelled() {

        List<Integer> list = Arrays.asList(1);
        for (int index = 1; index < 10; index ++) {
            Flux.fromStream(list.stream())
                    .flatMap(integer -> Mono.just(integer)
                            .flatMap(emittedIndex -> {
                                return callOutboundApi();
                            })
                            .transform(decorateWithCircuitBreakerForSpecificEndpoint())
                            .map(str -> "Success.... sending good response")
                            .onErrorResume(NullPointerException.class, ex -> {
                                ex.printStackTrace();
                                return Mono.just("some error occured, sending error response " + ex);})
                            .onErrorResume(CircuitBreakerOpenException.class, ex -> {
                                ex.printStackTrace();
                                return Mono.just("some error occured, sending error response " + ex);})
                            .map(emittedObject -> {
                                // do some further business logic;
                                return "modified object, after applying further business logic";
                            }))
[...]

PR #453, opened 2019-05-10
#311 Fix reactor rate limiter 

madgnome:
- rate limiting will only happen on subscription and not on emitting items
- rate limiting is now non blocking


RobWin:
Why can't we just fix the RateLimiterSubscriber?  RateLimiterSubscriber already tries to obtain a permission for subscription. Instead of `getPermission` we can use `reservePermission`.
The ResilienceBaseSubscriber already cancels the subscription if the ratelimiter rejects the subscription and forwards an EmptySubscription.INSTANCE

RobWin:
If you look at FluxLimitRequest I wonder if we also have to rate limit `request()`

RobWin:
My fix from yesterday solved a bug in ResilienceBaseSubscriber.onSubscribe. I think we just have to fix hookOnNext, hookOnComplete and hookOnComplete in RateLimiterSubscriber.

RobWin:
Ok, I see. You want to delay the subscription.

RobWin:
Do we have to delay the subscription or just the `resquest` from downstream to upstream.

RobWin:
Maybe we should keep on using getPermission, but with waitDuration = 0.
A delayed subscription might be not what users expect.

madgnome:
I liked your proposition of rate limiting request() instead so I did that.

RobWin:
I refactored r4j-reactor and r4j-rxjava2 in order to fix the open issues: #361 , #311, #357

PR #460, opened 2019-05-21
ratpack config updates, commonize framework configs

drmaas:


drmaas:
@RobWin ptal - this will take a bit more time to complete for all the other components.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/ratelimiter/RateLimiterConfigurationProperties.java
```diff
@@ -0,0 +1,225 @@
...
+import com.google.common.base.Strings;
+import io.github.resilience4j.core.ConfigurationNotFoundException;
+import io.github.resilience4j.core.lang.Nullable;
+import io.github.resilience4j.ratelimiter.RateLimiterConfig;
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+
+public class RateLimiterConfigurationProperties {
```
RobWin: Can we share the code between Ratpack and Spring Boot and move it into resilience4j-framework-common?

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/circuitbreaker/CircuitBreakerConfigurationProperties.java
```diff
@@ -0,0 +1,333 @@
...
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import java.lang.reflect.Constructor;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+public class CircuitBreakerConfigurationProperties {
```
RobWin: Can we share the code between Ratpack and Spring Boot and move it into resilience4j-framework-common?

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadConfigurationProperties.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.google.common.base.Strings;
+import io.github.resilience4j.core.ConfigurationNotFoundException;
+import io.github.resilience4j.core.lang.Nullable;
+
+import javax.validation.constraints.Min;
+import java.util.HashMap;
+import java.util.Map;
+
+public class BulkheadConfigurationProperties {
```
RobWin: Can we share the code between Ratpack and Spring Boot and move it into resilience4j-framework-common?

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/ratelimiter/RateLimiterConfigurationProperties.java
drmaas: @RobWin there are a few framework differences regarding AOP ordering, BeanUtils, StringUtils, etc. The inner class `BackendProperties` appears to be similar as well, but in ratpack the setters use a builder pattern and in spring they don't. We can extract out the common stuff, but need to decide on 1 way to do things.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/circuitbreaker/CircuitBreakerConfigurationProperties.java
drmaas: Same

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/bulkhead/BulkheadConfigurationProperties.java
drmaas: Same

drmaas:
@RobWin I'll take a look at extracting common code

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/ratelimiter/RateLimiterConfigurationProperties.java
drmaas: For example:

```
        public BackendConfig baseConfig(String baseConfig) {
            this.baseConfig = baseConfig;
            return this;
        }
```

vs. 

```
		public void setBaseConfig(String baseConfig) {
			this.baseConfig = baseConfig;
		}
```

RobWin: Ok, I see. I think we should remove StringUtils, BeanUtils anyway. 
Micronaut does not have them by default as well.

RobWin: Can Ratpack use a Builder pattern with a `set` prefix?
Micronaut allows it.

drmaas:
@RobWin commonized bulkhead ... if it looks good can do the rest.

RobWin:
LGTM

drmaas:
@RobWin ptal

resilience4j-core/src/main/java/io/github/resilience4j/core/StringUtils.java
```diff
@@ -0,0 +1,11 @@
+package io.github.resilience4j.core;
+
+import io.github.resilience4j.core.lang.Nullable;
+
+public class StringUtils {
+
+    public static boolean isNullOrEmpty(@Nullable String string) {
```
RobWin: Can we revert the logic to `isNotEmpty`.
It's usally used like `!StringUtils.isNullOrEmpty`.
I would prefer `StringUtils.isNotEmpty`

resilience4j-core/src/main/java/io/github/resilience4j/core/ClassUtils.java
```diff
@@ -0,0 +1,24 @@
...
+    public static <T> Predicate<T> instantiatePredicateClass(Class<? extends Predicate<T>> clazz) {
+        try {
+            Constructor<? extends Predicate> c = clazz.getConstructor();
+            if (c != null) {
+                return c.newInstance();
+            } else {
+                return null;
+            }
+        } catch (Exception e) {
+            throw new RuntimeException("Unable to create instance of class: " + clazz.getName());
```
RobWin: Can we create a custom `InstantiationException` which is a RuntimeException?

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/ratelimiter/configuration/RateLimiterConfigurationProperties.java
```diff
@@ -0,0 +1,230 @@
...
+import io.github.resilience4j.core.lang.Nullable;
+import io.github.resilience4j.ratelimiter.RateLimiterConfig;
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+
+public class RateLimiterConfigurationProperties {
+
+	private Map<String, BackendProperties> backends = new HashMap<>();
```
RobWin: Unfortunately this breaks backward compatibility.
The rateLimiter used unfortunately `limiters` instead of `backends`.
I really don't like the name `backends` anyway. Maybe we should rename it to `instances`.
For backward compatibility, we could provide both options. 
What do you think?

drmaas: @RobWin yeah - instances works. We can provide both for limiters. Do we need only `instances` for the others?

RobWin: I guess we just need a getter like getBackends() which returns the instances Map

RobWin:
Could you please merge the latest changes? We added support for thread-pool bulkheads in Spring. Could this be interesting for Ratpack as well?

drmaas:
@RobWin ptal

RobWin:
Alright

PR #461, opened 2019-05-21
Add the Threadpool Bulkhead Spring Support 

Romeh:


resilience4j-spring-boot2/src/test/resources/application.yaml
```diff
@@ -88,6 +88,12 @@ resilience4j.bulkhead:
       maxConcurrentCall: 1
     backendB:
       baseConfig: default
+    backendC:
+      threadPoolProperties:
```
RobWin: Not sure about this.
I think we need a new namespace like `resilience4j.threadpoolbulkhead` so that a user can also define a default configuration. 
In this way a user could also use the same name for both bulkheads. If he wants to use the backend name for both bulkhead heads, it would not be possible at the moment.

RobWin:
I was wondering if we should add an option to rename the threads used by our TreadhPoolBulkhead, or rename them automatically and use the bulkhead name for it.

resilience4j-spring-boot2/src/test/resources/application.yaml
Romeh: it would be possible to use the same back-end with the 2 types of Bulkhead , i will add a test case for it as we filter on threadPoolProperties if it is defined or not .

for shared config , the user still can do the same as the normal bulkhead , just he will add  threadPoolProperties section there in the shared or default part

resilience4j-annotations/src/main/java/io/github/resilience4j/bulkhead/annotation/Type.java
```diff
@@ -0,0 +1,11 @@
+package io.github.resilience4j.bulkhead.annotation;
+
+/**
+ * bulkhead implementation types
+ * <p>
+ * SEMAPHORE will invoke semaphore based bulkhead implementation
+ * THREADPOOL will invoke Thread pool based bulkhead implementation
+ */
+public enum Type {
```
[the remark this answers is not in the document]
Romeh: Yes i will do that !

resilience4j-spring-boot2/src/test/resources/application.yaml
RobWin: I would not use `threadPoolProperties` as a sub property. I looks like the semaphore-properties `maxConcurrentCall` and `maxWaitTime` would be inherited from default, if you use `baseConfig: default`, which is not the case.
I would really separate them into two namespaces.

hexmind:
> I was wondering if we should add an option to rename the threads used by our TreadhPoolBulkhead, or rename them automatically and use the bulkhead name for it.

https://github.com/resilience4j/resilience4j/pull/473

Romeh:
@RobWin  Example from spring boot 2 test configuration for fixed thread pool bulkhead separate name space:
```
  configs:
    default:
      threadPoolProperties:
        maxThreadPoolSize: 4
        coreThreadPoolSize: 2
        queueCapacity: 2
  backends:
    backendB:
      baseConfig: default
    backendC:
      threadPoolProperties:
        maxThreadPoolSize: 1
        coreThreadPoolSize: 1
        queueCapacity: 1

```

Romeh:
i will create new PR to check if travis will work properly

PR #465, opened 2019-05-24
FluxRateLimiter and MonoRateLimiter try to reserve a permission and d…

RobWin:
…elay the subscription if possible.


Romeh:
@RobWin , the Mono delay will execute that in default scheduler (Schedulers.parallel()) , is that would be fine ? to use that scheduler all the time implicitly ?

RobWin:
You suggest that the scheduler should also be configurable, correct?

Romeh:
at least give the user the option to instance the Reactive limiter with a custom scheduler if he likes ? what do u think ?

RobWin:
`delaySubscription` uses internally `Mono.delay`.
I had the fear that this creates a recursive loop.

PR #466, opened 2019-05-24
Update javadocs for resilience4j-bulkhead Bulkhead.java

ishanthilina:
A very simple Javadoc update. I found the same line in `resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java` as well. Let me know if that needs to be updated as well.


RobWin:
Thx

RobWin:
If you can like, you can fix it in CircuitBreaker as well

PR #473, opened 2019-05-25
Use bulkhead name for threads naming

hexmind:


resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/NamingThreadFactory.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+class NamingThreadFactory implements ThreadFactory {
+
+    private final ThreadGroup group;
+    private final AtomicInteger threadNumber = new AtomicInteger(1);
+    private final String prefix;
+
+    NamingThreadFactory(String prefix) {
+        this.group = getThreadGroup();
+        this.prefix = prefix + "-";
```
RobWin: I think we should change it to `bulkhead-prefix-`.
What do you think? Usually the prefix might be a name of a backend system.

hexmind: Do you mean something like `this.prefix = "bulkhead-$name-"`? Makes sense.

RobWin: Yes

PR #474, opened 2019-05-28
Fixed threadpool bulkhead spring

Romeh:
@RobWin Example from spring boot 2 test configuration for fixed thread pool bulkhead separate name space:

```
  configs:
    default:
      threadPoolProperties:
        maxThreadPoolSize: 4
        coreThreadPoolSize: 2
        queueCapacity: 2
  backends:
    backendB:
      baseConfig: default
    backendC:
      threadPoolProperties:
        maxThreadPoolSize: 1
        coreThreadPoolSize: 1
        queueCapacity: 1
```


Romeh:
found an issue with separate name space in DefaultEventConsumerRegistry , as now u can register the same backend name twice , one for semaphore and one for fixed thread pool which leads to override the string key name in the map of DefaultEventConsumerRegistry

RobWin:
Mh, then let's add a prefix to the event consumer id.

Romeh:
Done !

Romeh:
@RobWin  should i rename backends to instances as it is new feature anyhow , so let us dot it right from the beginning or it will be confusing for the user ? i can maybe add it to the rest of spring support and deprecate backends to not break the backward compatibility

RobWin:
We have a similar discussion in PR #460

RobWin:
Let's do it in a separate PR

PR #477, opened 2019-05-29
Don't prevent using other call adapters in CircuitBreakerCallAdapter (fixes #431)

### LINKED ISSUE #431: Unable to use resilience with retrofit and rxjava
I have the below code to use resilience4j with retrofit and rxjava:

new Retrofit.Builder()
                        .baseUrl(walletServerBaseUrl)
                        .addConverterFactory(JacksonConverterFactory.create())
                        .addCallAdapterFactory(new RxErrorHandlingCallAdapter())
						.addCallAdapterFactory(CircuitBreakerCallAdapter.of(circuitBreaker, (r) -> r.code() < 500)).build();

However, this is not working and even if circuit opens it still passes requests.

apottere:
The `CircuitBreakerCallAdapter` doesn't need to actually adapt the call, it just needs to delegate to whatever call adapter *should* be handling a certain return type.

This fixes #431 with one caveat: The `CircuitBreakerCallAdapter` needs to be added to retrofit before any other adapters.

### DIFF
--- a/resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/CircuitBreakerCallAdapter.java
+++ b/resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/CircuitBreakerCallAdapter.java
@@ -25,7 +25,6 @@ import retrofit2.Response;
 import retrofit2.Retrofit;
 
 import java.lang.annotation.Annotation;
-import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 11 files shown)


resilience4j-retrofit/src/main/java/io/github/resilience4j/retrofit/CircuitBreakerCallAdapter.java
```diff
@@ -63,30 +62,20 @@ private CircuitBreakerCallAdapter(final CircuitBreaker circuitBreaker, final Pre
     }
 
     @Override
-    public CallAdapter<?,?> get(Type returnType, Annotation[] annotations, Retrofit retrofit) {
-        if (getRawType(returnType) != Call.class) {
-            return null;
-        }
+    public CallAdapter<?, ?> get(Type returnType, Annotation[] annotations, Retrofit retrofit) {
+        @SuppressWarnings("unchecked")
+        CallAdapter<Object, Object> nextAdapter = (CallAdapter<Object, Object>) retrofit.nextCallAdapter(this, returnType, annotations);
```
apottere: AFAICT this is the only way to tell the compiler that what we're going is ok, input welcome though.

RobWin:
Thx, could you please add a test that's shows that the delegation works as expected?

apottere:
Sure thing

apottere:
I added some tests that show that the RxJava adapter does work with the new `CircuitBreakerCallAdapter`, and I also updated (and added tests for) the `RateLimiterCallAdapter`.

One thing to note is that in `RetrofitCircuitBreaker` and `RetrofitRateLimiter` I had to subclass `DecoratedCall` instead of returning an anonymous class.  The RxJava adapter calls `call.clone()`, but the anonymous classes weren't overriding `clone()`, so it was actually using `DecoratedCall.clone()` - but this returns a `DecoratedCall`, not whatever the anonymous class was.  Since anonymous classes don't have real constructors, there was no way to simply override `clone()` to preserve their behavior on clone.

RobWin:
Thank you very much

Issue #476, opened 2019-05-29
Why aren't there retrieval-only fns for registries?


KingMob:
The registries overload the 1-arity String-only fns (e.g., `.retry`, `.circuitBreaker`) for not just lookup, but creation, if one is not available.

Is there a reason there's no lookup-only fn that will return error/false/null (or throw) if you ask for a name not in the registry? It would be a nice way to guard against typos.

RobWin:
Typos are not allowed ;)
Any good name for such a method? 
`findCircuitbreaker`

KingMob:
Hmm, I guess I would personally go with `find` or `get`. That way it could be added to the `AbstractRegistry` and shared with the non-circuit-breaker registries. And it should return an `Optional`, I suppose?

I can make a PR later, if the above sounds good to you.

RobWin:
Let's use find and return an Optional

KingMob:
Sounds good. I'll try to get to it tonight, but it might have to wait til the weekend.

KingMob:
Implemented by #478

PR #478, opened 2019-05-30
Add Registry.find() method and test

KingMob:
### About
Add find() method that will only return an existing object in a Registry but will not create a new one.

Addresses #476.

### Notes
I added a test, miniscule as it is. Let me know if I forgot anything or you want changes; my Java's a little rusty.


RobWin:
Thx

PR #482, opened 2019-06-02
Fix typo in core_guides/timelimiter

dlsrb6342:
Fix sample code typo in core_guides/timelimiter.adoc.
Use Try.of() with method reference like another sample code.


RobWin:
Thx

PR #483, opened 2019-06-05
Consume the thrown exception in a Feign fallback

celcius112:
A useful feature present in Hystrix was the `FallbackFactory` which let us consume the exception thrown on error. This commit proposes to add this functionality in the Feign module by making it possible for a fallback to consume the exception.


RobWin:
Thank you. I have a look at it.

PR #487, opened 2019-06-09
adding threadpool bulkhead for ratpack, some refactoring, closes #486

### LINKED ISSUE #486: Add ratpack bulkhead configs
See 
https://github.com/resilience4j/resilience4j/pull/474
(https://github.com/resilience4j/resilience4j/commit/ec72266fe7e34cd4804972b1499bac17b871e2a2)

drmaas:


Romeh:
renaming to instance is good approach but what about backward compatibility with previous releases ?

drmaas:
@Romeh See the following - it guarantees backwards compatibility for system properties/yaml/properties.

```
	/**
	 * For backwards compatibility when setting backends in configuration properties.
	 */
	public Map<String, InstanceProperties> getBackends() {
		return instances;
	}
```

This was done for all of the other resilience types.

PR #496, opened 2019-06-14
sonar try out for kotlin

Romeh:


Romeh:
thsi PR need to more config changes to to see how we can enable Kotling code coverage

PR #501, opened 2019-06-17
honor global configs for ratpack addon, closes #500

### LINKED ISSUE #500: Honor default external configs in ratpack apps
@RobWin fyi

drmaas:


drmaas:
@RobWin if it looks good for circuitbreaker, will apply to remainder.

drmaas:
@RobWin the build is working locally via `./gradlew check` - not sure why it is failing in travis.

RobWin:
Yes, very strange.
The following like fails `circuitBreakerConfig2.empty `

RobWin:
Do you use an Apple? 
I think there could be a hidden character inside of `circuitBreakerConfig2.empty`.
Maybe you can convert it into HEX or some other unicode to check it.

RobWin:
We could try to close the PR and create it again and see what happens

RobWin:
Which JDK Version do use locally?
When I test it with JDK8 i get the following error

```
Condition failed with Exception:

circuitBreakerConfig2.empty
|                     |
Optional.empty        groovy.lang.MissingPropertyException: No such property: empty for class: java.util.Optional
```

drmaas:
Issue was the `isEmpty` method was added in jdk11.

```
    /**
     * If a value is  not present, returns {@code true}, otherwise
     * {@code false}.
     *
     * @return  {@code true} if a value is not present, otherwise {@code false}
     * @since   11
     */
    public boolean isEmpty() {
        return value == null;
    }
```

Issue #509, opened 2019-06-21
Support for extending Micrometer tags


batigoal82:
I want to be able to extend the tags added to a gauge in the micrometer library.

But currently all classes responsible for metrics tags have a private constructor and thus cannot be easily extended.

* `TaggedRetryMetrics`
* `TaggedCirciutBreakerMetrics`
* `TaggedBulkheadMetrics`
* `TaggedRateLimiterMetrics`

Could you make the classes extendable or add a way I could provide the bean as in `RetryMetricsAutoConfiguration` myself that is not a closed class?

 Thanks!

RobWin:
Would you like to create a PR?

batigoal82:
yes, I will create a PR.

RobWin:
Do you want to add global tags, which should be added to every gauge or only a specific one?

batigoal82:
> Do you want to add global tags, which should be added to every gauge or only a specific one?

I would like to add specific ones.

RobWin:
Ok, is it a static value or a dynamic value which must be discovered at runtime?
I'm asking if it make sense to provide an option to configure static tags via the config file.

batigoal82:
> Ok, is it a static value or a dynamic value which must be discovered at runtime?
> I'm asking if it make sense to provide an option to configure static tags via the config file.

It is a dynamic value which is discovered at runtime. I like your idea regarding configurable static tags although I have a doubt that they would be used.

RobWin:
The beans `TaggedRetryMetrics` in `RetryMetricsAutoConfiguration` already use the annotation `ConditionalOnMissingBean`. Isn't it already possible to overwrite the bean (even without your PR)?

Could you use a `MeterFilter` to add a additional tag to certain metrics? See https://micrometer.io/docs/concepts#_transforming_metrics and https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html#production-ready-metrics-per-meter-properties

siepkes:
Were any API's for this added?

In my case I have a larger application which consists of modules and was looking to add a simple tag with something like `module=<MODULE NAME>` for metrics used in specific a module. So while the tag name is set at compile time it's not a static tag in the sense that it needs to be added globally to all metrics.

RobWin:
Hi,
no, we did not add anything yet. A PR is welcome. 
Do  you need the `module` tag to filter for all CircuitBreakers instances which are used in a certain module?

[...]

PR #510, opened 2019-06-21
Issue #509: Support for extending Micrometer tags

batigoal82:
For extending the tags provided by resilience4j, the classes that implement
MeterBinder for the respective concern(Retry, CircuitBreaker, RateLimiter
and Bulkhead) need to be extendable. With this commit, they all now implement
respective interfaces and these interfaces are set as the return type for
the respective spring boot auto configuration.


RobWin:
I close this PR. We should find a different solution.

PR #516, opened 2019-06-24
adding a demo ratpack java app

drmaas:


RobWin:
Could you please create a separate small repo like this [one](https://github.com/resilience4j/resilience4j-spring-boot2-demo), which we can be transfered into the Resilience4j project

Issue #201, opened 2018-02-06
Adaptive capacity management (RateLimiter + BulkHead) decorator


RobWin:
See concept: https://www.youtube.com/watch?v=m64SWl9bfvk

storozhukBM:
## Implementation thoughts.

As Jon Moore mentioned in his talk we will can increased latency as implicit indication for adaptive concurrency adjustments.

#### Adaptive capacity management prerequisites
1. You should have a system with relatively stable response latency, because we will use latency measures to adapt concurrency limits.
2. To configure you system properly you should figure out two things: 
2.1 Desirable average throughput per second (later **_`X`_**) [Example: 30 req/sec]
2.2 Desirable average request latency in seconds per operation (later **_`R`_**) [Example: 0.1 sec/op]
2.3 Maximum acceptable request latency (later **_`Rmax`_**). This number should be set wisely, because it can eliminate all adaptive capabilities, system will do its best to never reach such latency, so you can set it 20-30 % higher than your usual average latency.

#### Implementation
Resilience4j will provide new Bulkhead implementation called `AdaptiveBulkhead`. It will have following config params:
 1. Desirable average throughput = X
 2. Desirable request latency = R
 3. Maximum acceptable request latency = Rmax (default `R * 1.3`)
 4. Window duration for adaptation = Wa (default 5 sec)
 5. Window duration for reconfiguration = Wr (default 900 sec)
---
From this params we will calculate:
1. Initial average number of concurrent requests (later **_`N`_**).
Example:  `N = X * R` = `30 * 0.1` = `3` **[op]**
2. Initial max latency of current window (later **_`cRmax`_**).
Example:  `cRmax = min(R * 1.2, Rmax)` = `min(0.1 * 1.2, 0.13)` = `0.12` **[op]**
3. Size of adaptation sliding window (later **_`WaN`_**)
Example: `WaN = Wa * X = 30 * 5 = 150`
4. Size of reconfiguration sliding window (later **_`WrN`_**)
Example: `WrN = Wr / Wa = 900 / 5 = 180`
---

From now on we will have two separate functions working inside `AdaptiveBulkhead`:
##### First function is `constantAdaptation`
It will constantly measure request latencies and add them to adaptation sliding window. 
[...]

PR #520, opened 2019-06-25
Adaptive bulkhead implementation

Romeh:
Adaptive bullhead based into https://github.com/resilience4j/resilience4j/issues/201 discussion 

The adaptive bulkhead implementation is as the following :
**- the adaptive bulkhead workflow :** 
Untitled Diagram

**- the AIMD limit policy :**
AIMDFlow
Next action will be providing Adaptive Thread pool bulkhead once this is reviewed


resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/LimitAdapter.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.function.Consumer;
+
+import io.github.resilience4j.bulkhead.Bulkhead;
+import io.github.resilience4j.bulkhead.BulkheadConfig;
+import io.github.resilience4j.bulkhead.event.BulkheadLimit;
+import io.github.resilience4j.core.lang.NonNull;
+
+
+/**
+ * Limit adopter interface for adoptive bullhead {@link AdaptiveBulkhead}
```
drmaas: spelling - `Limit adapter interface for adaptive bulkhead {@link AdaptiveBulkhead}`

Romeh:
Still need to add more unit testing

Romeh:
@RobWin  it is work in progress ,I have pushed the first round of changes , I would really like to have ur review for the draft design /API changes to accommodate any change sooner !

Still need to be polished more after last changes !

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/AdaptiveLimitBulkhead.java
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	@Override
+	public void releasePermission() {
+		bulkhead.releasePermission();
+		inFlight.decrementAndGet();
+	}
+
+	@Override
+	public void onComplete(Duration callTime, boolean isSuccess) {
```
RobWin: Can we replace `onComplete` by `onSuccess` and `onError` to make the API comparable to the CircuitBreaker.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,546 @@
...
+			} finally {
+				if (start != null) {
+					Instant finish = Instant.now();
+					bulkhead.onComplete(Duration.between(start, finish), true);
+				}
+			}
+		};
+	}
+
+	static void handleError(AdaptiveBulkhead bulkhead, @Nullable Instant start, Exception e) {
```
RobWin: Can we move this into an `onError` method` of the AdaptiveBulkhead implementation.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkheadConfig.java
```diff
@@ -0,0 +1,127 @@
...
+
+		private Builder() {
+			adaptiveBulkheadConfig = new AdaptiveBulkheadConfig<>();
+		}
+
+		private Builder(AdaptiveBulkheadConfig<T> adaptiveBulkheadConfig) {
+			this.adaptiveBulkheadConfig = adaptiveBulkheadConfig;
+		}
+
+		public Builder<T> adaptIfError(Predicate<Exception> adaptIfError) {
```
RobWin: I think we should implement it similar to the CircuitBreaker.
Some exceptions can be totally ignored. Some exceptions count as a success and some exceptions should increase the failure rate.

RobWin: Means: We need `ignoreExceptions` and `recordExceptions`.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/config/AbstractConfig.java
```diff
@@ -0,0 +1,211 @@
...
+	private static final double DEFAULT_SLOW_CALL_RATE_THRESHOLD = 50.0; // Percentage
+	private static final int DEFAULT_SLIDING_WINDOW_SIZE = 100;
+	private static final long DEFAULT_SLOW_CALL_DURATION_THRESHOLD = 5; // Seconds
+	private static final int DEFAULT_SLIDING_WIN_TIME = 10; // Seconds
+
+	double failureRateThreshold = DEFAULT_FAILURE_RATE_THRESHOLD;
+	int slidingWindowSize = DEFAULT_SLIDING_WINDOW_SIZE;
+	SlidingWindow slidingWindowType = DEFAULT_SLIDING_WINDOW_TYPE;
+	double slowCallRateThreshold = DEFAULT_SLOW_CALL_RATE_THRESHOLD;
+	Duration desirableLatency = Duration.ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD);
```
RobWin: The name `desirableLatency` is nice. But maybe we should use the same name like the CircuitBreakerConfig does which is `slowCallDurationThreshold`

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,546 @@
...
+	 *
+	 * @param name                   the name of the bulkhead
+	 * @param bulkheadConfigSupplier custom configuration supplier
+	 * @return a Bulkhead instance
+	 */
+	static AdaptiveBulkhead of(String name, Supplier<AdaptiveBulkheadConfig> bulkheadConfigSupplier) {
+		return AdaptiveLimitBulkhead.factory().createAdaptiveLimitBulkhead(name, bulkheadConfigSupplier.get());
+	}
+
+	interface AdaptiveBulkheadMetrics extends Bulkhead.Metrics {
```
RobWin: We should call the interface `Metrics` as well. 
We always used `Bulkhead.Metrics`, `CircuitBreaker.Metrics`, `RateLimiter.Metrics`.

Can we also include `getNumberOfFailedCalls`,  `getNumberOfSuccessfulCalls`, and `getNumberOfSlowCalls`. 
I think we can should getAverageLatencyMillis.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,546 @@
...
+
+		/**
+		 * @return average latency for service calls in millis
+		 */
+		double getAverageLatencyMillis();
+
+		/**
+		 * @return current failure rate for recorded calls if error check is enabled for related exceptions
+		 */
+		double getFailureRate();
```
RobWin: Should be float

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,546 @@
...
+
+		/**
+		 * @return current slow call rate for the recorded calls
+		 */
+		double getSlowCallRate();
```
RobWin: Should be float

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	/**
+	 * An EventPublisher which can be used to register event consumers.
+	 */
+	interface AdaptiveEventPublisher extends io.github.resilience4j.core.EventPublisher<BulkheadLimit> {
+
+		EventPublisher onLimitIncreased(EventConsumer<BulkheadOnLimitDecreasedEvent> eventConsumer);
```
RobWin: We should also add `onSuccess`, `onError`, `onIgnoredError`, 'onCallNotPermitted'  and `onStateTransition`.  
We should keep a state in the AdaptiveBulkhead implementation and report state changes from:  Slow Start -> Congestion Avoidance

`onCallNotPermitted` reports when calls are rejected, because the bulkhead is full.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/AdaptiveLimitBulkhead.java
RobWin: Please use `long duration, TimeUnit durationUnit` instead of `Duration callTime`.
The reasons is that we don't want to create a Duration object for every call. 
TimeUnit instead is just an enum and a primitve. It should result in less garbage.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/amid/AIMDLimiter.java
```diff
@@ -0,0 +1,219 @@
...
+	private final Consumer<BulkheadLimit> publishEventConsumer;
+	private final Metrics metrics;
+	private final AdaptiveBulkheadConfig<AIMDConfig> amidConfigAdaptiveBulkheadConfig;
+	private final long desirableLatency;
+
+
+	public AIMDLimiter(@NonNull AdaptiveBulkheadConfig<AIMDConfig> config, Consumer<BulkheadLimit> publishEventConsumer) {
+		this.amidConfigAdaptiveBulkheadConfig = config;
+		if (amidConfigAdaptiveBulkheadConfig.getConfiguration().getSlidingWindowType() == AIMDConfig.SlidingWindow.COUNT_BASED) {
+			this.metrics = new FixedSizeSlidingWindowMetrics(config.getConfiguration().getSlidingWindowSize());
```
RobWin: I think the limit policy should not care how metrics are measured. The limit policy implementation should just get some input and return some adaption value. Like a function.  The adaption of the bulkhead itself should not be part of the Limit Policy, but part of the AdaptiveBulkhead implementation.

We should move this code into the AdaptiveLimitBulkhead.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/amid/AIMDLimiter.java
```diff
@@ -0,0 +1,219 @@
...
+		desirableLatency = amidConfigAdaptiveBulkheadConfig.getConfiguration().getDesirableLatency().toNanos();
+	}
+
+	@Override
+	public void adaptLimitIfAny(@NonNull Bulkhead bulkhead, @NonNull Duration callTime, boolean isSuccess, int inFlight) {
+		if (LOG.isDebugEnabled()) {
+			LOG.debug("starting the adation of the limit for callTime :{} , isSuccess: {}, inFlight: {}", callTime.toMillis(), isSuccess, inFlight);
+		}
+		Snapshot snapshot;
+		if (isSuccess) {
```
RobWin: We should move this code into the AdaptiveLimitBulkhead.

RobWin (changes requested):

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: why ? handleError is needed into the default methods of Adapive bulkhead interface

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: onCallNotPermitted will be covered by the internal semaphore bulkhead events already !, covered the rest except the state events as we do not have implicit states in our implementations

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+	 * @return bulkhead config
+	 */
+	AdaptiveBulkheadConfig getBulkheadConfig();
+
+	/**
+	 * Get the AdaptiveMetrics of this Bulkhead.
+	 *
+	 * @return the AdaptiveMetrics of this Bulkhead
+	 */
+	AdaptiveMetrics getMetrics();
```
RobWin: I think we should stick to `AdaptiveBulkead.Metrics` like in our other components. Even if Sonar complains about it ;)

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+		 * Returns the current number of successful buffered calls in the ring buffer.
+		 *
+		 * @return the current number of successful buffered calls in the ring buffer
+		 */
+		int getNumberOfSuccessfulCalls();
+
+		/**
+		 * @return average latency for service calls in millis
+		 */
+		double getAverageLatencyMillis();
```
RobWin: Do we need to make `getAverageLatencyMillis` public?

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+	 *
+	 * @param name                   the name of the bulkhead
+	 * @param bulkheadConfigSupplier custom configuration supplier
+	 * @return a Bulkhead instance
+	 */
+	static AdaptiveBulkhead of(String name, Supplier<AdaptiveBulkheadConfig> bulkheadConfigSupplier) {
+		return AdaptiveLimitBulkhead.factory().createAdaptiveLimitBulkhead(name, bulkheadConfigSupplier.get());
+	}
+
+	interface AdaptiveMetrics extends Bulkhead.Metrics {
```
RobWin: We added `getNumberOfSlowSuccessfulCalls` to `CircuitBreaker.Metrics`C. Can we add this here as well?

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+		/**
+		 * @return current slow call rate for the recorded calls
+		 */
+		double getSlowCallRate();
+	}
+
+	/**
+	 * An EventPublisher which can be used to register event consumers.
+	 */
+	interface AdaptiveEventPublisher extends io.github.resilience4j.core.EventPublisher<BulkheadLimit> {
```
RobWin: Can we rename the base interface to `AdaptiveBulkeadEvent` comparable to `CircuitBreakerEvent` or `BulkheadEvent`

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+	 * @param <T>      the type of results supplied by this supplier
+	 * @return a supplier which is decorated by a Bulkhead.
+	 */
+	static <T> CheckedFunction0<T> decorateCheckedSupplier(AdaptiveBulkhead bulkhead, CheckedFunction0<T> supplier) {
+		return () -> {
+			long start = 0;
+			boolean isFailed=false;
+			bulkhead.acquirePermission();
+			try {
+				start = Instant.now().toEpochMilli();
```
RobWin: Can we use everywhere `final long start = System.nanoTime();`.
It reduces the amount of `Instant` classes which we create per call and improves GC.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/AdaptiveLimitBulkhead.java
```diff
@@ -0,0 +1,409 @@
...
+		return eventData;
+	}
+
+
+	/**
+	 * @param callTime the call duration time
+	 * @param durationUnit the duration unit
+	 * @param throwable the error exception
+	 */
+	private void handleError(long callTime, TimeUnit durationUnit, Throwable throwable) {
```
RobWin: Could you please add some JavaDoc description what do method does.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/AdaptiveLimitBulkhead.java
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+
+	/**
+	 * @param callTime  the call duration
+	 * @param isSuccess is the call successful or not
+	 * @param inFlight  current in flight calls
+	 * @return the update limit result DTO @{@link LimitResult}
+	 */
+	protected LimitResult record(@NonNull long callTime, boolean isSuccess, int inFlight) {
```
RobWin: Could you please add some JavaDoc description what do method does.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+				consumer.accept(t);
+			} catch (Exception e) {
+				isFailed=true;
+				bulkhead.onError( start,TimeUnit.MILLISECONDS, e);
+				throw e;
+			} finally {
+				if (start != 0 && !isFailed) {
+					bulkhead.onSuccess(start, TimeUnit.MILLISECONDS);
```
RobWin: Why do we need `onSuccess` in the finally part?
Shouldn't it be the total duration instead of `start`?
`long durationInNanos = System.nanoTime() - start;`

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+		return t -> {
+			long start = 0;
+			boolean isFailed=false;
+			bulkhead.acquirePermission();
+			try {
+				start = Instant.now().toEpochMilli();
+				consumer.accept(t);
+			} catch (Exception e) {
+				isFailed=true;
+				bulkhead.onError( start,TimeUnit.MILLISECONDS, e);
```
RobWin: Shouldn't it be the total duration instead of start?

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+
+	/**
+	 * Records a successful call and releases a permission.
+	 */
+	void onSuccess(long startTime, TimeUnit durationUnit);
+
+	/**
+	 * Records a failed call and releases a permission.
+	 */
+	void onError(long startTime, TimeUnit durationUnit,Throwable throwable);
```
RobWin: Shouldn't it be the total `duration` instead of `startTime`?

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
```diff
@@ -0,0 +1,575 @@
...
+	 * <p>
+	 * Should only be used when a permission was acquired but not used. Otherwise use
+	 * {@link AdaptiveBulkhead#onSuccess(long, TimeUnit)} to signal a completed call and release a permission.
+	 */
+	void releasePermission();
+
+	/**
+	 * Records a successful call and releases a permission.
+	 */
+	void onSuccess(long startTime, TimeUnit durationUnit);
```
RobWin: Shouldn't it be the total `duration` instead of `startTime`?

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/LimitPolicy.java
```diff
@@ -0,0 +1,44 @@
...
+package io.github.resilience4j.bulkhead.adaptive;
+
+import io.github.resilience4j.bulkhead.Bulkhead;
+import io.github.resilience4j.bulkhead.BulkheadConfig;
+import io.github.resilience4j.core.lang.NonNull;
+import io.github.resilience4j.core.metrics.Snapshot;
+
+
+/**
+ * Limit adapter interface for adoptive bulkhead {@link AdaptiveBulkhead}
```
RobWin: Could you please add some documentation the reason why this Interface exists.
For example that in future new concurrency limit algorithms could be added.

Could you also add some documentation about the usage of  `LimitResult`. What does the AdaptiveBulkhead do with this LimitResult. It should help developers to implement new policies.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/LimitResult.java
```diff
@@ -0,0 +1,50 @@
+package io.github.resilience4j.bulkhead.adaptive;
+
+import java.util.Objects;
+
+/**
+ * limit result DTO
+ */
+public class LimitResult {
```
RobWin: Could you also add some documentation about the usage of LimitResult. What does the AdaptiveBulkhead do with this LimitResult. It should help developers to implement new policies.

RobWin (changes requested):

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/config/AbstractConfig.java
```diff
@@ -0,0 +1,212 @@
...
+import java.time.Duration;
+import java.util.Objects;
+
+import io.github.resilience4j.bulkhead.BulkheadConfig;
+import io.github.resilience4j.core.lang.NonNull;
+
+/**
+ * abstract common configuration
+ */
+public class AbstractConfig {
```
RobWin: I think we can move this into `AdaptiveBulkheadConfig`.
The way how we measure metrics is not changable by the users. We provide to options: count-based and time-based. Maybe we add a third in the future, but that's up to us.

RobWin (changes requested):

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: in case if community would like to introduced new policies like moving average ..ect , it can be useful for them especially we export that anyhow through our metrics

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: I guess the user can easily get that if he want to have it by getNumberOfSlowCalls()-getNumberOfFailedCalls() , so i do not feel it is needed as mainly he will be concerned about slow calls in general i guess

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: I feel it is like more optimized in current impl , as the allocation of start to now will only happen IF bulkhead.acquirePermission() pass plus we need the reference of start time in the finally block .

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: Inside OnError we do that calculation as if it is an ignored error we do not care about it

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: yes will rename it

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: here it is only start time as it is calculated into the onError impl

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: yes will rename it

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/AdaptiveBulkhead.java
Romeh: yes it should be duration , updated accordingly, we need on success on finally block as we need record the successful calls if they are not failed one !

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/internal/config/AbstractConfig.java
Romeh: Can you please explain more what you want here :) ? as Abstract config is our way to impose what is the current common configuration and metrics so for different policies ,users will just extend that one like we did for AIMD policy config class and add their extra properties if needed

RobWin: Yes, but the common configuration can be part of `AdaptiveBulkheadConfig`, or? 
The policy gets a `AdaptiveBulkheadConfig<WhatEverLimiter>` injected, or?

RobWin: I'm not talking about the class itself, but the content of the class can be moved into the common AdaptiveBulkheadConfig.

Romeh: I would like to keep it like that as it is as merging the content of AsbtractConfi into AdaptiveBulkHeadConfig will make the class too big and not easy to maintain , like that for now it is well encapsulated

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/adaptive/LimitPolicy.java
```diff
@@ -0,0 +1,47 @@
...
+ * There is a default implementation {@link io.github.resilience4j.bulkhead.adaptive.internal.amid.AimdLimiter}
```
jimhooker2002, line 29: Minor: amid -> aimd.

RobWin: This is an old branch. I will close this PR

jimhooker2002: (that package name is incorrect)

mdavidn:
For readers like me: This effort moved to [other branches](https://github.com/resilience4j/resilience4j/tree/adaptive_bulkhead_4). See #201. No new PR at time of writing.

Issue #538, opened 2019-07-09
Document or remove minimum interval for retry


chrylis:
In testing a component that's supposed to retry a failed REST call onto a mock REST client, I set the retry interval to 1ms. This produced an exception at runtime saying that the minimum interval was 10ms. This isn't documented anywhere, and it seems like the sort of constraint that either shouldn't be enforced by the framework, or if it is, clearly documented in both Javadoc and the reference documentation.

RobWin:
Hi, 
we can remove the constraint. Would you like to create a PR?

Romeh:
@RobWin  the needed PR got created for ur review

RobWin:
Merged

dalexander01:
Hey quick question, since the constraint has been removed, what happens when the wait duration is 0?

RobWin:
The retry attempt is executed immediately

dalexander01:
Ok makes sense.

PR #539, opened 2019-07-10
#538 loesn up retry interval value

Romeh:


resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -143,10 +143,9 @@ public Builder(RetryConfig baseConfig) {
 		}
 
 		public Builder<T> waitDuration(Duration waitDuration) {
-			if (waitDuration.toMillis() < 10) {
-				throw new IllegalArgumentException("waitDurationInOpenState must be at least 10ms");
+			if (waitDuration.toMillis() >= 0) {
+				this.intervalFunction = (x) -> waitDuration.toMillis();
```
RobWin: It should throw an exception if the value is < 0

Romeh: is it possible that someone will try set value of interval with less than 0 (- value) :)?

Romeh: updated anyhow :)

PR #542, opened 2019-07-11
properly handle onComplete for ratpack circuit transformer by releasing permission

fromanator:


fromanator:
Verified that the test breaks by permanently being stuck in `HALF_OPEN` state returning `CallNotPermittedException` before the fix.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/circuitbreaker/CircuitBreakerTransformer.java
```diff
@@ -77,6 +77,7 @@ public void error(Throwable throwable) {
 
                     @Override
                     public void complete() {
+                        circuitBreaker.releasePermission();
```
RobWin: `releasePermission` should only be called, if downstream obtained a permission to upstream, but upstream was cancelled afterwards.
If the upstream completes successfully, we should call circuitbreaker.onSuccess instead.
If upstream handles the error and therefore onComplete is invoked, we should assume that the call has completed successfully. We can't assume that the call was cancelled.

Have a look at the RxJava2 `CompletableCircuitBreaker` or Reactor `CircuitBreakerSubscriber` implementatio

fromanator: Ratpack's name `complete` is very misleading here. Effectively it's a cancellation of the Promise: https://ratpack.io/manual/current/api/ratpack/exec/Downstream.html#complete--

Basically if the upstream execution had an error and swallowed it, then it calls `complete` on all the executions downstream of it to notify them that there's no work to do since it will not provide any data to them.

resilience4j-ratpack/src/test/groovy/io/github/resilience4j/ratpack/circuitbreaker/CircuitBreakerTransformerSpec.groovy
```diff
@@ -93,53 +95,84 @@ class CircuitBreakerTransformerSpec extends Specification {
...
         !r.error
         r.throwable == null
         breaker.state == CircuitBreaker.State.HALF_OPEN
 
-        when:
+        and: "do a call where the Upstream promise handles the error, thus calling complete on the circuit"
+
+        when: "error on half open, retry swallows exception thrown"
+        r = ExecHarness.yieldSingle {
+            Blocking.<String> get { throw e }.onError { exception -> Promise.value("not foo") }
```
fromanator: Here's what it looks like to swallow the exception. If you dig into the Ratpack code if we handle an error and then produce a promised value, then all the downstreams will have a `complete` call propagated down effectively canceling them.

resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/circuitbreaker/CircuitBreakerTransformer.java
RobWin: Ok, then its fine

RobWin:
Could you please check the bulkhead transformer as well?

fromanator:
@RobWin the bulkhead transformer looks like it's already releasing the permission when `complete` is called.

Issue #537, opened 2019-07-09
Spring Boot auto-configuration should use class conditions


chrylis:
I'm wanting to use the Spring Boot 2 starter just to supply a `RetryRegistry` that's already set up with Micrometer. However, startup fails with `ClassNotFoundException` because I don't have Spring AOP/AspectJ added to my project.

The aspect support is useful but not always needed in a Boot project, and it should be wrapped in a nested class that is `@ConditionalOnClass(ProceedingJoinPoint.class)` or similar.

RobWin:
Thx, would you like to create a PR?

chrylis:
Just getting back to some of this; does the new commit cover this, or are you still looking for a PR?

RobWin:
Yes, the PR covers it.

PR #548, opened 2019-07-22
Issue #537: Resilience4j Aspects should be only available when Aspect…

RobWin:
…j is on the classpath.

### DIFF
--- a/resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerConfiguration.java
+++ b/resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerConfiguration.java
@@ -56,6 +56,7 @@ public class CircuitBreakerConfiguration {
 	}
 
 	@Bean
+	@Conditional(value = {AspectJOnClasspathCondition.class})
 	public CircuitBreakerAspect circuitBreakerAspect(CircuitBreakerRegistry circuitBreakerRegistry,
 													 @Autowired(required = false) List<CircuitBreakerAspectExt> circuitBreakerAspectExtList,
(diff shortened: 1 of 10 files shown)


resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerConfiguration.java
```diff
@@ -64,13 +64,13 @@ public CircuitBreakerAspect circuitBreakerAspect(CircuitBreakerRegistry circuitB
 
 
 	@Bean
-	@Conditional(value = {RxJava2OnClasspathCondition.class})
+	@Conditional(value = {RxJava2OnClasspathCondition.class, AspectJOnClasspathCondition.class})
```
Romeh: I guess we need to do the same for the rest (Bulkhead ext..ect) plus we need to do the same on the circuit breaker aspect bean injection

RobWin: Which class do you mean?

Romeh: RxJava2RateLimiterAspectExt,ReactorRateLimiterAspectExt beans injection in RateLimiterConfigutation , same for Retry and Bulkhead plus in CircuitBreakerConifguration , add the same condition over CircuitBreakerAspect bean injection

RobWin: done

Issue #547, opened 2019-07-18
Circuit Breaker buffer time interval handling


codingL3gend:
is there any way to implement a time interval that the circuit breaker could use to evaluate request through the breaker over that interval to determine the failure rate threshold against as well. currently with this being a difference in resilience4j in comparison to Hystrix if the buffer size in the closed state was a value of 100 and say only 60 request came through the breaker in 5 mins and 30 failed and 30 succeeded the breaker would not calculate the failure rate until it reaches the 100. this is problematic because if some time later 40 more request came through the system  and the breaker still had history of the request and then 20 failed and then 20 succeeded at this time the breaker would move to an open state even though the backend system may not be experiencing issues. resetting the breaker does not solve this problem since the metrics would be lost and forcing the breaker to change state based on a calculation of the failure rate manually is not ideal either. is there an elegant way of handling such a situation?

RobWin:
Hu, thank you for your question.
I think there are 3 options

1) You could reduce the ring buffer size to a lower value so that the failure rate is calculated faster/earlier.

2) We implement a different CircuitBreaker strategy which uses a sliding time window.

3) We implement a different CircuitBreaker strategy which uses a weighted mean where older values contribute less.

RobWin:
4 option is of course that you keep Hystrix

codingL3gend:
@RobWin thanks for the reply, to your points mentioned we considered option 1 which is only slightly viable because it essentially will yield the same results. simply having the breaker setup to fail faster doesn't entirely address the issue since the tuning will have to be just right so that there are no preemptive moments of the breaker opening simply due to minor blips. of course with enough test and tuning that approach could perhaps get to an ideal configuration but the issue still remains that the breaker still will maintain state of request over time and could cause issues if the throughput over time wasn't that great. option 2 would be great, that is an approach that we utilize with the Polly framework for .net and it is more efficient. option 3 is also a great idea since it would help to get rid of outliers much more efficiently. option 4 is viable but we haven't done an implementation with Hystrix because research pointed to resilience4j as being better. i like resilience4j as a framework; it's just this one hick up.

RobWin:
@codingL3gend Please be aware that you still need to tune a CircuitBreaker that uses a sliding time window implementation.
Polly as well as Hystrix require you to configure `minimumThroughput` or `requestVolumeThreshold` according to the request rate of your system. Otherwise the CircuitBreaker will not trip open even if all calls failed.
[...]

PR #556, opened 2019-08-07
Issue #547: Added a Sliding Time Window implementation to measure met…

RobWin:
First draft of a Sliding Time Window metric store is pushed.

The implementation is backed by a sliding time window that aggregates only the calls made in the last N seconds.
The sliding time window is implemented with a circular array of N buckets.
If the time window size is 10 seconds, the circular array has always 10 buckets. Every bucket
stores the outcome of all calls which happen in a certain epoch second. The head bucket of the circular array stores the call outcomes of the current epoch second. The other buckets stored the call outcomes of the previous N-1 epoch seconds.

The sliding window does not store call outcomes (tuples) individually, but incrementally updates partial aggregations (bucket) and a total aggregation.
The total aggregation is updated incrementally when a new call outcome is recorded. When the oldest bucket is evicted, the partial aggregation of that bucket
is subtracted from the total aggregation and the bucket is reset. (Subtract-on-Evict)

The time to retrieve a Snapshot is constant 0(1), since the Snapshot is pre-aggregated and is independent of the time window size.
The space requirement (memory consumption) of this implementation should be nearly constant O(n), since the call outcomes (tuples) are not stored individually.
Only N partial aggregations and 1 total aggregation are created.

Feel free to review it ;)

This implementation would also allow us to track slow calls and adapt the CircuitBreaker so that it can transition to OPEN when the percentage of slow calls is above a certain threshold. It would solve #381


resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/Snapshot.java
```diff
@@ -0,0 +1,87 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+package io.github.resilience4j.core.metrics;
+
+import java.time.Duration;
+
```
Romeh: Thinking loud here , Small question for discussion , can we maybe have Snapshot<T> and same for Metrics<T> and let the user define what type of APIs/counts he can use for metrics by call get T . (api call) or we should enforce what type of stats he can get and should follow that ?

resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/SlidingTimeWindowMetrics.java
```diff
@@ -0,0 +1,134 @@
...
+ * The sliding window does not store call outcomes (tuples) individually, but incrementally updates partial aggregations (bucket) and a total aggregation.
+ * The total total aggregation is updated incrementally when a new call outcome is recorded. When the oldest bucket is evicted, the partial total aggregation of that bucket
+ * is subtracted from the total totalAggregation and the bucket is reset. (Subtract-on-Evict)
+ *
+ * The time to retrieve a Snapshot is constant 0(1), since the Snapshot is pre-aggregated and is independent of the time window size.
+ * The space requirement (memory consumption) of this implementation should be nearly constant O(n), since the call outcomes (tuples) are not stored individually.
+ * Only {@code N} partial aggregations and 1 total total aggregation are created.
+ */
+public class SlidingTimeWindowMetrics implements Metrics {
+
```
Romeh: I feel we should separate the sliding window logic and storage into separate class aside from the metrics , so anyone want to use outside the metrics Interface , he can do it , i am thinking about the adaptive bulkhead case here , what do u think ?

resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/SlidingWindowMetrics.java
```diff
@@ -0,0 +1,95 @@
...
+        latestMeasurement.reset();
+        return latestMeasurement;
+    }
+
+    /**
+     * Returns the head partial aggregation of the circular array.
+     *
+     * @return the head partial aggregation of the circular array
+     */
+    Measurement getLatestMeasurement(){
```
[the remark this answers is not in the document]
RobWin: Fixed

resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/SnapshotImpl.java
```diff
@@ -0,0 +1,96 @@
...
+ *
+ *
+ */
+package io.github.resilience4j.core.metrics;
+
+import java.time.Duration;
+
+public class SnapshotImpl implements Snapshot {
+
+    private long totalDurationInMillis;
```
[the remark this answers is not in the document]
RobWin: Fixed

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/SlidingWindowCircuitBreakerMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Records a call which was not permitted, because the CircuitBreaker state is OPEN.
+     */
+    void onCallNotPermitted() {
+        numberOfNotPermittedCalls.increment();
+    }
+
+    @Override
+    public float getFailureRate() {
```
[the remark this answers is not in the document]
RobWin: This class is deleted. Ignore it please.

resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/Snapshot.java
RobWin: Can you give an example? I don't understand it

Romeh: per discussion not applicable

wl1244hotmai:
Another question is that, is there any way don't to record the rate of slow call? Sometimes we already had timeout settings in http connection, so we don't need it. 
After this change, I may need to configure the slow SLOW_CALL_DURATION_THRESHOLD to a big one to avoid trigger it. It there any way to disable it?

RobWin:
Hi, could you please add your comments to the issue #547 instead of the the PR?

Issue #540, opened 2019-07-10
resilience4j-spring-boot2 con't auto recognize config with IDEA


bugCreate:
when use resilience4j-spring-boot2:0.13.0,IDEA can associate resilience4j‘s configuration and recognize the config.but,when use resilience4j-spring-boot2:0.16.0,IDEA can't  associate and recognize.
pom
<dependencies>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-actuator</artifactId>
        </dependency>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-web</artifactId>
        </dependency>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-aop</artifactId>
        </dependency>
        <dependency>
            <groupId>io.github.resilience4j</groupId>
            <artifactId>resilience4j-spring-boot2</artifactId>
            <version>0.16.0</version>
        </dependency>

        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-test</artifactId>
            <scope>test</scope>
        </dependency>
    </dependencies>

resilience4j-spring-boot2:0.16.0
![image](https://user-images.githubusercontent.com/34771055/60953919-ac8a1900-a330-11e9-9a2c-ba5847bca924.png)
![image](https://user-images.githubusercontent.com/34771055/60953937-b744ae00-a330-11e9-9d57-064be62b7d8f.png)
resilience4j-spring-boot2:0.13.0
![image](https://user-images.githubusercontent.com/34771055/60954070-f3780e80-a330-11e9-980a-14ef3d9b2a83.png)
![image](https://user-images.githubusercontent.com/34771055/60954088-fa9f1c80-a330-11e9-9b82-0618118f230d.png)

PR #566, opened 2019-08-14
Support auto complete config in spring boot

dlsrb6342:
Issue #540 

Please refer https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html
> With Gradle 4.5 and earlier, the dependency should be declared in the compileOnly configuration, as shown in the following example:
> ```
> dependencies {
> 	compileOnly "org.springframework.boot:spring-boot-configuration-processor"
> }
> ```
> With Gradle 4.6 and later, the dependency should be declared in the annotationProcessor configuration, as shown in the following example:
> ```
> dependencies {
> 	annotationProcessor "org.springframework.boot:spring-boot-configuration-processor"
> }
> ```

Because resilience4j gradle version is `5.3.1`, we have to use `annotationProcessor` not `compileOnly`
https://github.com/resilience4j/resilience4j/blob/b84a70a95fff1cd60d30c878d8deaafd710d535f/gradle/wrapper/gradle-wrapper.properties#L6

image
I tested this in local, it works well.

### DIFF
--- a/resilience4j-spring-boot-common/build.gradle
+++ b/resilience4j-spring-boot-common/build.gradle
@@ -2,9 +2,9 @@ dependencies {
     compile project(':resilience4j-spring')
     
     compileOnly(libraries.spring_boot2_aop)
-    compileOnly(libraries.spring_boot2_config_processor)
-    compileOnly(libraries.spring_boot2_autoconfigure_processor)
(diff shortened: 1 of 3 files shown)


RobWin:
Thank you very much

Issue #568, opened 2019-08-15
CircuitBreaker: How to treat exceptions as a success


harishmbm:
Hi,

I have the following setup:

1. A 3rd party service (runs on localhost:8090/sample) and it always returns bad request (400).
```
@RestController
@RequestMapping
public class SampleController {
    @GetMapping("/sample")
    public ResponseEntity helloWorld(HttpServletResponse response) {
        System.out.println("Request received");
        return ResponseEntity.badRequest().body("{\"message\":\"this is a bad request\"}");
    }
}
```
2. In my main application where I am using circuit breaker, I am using RestTemplate to communicate it with the 3rd party service. I have ignored `org.springframework.web.client.HttpClientErrorException` from the circuit breaker.
3. The circuit breaker annotation is kept at method level like below:
```
    @CircuitBreaker(name = "callOtherService", fallbackMethod = "errorExternal")
    public String callOtherService() {
        System.out.println("Call received");
        return restTemplate.getForObject(URI.create("http://localhost:8090/sample"), String.class);
    }

public String errorExternal(Throwable throwable) {
        return "This is error from fallback triggered by external error";
    }
```
4. I am using following configuration (see `callOtherService` section)
```
resilience4j.circuitbreaker:
  configs:
    default:
      registerHealthIndicator: true
      ringBufferSizeInClosedState: 4
      ringBufferSizeInHalfOpenState: 2
      automaticTransitionFromOpenToHalfOpenEnabled: true
      waitDurationInOpenState: 20s
      failureRateThreshold: 50
      eventConsumerBufferSize: 10
      ignoreExceptions:
       - com.resilience4j.exception.BusinessException
       - feign.FeignException
      recordExceptions:
        - java.net.SocketTimeoutException
        - java.net.ConnectException
    shared:
      registerHealthIndicator: true
      ringBufferSizeInClosedState: 4
      ringBufferSizeInHalfOpenState: 2
      waitDurationInOpenState: 20s
      failureRateThreshold: 50
      eventConsumerBufferSize: 10
      ignoreExceptions:
        - com.resilience4j.exception.BusinessException
  instances:
    callOtherService:
      baseConfig: default
      ringBufferSizeInClosedState: 4
      registerHealthIndicator: true
      ignoreExceptions:
        # We need to ignore 4xx errors returned by the server as they are valid business case.
        - org.springframework.web.client.HttpClientErrorException
      recordExceptions:
        # We need to record all http errors
[...]

PR #573, opened 2019-08-16
Issue #568: Changed the logic in the CircuitBreaker how to handle exceptions.

RobWin:
Changed the logic in the CircuitBreaker how to handle exceptions so that it is also possible to count exceptions as a success.
The list of ignored exceptions has always precedence. If an exception is ignored it neither counts as a success nor failure.
If the list of recorded exceptions only contains some exceptions, all others count as a success, unless they are part of the list of ignored exceptions.


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -100,8 +104,12 @@ public int getSlidingWindowSize() {
         return slidingWindowSize;
     }
 
-    public Predicate<Throwable> getRecordFailurePredicate() {
```
Romeh: May we can add deprecated instead and let it return recordExceptionPredicate as well to support backward compatible code with cb config object  ?

RobWin: Actually it's dangerous, because the implementation has also changed. I think it should be even a non public method or?

RobWin: It's a non backward compatible change.

PR #579, opened 2019-08-17
Support of @CircuitBreaker annotation along with @FeignClient

harishmbm:


RobWin:
Hi, 
please add a test which verifies that the changes work.

harishmbm:
@RobWin  Please see it again, test case added to support the functionality

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
```diff
@@ -132,7 +133,21 @@ private CircuitBreaker getCircuitBreakerAnnotation(ProceedingJoinPoint proceedin
...
+			return extractCircuitBreakerAnnotationFromProxyClass(proceedingJoinPoint);
+		} else {
+			return AnnotationExtractor.extract(proceedingJoinPoint.getTarget().getClass(), CircuitBreaker.class);
+		}
+	}
+
+	@Nullable
+	private CircuitBreaker extractCircuitBreakerAnnotationFromProxyClass(ProceedingJoinPoint proceedingJoinPoint) {
+		if (proceedingJoinPoint.getTarget().getClass().getInterfaces().length > 0) {
+			return AnnotationExtractor.extract(proceedingJoinPoint.getTarget().getClass().getInterfaces()[0], CircuitBreaker.class);
```
RobWin: Does that work, if the Proxy implements multiple interfaces

harishmbm: In case of FeignClient the proxy will be implementing only single interface i.e. the original Feign interface. I cannot assume any case where there could be multiple interfaces while using a @FeignClient annotation. 

Though it could be a valid case with some other integrations rather than openfeign. In such case this condition should be converted to a for loop to loop all the interfaces to find the @CircuitBreaker annotation and return from the first match. Please suggest if this case is a valid scenario to cover here, I'll update the code if you suggest.

RobWin: In my opinion we should support proxies which implement multiple interfaces.

harishmbm: Done @RobWin, kindly check again.

RobWin:
Could you implement it as well for RateLimiter, Retry and Bulkhead?

RobWin:
This Twitter tweet is interesting. We are facing the same issues with the current implementation I guess.

https://twitter.com/openwms/status/1164203398318690307?s=19

harishmbm:
I would need time to understand these patterns and I would try to implement it over weekend in a separate PR. Could you please merge this PR first and let me know when the release 1.0.0 is scheduled because this specific feature is needed for my production code.

RobWin:
There are two other PRs which we want to include into a 1.0.0.
But I discuss it internally if we can move them to a 1.1.0.

harishmbm:
@RobWin I have implemented the same for Retry, ratelimiter and bulkhead too. Kindly review the changes and merge it.

Issue #531, opened 2019-07-08
proposal : Metrics in resilience4j-timelimiter


fvigotti:
Hi,
I see that there are no registry nor metrics for timelimiter which would be very useful  :) I want to leave my upvote if someone have time to implement it would be nice,  ( and the only advantage vs other methods in kotlin such as `withTimeout` )

RobWin:
Would you like to create a PR?

PR #583, opened 2019-08-20
#531 Time Limiter Events

hexmind:
#531 - stage 1 (events)

### DIFF
--- a/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
+++ b/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
@@ -1,9 +1,15 @@
 package io.github.resilience4j.timelimiter;
 
+import io.github.resilience4j.core.EventConsumer;
+import io.github.resilience4j.timelimiter.event.TimeLimiterEvent;
+import io.github.resilience4j.timelimiter.event.TimeLimiterOnFailureEvent;
+import io.github.resilience4j.timelimiter.event.TimeLimiterOnSuccessEvent;
+import io.github.resilience4j.timelimiter.event.TimeLimiterOnTimeoutEvent;
 import io.github.resilience4j.timelimiter.internal.TimeLimiterImpl;
 
 import java.time.Duration;
-import java.util.concurrent.*;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
 import java.util.function.Supplier;
 
 /**
@@ -11,13 +17,15 @@ import java.util.function.Supplier;
  */
 public interface TimeLimiter {
 
+    String DEFAULT_NAME = "UNDEFINED";
+
     /**
      * Creates a TimeLimiter decorator with a default TimeLimiterConfig configuration.
      *
      * @return The {@link TimeLimiter}
      */
     static TimeLimiter ofDefaults() {
-        return new TimeLimiterImpl(TimeLimiterConfig.ofDefaults());
+        return new TimeLimiterImpl(DEFAULT_NAME, TimeLimiterConfig.ofDefaults());
     }
 
     /**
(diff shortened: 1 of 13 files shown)


resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -1,23 +1,31 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
 import java.util.function.Supplier;
 
 /**
  * A TimeLimiter decorator stops execution after a configurable duration.
  */
 public interface TimeLimiter {
 
+    String DEFAULT_NAME = "TimeLimiter";
```
hexmind: or maybe better "UNDEFINED" or null?

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -27,7 +35,18 @@ static TimeLimiter ofDefaults() {
...
+    }
+
+    /**
+     * Creates a TimeLimiter decorator with a TimeLimiterConfig configuration.
+     *
+     * @param name the String
+     * @param timeLimiterConfig the TimeLimiterConfig
+     * @return The {@link TimeLimiter}
+     */
+    static TimeLimiter of(String name, TimeLimiterConfig timeLimiterConfig) {
```
hexmind: Should I overload other constructors and factory methods (adding version with `name` parameter)?

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -54,26 +72,7 @@ static TimeLimiter of(Duration timeoutDuration) {
      * @return a future supplier which is restricted by a {@link TimeLimiter}.
      */
     static <T, F extends Future<T>> Callable<T> decorateFutureSupplier(TimeLimiter timeLimiter, Supplier<F> futureSupplier) {
-        return () -> {
```
hexmind: I moved the body to implementation class mainly because of `publishTimeLimiterEvent` call but also because all logic was in a static method in an interface instead of impl.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -27,7 +35,18 @@ static TimeLimiter ofDefaults() {
...
      */
     static TimeLimiter of(TimeLimiterConfig timeLimiterConfig) {
-        return new TimeLimiterImpl(timeLimiterConfig);
+        return of(DEFAULT_NAME, timeLimiterConfig);
+    }
+
+    /**
+     * Creates a TimeLimiter decorator with a TimeLimiterConfig configuration.
+     *
+     * @param name the String
```
RobWin: `the name of the TimeLimiter`

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
RobWin: Yes, I agree. UNDEFINED might be better.

hexmind:
> The TimeLimiterOperator is of course out of the scope of this PR.
> Also Metrics for the TimeLimiter are out of the scope of this PR.

Out of scope but ` Requested changes must be addressed to merge this pull request`...
Do you see the need for other metrics than `getNumberOfXCalls` (Successful/Failed/TimedOut)?

RobWin:
Yes, the Operator is out of scope, but we can already implement the methods onSuccess, onError, onTimeout.

RobWin:
No, i think other metrics are not necessary.

PR #586, opened 2019-08-24
Issue #531: Added metrics to TimeLimiter

hexmind:
#531


RobWin:
Thank you very much for this PR.
Our module `resilience4j-metrics` is a little bit outdated compared to `resilience4j-micrometer`.  Because many users have migrated from Dropwizard to Micrometer. The micrometer module (e.g. TaggedCircuitBreakerMetrics) is more advanced and allows us to dynamically add and remove metrics.

Could you please also add the new metrics to `resilience4j-micrometer` and `resilience4j-prometheus`?

hexmind:
Sure but maybe in next PR. There are `Files changed 14 ` already.
I will also apply current comments and suggestions to future micrometer changes.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/utils/MetricNames.java
```diff
@@ -0,0 +1,8 @@
+package io.github.resilience4j.timelimiter.utils;
+
+public class MetricNames {
+    public static final String DEFAULT_PREFIX = "resilience4j.timelimiter";
+    public static final String SUCCESSFUL_CALLS = "number_of_successful_calls";
```
RobWin: Can we please use the same names like the CircuitBreaker:

 public static final String SUCCESSFUL = "successful";
 public static final String FAILED = "failed";
 public static final String TIMEOUT = "timeout";

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -79,23 +100,53 @@ public void onSuccess() {
...
-}
+    public class TimeLimiterMetrics implements Metrics {
+
+        @Override
+        public long getNumberOfSuccessfulCalls() {
+            return successes.longValue();
+        }
+
+        @Override
+        public long getNumberOfErrorCalls() {
```
RobWin: Please rename to  `getNumberOfFailedCalls`, but only if we really measure metrics in the TimeLimiter itself.
I know it's confusing, because the public method is called `onError` and the event `TimeLimiterOnErrorEvent`, but I would like to use the same names like the CircuitBreaker uses.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -6,27 +6,37 @@
...
+    private final LongAdder successes;
+    private final LongAdder errors;
+    private final LongAdder timeouts;
+    private final TimeLimiterMetrics metrics;
     private final TimeLimiterEventProcessor eventProcessor;
 
     public TimeLimiterImpl(String name, TimeLimiterConfig timeLimiterConfig) {
         this.name = name;
         this.timeLimiterConfig = timeLimiterConfig;
+        this.successes = new LongAdder();
```
RobWin: I've a mixed feeling if we should really measure successful, failed and timeout calls here.
It means a TimeLimiter measures calls even if the user is not interested in the metrics. Other components like RateLimiter, CircuitBreaker or Bulkhead can provide some internal metrics, because they need some them anyway to function properly. And other some other metrics of the CircuitBreaker or RateLimiter are measured by consiming the events. Please see [TaggedCircuitBreakerMetrics](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/TaggedCircuitBreakerMetrics.java#L130-L134).

I think I would prefer that a TimeLimiter itself is lightweight and does not measure metrics itself. The `TimeLimiterMetrics` class can register event consumers and increases counters.

What is your opinion?
@Romeh @hexmind @storozhukBM

hexmind: You are right, only `cache` and `retry` components record and store metrics. Maybe we can provide metrics on demand only if the user is not interested in the metrics? Something like below:
```
TimeLimiterImpl
// ...
@Override
    public Metrics createMetrics() {
        return new Metrics() {
            private final LongAdder successes = new LongAdder();
            private final LongAdder errors = new LongAdder();
            private final LongAdder timeouts = new LongAdder();

            {
                getEventPublisher().onSuccess(e -> successes.increment());
                getEventPublisher().onError(e -> errors.increment());
                getEventPublisher().onTimeout(e -> timeouts.increment());
            }

            @Override
            public long getNumberOfSuccessfulCalls() {
                return successes.longValue();
            }

            @Override
            public long getNumberOfFailedCalls() {
                return errors.longValue();
            }

            @Override
            public long getNumberOfTimedOutCalls() {
                return timeouts.longValue();
            }
        };
    }
```

RobWin: I would prefer to implement it similar to the other components and register event consumers in the Dropwizard Metrics / MicroMeter / Prometheus implementations.

hexmind: I will keep only an interface in time limiter and add implementations in proper modules later.

RobWin: You can delete the interface. The metrics module is independent.

hexmind:
@RobWin It is ready for review.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -121,6 +128,25 @@ static TimeLimiter of(Duration timeoutDuration) {
      */
     void onError(Throwable throwable);
 
+
+    interface Metrics {
```
RobWin: Could you please delete this interface

RobWin:
AWesome thanks,

RobWin:
Would you like to create a PR for prometheus and micrometer as well?

hexmind:
Yeah. And TimeLimiterOperator also.

Issue #571, opened 2019-08-16
HealthIndicator with registry


dlsrb6342:
### Enhancement
Currently, in resilience4j-spring-boot2, `HealthIndicator` is made with each `CircuitBreaker` or `RateLimiter`. 

If make `HealthIndicator` with Registry, it doesn't need to add `HealthIndicator` Singleton on each time `CircuitBreaker` is created.
We can just return with `CircuitBreakerRegistry.getAllCircuitBreakers()`.
```java
public class CircuitBreakerHealthIndicator implements HealthIndicator {
	
	private final CircuitBreakerRegistry circuitBreakerRegistry;

	@Override
	public Health health() {
		return Optional.of(CircuitBreakerRegistry.getAllCircuitBreakers())
			...;
	}
} 
```

With [`CompositeHealthIndicatorConfiguration`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/health/CompositeHealthIndicatorConfiguration.java), there is no need to register to `HealthIndicatorRegistry` manually. So it would be more easier to control `HealthIndicator`.

Additionally, without `org.springframework.boot:spring-boot-starter-actuator`, resilience4j-spring-boot2 cannot run. Error log is below. 
```
Caused by: java.lang.ClassNotFoundException: org.springframework.boot.actuate.health.HealthIndicatorRegistry
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583) ~[na:na]
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178) ~[na:na]
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521) ~[na:na]
	... 48 common frames omitted
```

RobWin:
Yes, true. Would you like to create a PR? But please also have a look at `RateLimiterHealthIndicator`
We can also use `@Conditional(value = {HealthIndicator.class})` to make the beans optional.
But we should also add `@Conditional(value = {Endpoint.class})` to the custom Actuator endpoints then.

dlsrb6342:
Ok I'll submit PR soon. 
And one more thing, only CircuitBreaker and RateLimiter have `HealIndicator`. 
Bulkhead or Retry doesn't need it?

RobWin:
[...]

PR #550, opened 2019-07-29
WIP : Add resilience4j-spring-cloud module

dlsrb6342:
Issue resilience4j#546

Registry Bean declarations are moved to ConfigurationMissingBean from AbstractConfigurationMissingBean. Because spring-cloud-context versions are different between boot1 and boot2.

#### Example Repo
https://github.com/dlsrb6342/resilience4j-spring-cloud-config
https://github.com/dlsrb6342/config-test

I tested with above project, and it works well.

PR #587, opened 2019-08-25
HealthIndicators with registry

dlsrb6342:
Issue #571

### Changes
* HealthIndicator with registry not instance.
  * `RateLimiterHealthIndicator` -> `RateLimitersHealthIndicator`
  * `CircuitBreakerHealthIndicator` -> `CircuitBreakersHealthIndicator`
* Register HealthIndicator as Bean in `HealthIndicatorAutoConfiguration`.
* actuator health endpoint response will be changed. Please refer https://github.com/resilience4j/resilience4j/issues/571#issuecomment-522309022.

After merge #550 I will edit spring-cloud module.


resilience4j-spring-boot/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/BulkheadAutoConfiguration.java
```diff
@@ -42,15 +42,19 @@
 @AutoConfigureBefore(EndpointAutoConfiguration.class)
 public class BulkheadAutoConfiguration {
 
-	@Bean
+	@Configuration
 	@ConditionalOnClass(value = {Endpoint.class})
-	public BulkheadEndpoint bulkheadEndpoint(BulkheadRegistry bulkheadRegistry, ThreadPoolBulkheadRegistry threadPoolBulkheadRegistry) {
-		return new BulkheadEndpoint(bulkheadRegistry, threadPoolBulkheadRegistry);
-	}
+	public static class BulkheadEndpointConfiguration {
```
dlsrb6342: I wrapped Endpoint Bean with Configuration. In spring-boot1, it causes Error because of `AbstractEndpoint` which is used in custom endpoint class.

resilience4j-timelimiter/src/test/java/io/github/resilience4j/timelimiter/TimeLimiterEventPublisherTest.java
```diff
@@ -71,7 +71,7 @@ public void shouldConsumeOnTimeoutEvent() {
         timeLimiter.getEventPublisher()
                 .onTimeout(event -> logger.info(event.getEventType().toString()));
         Supplier<CompletableFuture<String>> futureSupplier = () ->
-                CompletableFuture.supplyAsync(this::fail);
+                CompletableFuture.supplyAsync(this::timeout);
```
dlsrb6342: I added `timeout()` because `fail()` doesn't make TIMEOUT event. It makes ERROR event.

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/circuitbreaker/monitoring/health/CircuitBreakersHealthIndicator.java
```diff
@@ -37,20 +41,35 @@
...
-                .map(this::mapBackendMonitorState)
-                .orElse(Health.up().build());
+        Map<String, Health> healths = circuitBreakerRegistry.getAllCircuitBreakers().toJavaStream()
+                .filter(this::isRegisterHealthIndicator)
+                .collect(Collectors.toMap(CircuitBreaker::getName, CircuitBreakersHealthIndicator::mapBackendMonitorState));
+
+        return healthAggregator.aggregate(healths);
+    }
+
+    private boolean isRegisterHealthIndicator(CircuitBreaker circuitBreaker) {
```
dlsrb6342: This method is used to check `registerHealthIndicator` value. It is same with `RateLimitersHealthIndicator`. 
In this method, `circuitBreakerProperties.findCircuitBreakerProperties(name)` always null  with **dynamic instance** because dynamic instance doesn't have any InstanceProperties. 
So **dynamic instance always register healthIndicator**. Is it ok?

RobWin: Yes, that should be ok.

PR #606, opened 2019-09-08
Add MetricsPublisher interface

### LINKED ISSUE #603: Metrics Publisher on each instance
About https://github.com/resilience4j/resilience4j/pull/550#issuecomment-526825713, 
Tagged*Metrics cannot collect metrics after Registry refreshed in spring-cloud. 

With Metrics Publisher on each instance, it could be resolved. 
Please refer [HystrixMetricsPublisherFactory](https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/strategy/metrics/HystrixMetricsPublisherFactory.java) and [HystrixMetricsPublisher](https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/strategy/metrics/HystrixMetricsPublisher.java)

```java
public class MetricsPublisherFactory<E> {
	// Micrometer, Prometheus, Dropwizard Metric
	private List<MetricsPublisher> metricsPublishers;

	public void getOrCreateMetricsPublisher(E entry) { ... }

	public void removeMetricsPublisher(E entry) { ... }
	
}

// AbstractRegistry
public class AbstractRegistry<E, C> implements Registry<E, C> {
	
	private MetricsPublisherFactory metricsPublisherFactory;

	protected E computeIfAbsent(String name, Supplier<E> supplier){
		return entryMap.computeIfAbsent(Objects.requireNonNull(name, NAME_MUST_NOT_BE_NULL), k -> {
			E entry = supplier.get();
			eventProcessor.processEvent(new EntryAddedEvent<>(entry));
			metricsPublisherFactory.getOrCreateMetricsPublisher(entry);
			return entry;
		});
	}

	@Override
	public Optional<E> remove(String name){
		Optional<E> removedEntry = Optional.ofNullable(entryMap.remove(name));
		removedEntry.ifPresent(entry -> eventProcessor.processEvent(new EntryRemovedEvent<>(entry)));
		metricsPublisherFactory.removeMetricsPublisher(entry);
		return removedEntry;
	}

}
```

In this way, resilience4j would be better to use with spring-cloud.

dlsrb6342:
Resolves #603

### Changes
* Add `MetricsPublisher` interface
* For backward compatibility, I preserved old metrics classes.
  * Setting `resilience4j.**.metrics.legacy.enabled` as true, enable old one and disable new one.
* Add `MetricsPublisher` implementation on resilience4j-prometheus, resilience4j-micrometer, resilience4j-metrics
* Add `CompositeMetricsPublisher` on `AbstractRegistry`.
  * Default `CompositeMetricsPublisher` have empty `metricPublishers` list.

I tested this with spring-boo1 and spring-boot2. It works well. Also it can add dynamic instances.
After review, I'll add some tests about `MetricsPublisher`.


RobWin:
Thank you.
Could you please add the Apache License Headers to all new files?

dlsrb6342:
Added.
One question, Why is `ThreadPoolBulkheadMetrics` excluded from spring modules?

RobWin:
I think I implemented ThreadPoolBulkheadMetrics in this release and did not have time yet to add it to Spring Boot modules. Would you like to add it?

dlsrb6342:
@RobWin 
#### Changes
* Added AutoConfiguration of `ThreadPoolBulkheadMetrics`.
* Rearrange ConditionalOnClass on Metrics Auto Configuration.

RobWin:
Thank you

RobWin:
I reviewed the PR and I would like to discuss my ideas:
The AbstractRegistry already has RegistryEventProcessor which publishes events when an entry has been added, removed or replaced. 
The PR introduces a new interface `MetricsPublisher` which must be added to the Registry. Internally the MetricsPublisher acts similar to the RegistryEventProcessor. 

Couldn't we instead revert the logic and register a EventConsumers to the registry, after the registry has been created? This would loosely couple both components via the `RegistryEventProcessor`.
That way we just need a `CircuitBreakerMetricsPublisher` bean with `@RefreshScope`, or?

```
CircuitBreakerMetricsPublisher
     .ofCircuitBreakerRegistry(circuitBreakerRegistry, meterRegistry);
```

and internally

```
private CircuitBreakerMetricsPublisher(MetricNames names, CircuitBreakerRegistry circuitBreakerRegistry, MeterRegistry meterRegistry) {
        this(names, circuitBreakerRegistry);
        this.circuitBreakerRegistry.getEventPublisher()
                .onEntryAdded(entry -> addMetrics(meterRegistry, entry.getAddedEntry()));
        this.circuitBreakerRegistry.getEventPublisher()
                .onEntryRemoved(entry -> removeMetrics(meterRegistry, entry.getRemovedEntry().getName()));
        this.circuitBreakerRegistry.getEventPublisher()
                .onEntryReplaced(entry -> {
                    removeMetrics(meterRegistry, entry.getOldEntry().getName());
                    addMetrics(meterRegistry, entry.getNewEntry());
                });
    }
```

What do you think?

dlsrb6342:
Good. I think your code seems to work well. I'll change `MetricsPublisher` to use `RegistryEventProcessor` and test it.

dlsrb6342:
I want to support custom `MetricsPublisher`. If someone wants to add custom metrics, just add new `MetricsPublisher` implementation as a Bean. `CompositeMetricsPublisher` will collect `MetricsPublisher` beans. 

So adding `eventConsumer` only on `CompositeMetricsPublisher` will be the solution that achieve both your idea and my idea.
```java
public CompositeMetricsPublisher(List<MetricsPublisher<E>> delegates, AbstractRegistry<E, ?> abstractRegistry) {
    this.delegates = new ArrayList<>(requireNonNull(delegates));
    abstractRegistry.getEventPublisher().onEntryAdded(event -> publishMetrics(event.getAddedEntry()));
    abstractRegistry.getEventPublisher().onEntryRemoved(event -> removeMetrics(event.getRemovedEntry()));
    abstractRegistry.getEventPublisher().onEntryReplaced(event -> {
        removeMetrics(event.getOldEntry());
        publishMetrics(event.getNewEntry());
    });
}
```

What do you think about this?

RobWin:
Sounds good, but the `CompositeMetricsPublisher` is a separate bean and the `AbstractRegistry` doesn't know anything about this class, or?

dlsrb6342:
Yes `CompositeMetricsPublisher` is a separate bean. But I tested that code. It doesn't work.

Bean with `@RefreshScoped` is refreshed when it is used after Refresh event occurs. But with upper code, `CompositeMetricsPublisher`  bean is never called after Refresh event because refreshed `AbstractRegistry.RegistryEventProcessor` is empty.

dlsrb6342:
What do you think creating `RegistryEventProcessor` as Bean? If we do that, we don't need to re-register `EventConsumer` on `RegistryEvenProcessor`.

RobWin:
> Yes CompositeMetricsPublisher is a separate bean. But I tested that code. It doesn't work.
Bean with @RefreshScoped is refreshed when it is used after Refresh event occurs. But with upper code, CompositeMetricsPublisher bean is never called after Refresh event because refreshed AbstractRegistry.RegistryEventProcessor is empty.

Not sure if I understand it correct. The CompositeMetricsPublisher bean is not created again, after the config has changed?

dlsrb6342:
Yes. For re-creating bean, bean's any method have to be called. But after refresh, AbstractRegistry's RegistryEventProcessor is reset. So CompositeMetricsPublisher bean is never called.

RobWin:
Even if the bean `CompositeMetricsPublisher` itself has `@RefreshScope`?

RobWin:
I think we should not extract the `RegistryEventProcessor` as a separate bean.

RobWin:
Maybe should just stick with for current implementation.

RobWin:
Unfortunately the `EventConsumer<RegistryEvent>` is not really helpful.
The alternative would be to register three consumers:

- EventConsumer<EntryAddedEvent<E>>
- EventConsumer<EntryRemovedEvent<E>>
- EventConsumer<EntryReplacedEvent<E>>

But I think that's not very nice.
I actually like the initial MetricsPublisher-based solution more and more ;)

dlsrb6342:
The cause of all this is AbstractRegistry's constructor.

```java
public AbstractRegistry(C defaultConfig) {
	this.configurations = new ConcurrentHashMap<>();
	this.entryMap = new ConcurrentHashMap<>();
	
	// this code make empty RegistryEventProcessor after refresh
	this.eventProcessor = new RegistryEventProcessor();
	
	this.configurations.put(DEFAULT_CONFIG, Objects.requireNonNull(defaultConfig, CONFIG_MUST_NOT_BE_NULL));
}
```

If we can pass eventConsumer to RegistryEventProcessor in constructor, these issue will be resolved

RobWin:
Which EventConsumer?

dlsrb6342:
I thought all three eventConsumer as you commented. But I think initial one is better than this too. 
I will check more about using RegistryEventConsumer. 

Why do you think that we should not extract RegistryEventConsumer as a bean?

dlsrb6342:
Sorry for close.

RobWin:
>Why do you think that we should not extract RegistryEventConsumer as a bean?

Are you talking about the internal RegistryEventProcessor in `AbstractRegistry`? I think it should be an internal detail of the Registry and not an external EvenProcessor. 
Or are you talking about something different. This time you called it `RegistryEventConsumer`?

dlsrb6342:
Sorry for typo. I mean internal RegistryEventProcessor. 

> > Why do you think that we should not extract RegistryEventConsumer as a bean?
> 
> Are you talking about the internal RegistryEventProcessor in `AbstractRegistry`? I think it should be an internal detail of the Registry and not an external EvenProcessor.

In this point, I agree with you. How about support the way that add some eventConsumer to RegistryEventProcessor in AbstractRegistry's constructor?

RobWin:
The solution to add event consumers to the constructor of a Registry (CircuitBreakerRegistry) would be advantageous to the `MetricsPublisher` interface, because it could be used for other use cases as well. Not only for publishing metrics.

dlsrb6342:
Yes I agree with that.

```java 
public interface RegistryEventConsumer<E> {
	void consumeEntryAddedEvent(EntryAddedEvent<E> entryAddedEvent);
	void consumeEntryRemovedEvent(EntryRemovedEvent<E> entryRemoveEvent);
	void consumeEntryReplacedEvent(EntryReplacedEvent<E> entryReplacedEvent);
}

public AbstractRegistry(C defaultConfig, List<RegistryEventConsumer<E>> registryEventConsumers) {
	this.configurations = new ConcurrentHashMap<>();
	this.entryMap = new ConcurrentHashMap<>();
	this.eventProcessor = new RegistryEventProcessor(registryEventConsumers);
	this.configurations.put(DEFAULT_CONFIG, Objects.requireNonNull(defaultConfig, CONFIG_MUST_NOT_BE_NULL));
}
```
Code will be like this. As you said, it would be useful for any cases.

RobWin:
Yes, very nice. That's perfect

dlsrb6342:
Good. I will apply what we have discussed.

RobWin:
Btw, if you want to contribute even more to the project, I can invite you into our Slack channel.

dlsrb6342:
Yes I really want. It would be great if you invited me. 
Do you need any information for invitation?

RobWin:
Done

dlsrb6342:
@storozhukBM 
Merged latest master.

storozhukBM:
Thank you

PR #619, opened 2019-09-16
[Feedback wanted] Disable stack traces for CallNotPermittedExceptions

### LINKED ISSUE #618: Feature: Expose option to suppress stack traces
In high throughput systems, the opening of a circuit breaker leads to log spam and increases the pressure on disk iops.
In most cases, the stack trace produced does not add any value as the root cause is already known: the execution failed because the circuit breaker was open/half open.

My request is to expose a config option that would allow users of this library to specify if they want the stack trace to be suppressed for CallNotPermittedExceptions. I believe this would be a simple and easy way to address this issue.

I'm happy to raise a pr if there is interest in adding this feature.

worldtiki:
Fixes #618.

Created a static method that constructs a regular CallNotPermitedException or a similar one but  without a stack trace,

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
@@ -25,13 +25,23 @@ package io.github.resilience4j.circuitbreaker;
 public class CallNotPermittedException extends RuntimeException {
 
     /**
-     * The constructor with a CircuitBreaker.
+     * Static method to construct a {@link CallNotPermittedException} with a CircuitBreaker.
      *
      * @param circuitBreaker the CircuitBreaker.
      */
-    public CallNotPermittedException(CircuitBreaker circuitBreaker) {
-        super(String.format("CircuitBreaker '%s' is %s and does not permit further calls", circuitBreaker.getName(), circuitBreaker.getState()));
+    public static CallNotPermittedException getCallNotPermittedException(CircuitBreaker circuitBreaker) {
+        boolean disableStackTraces = circuitBreaker.getCircuitBreakerConfig().isDisableStackTraces();
+
+        String message = String.format("CircuitBreaker '%s' is %s and does not permit further calls", circuitBreaker.getName(), circuitBreaker.getState());
+
+        if (disableStackTraces) {
+            return new CallNotPermittedException(message, false);
+        } else {
+            return new CallNotPermittedException(message, true);
+        }
+    }
+
+    private CallNotPermittedException(String message, boolean writableStackTrace) {
+        super(message, null, false, writableStackTrace);
     }
 }
-
-
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
@@ -30,15 +30,11 @@ public class CallNotPermittedException extends RuntimeException {
      * @param circuitBreaker the CircuitBreaker.
      */
     public static CallNotPermittedException getCallNotPermittedException(CircuitBreaker circuitBreaker) {
-        boolean disableStackTraces = circuitBreaker.getCircuitBreakerConfig().isDisableStackTraces();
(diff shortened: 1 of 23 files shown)


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
```diff
@@ -25,13 +25,23 @@
...
-    public CallNotPermittedException(CircuitBreaker circuitBreaker) {
-        super(String.format("CircuitBreaker '%s' is %s and does not permit further calls", circuitBreaker.getName(), circuitBreaker.getState()));
-    }
-}
+    public static CallNotPermittedException getCallNotPermittedException(CircuitBreaker circuitBreaker) {
+        boolean disableStackTraces = circuitBreaker.getCircuitBreakerConfig().isDisableStackTraces();
+
+        String message = String.format("CircuitBreaker '%s' is %s and does not permit further calls", circuitBreaker.getName(), circuitBreaker.getState());
 
+        if (disableStackTraces) {
```
RobWin: Can we use `return new CallNotPermittedException(message, disableStackTrace);` instead?

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/configuration/CircuitBreakerConfigurationProperties.java
```diff
@@ -211,6 +215,9 @@ public InstanceProperties getBackendProperties(String backend) {
 		@Nullable
 		private Boolean automaticTransitionFromOpenToHalfOpenEnabled;
 
+		@Nullable
+		private Boolean disableStackTraces;
```
RobWin: Can we rename it to `writableStackTraceEnabled` analog to `automaticTransitionFromOpenToHalfOpenEnabled`.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -174,6 +181,7 @@ public Builder(CircuitBreakerConfig baseConfig) {
             this.automaticTransitionFromOpenToHalfOpenEnabled = baseConfig.automaticTransitionFromOpenToHalfOpenEnabled;
             this.slowCallRateThreshold = baseConfig.slowCallRateThreshold;
             this.slowCallDurationThreshold = baseConfig.slowCallDurationThreshold;
+            this.disableStackTraces = baseConfig.disableStackTraces;
```
RobWin: Can we rename it to `writableStackTraceEnabled` and set it to `true` by default in order to be backward compatible.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -216,6 +224,17 @@ public Builder slowCallRateThreshold(float slowCallRateThreshold) {
             return this;
         }
 
+        /**
+         * Disables writable stack traces. When set to true, {@link Exception#getStackTrace()} returns a zero length array.
+         * This may be used to reduce log spam when the circuit breaker is open as the cause of the exceptions is already known (the circuit breaker is short-circuiting calls).
+         *
+         * @return the CircuitBreakerConfig.Builder
+         */
+        public Builder disableStackTraces(boolean disableStackTraces) {
```
RobWin: Please rename to `writableStackTraceEnabled(...)`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -120,6 +122,10 @@ public int getMinimumNumberOfCalls() {
         return minimumNumberOfCalls;
     }
 
+    public boolean isDisableStackTraces() {
```
RobWin: Please rename to `isWritableStackTraceEnabled`

RobWin:
Awesome thanks.
Would you like to implement a similar feature for Bulkhead, RateLimiter in an another PR?

worldtiki:
> Awesome thanks.
> Would you like to implement a similar feature for Bulkhead, RateLimiter in an another PR?

Sure.

Thanks for reviewing the changes so quickly :)

PR #620, opened 2019-09-16
Fix non deterministic test

worldtiki:
Depending on the number of cpus, the default core threadpool size can be less than 10 which causes the test to fail.

The default is calculated in ThreadPoolBulkheadConfig:
`DEFAULT_CORE_THREAD_POOL_SIZE = Runtime.getRuntime().availableProcessors() > 1 ? Runtime.getRuntime().availableProcessors() - 1 : 1;`

If this is less than 10, the test fails with the error :

> maxThreadPoolSize must be a greater than or equals to coreThreadPoolSize


RobWin:
Thx

PR #621, opened 2019-09-16
#625 Fix for some issues raised by sonar

storozhukBM:


RobWin:
Shouldn't we use `AcquirePermissionCancelledException` in the SemaphoreBulkhead  as well?

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
```diff
@@ -209,6 +222,7 @@ boolean tryEnterBulkhead() {
             try {
                 callPermitted = semaphore.tryAcquire(timeout, TimeUnit.MILLISECONDS);
             } catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
```
RobWin: Shouldn't we throw a `AcquirePermissionCancelledException` as well instead of setting the interrupt flag?

RobWin: It is not backward-compatible, but I think it's more of a bugfix than API change.

storozhukBM: I'm not sure it is a good idea to throw exceptions from this "low-level" method.
The idea was that the user would be able to handle it in their code. And if the caller thread is a part of some thread pool, the standard library code will handle this flag properly and so on.

RobWin: But what's the difference to RateLimiter.waitForPermission?

storozhukBM: My thoughts process behind it was the following:
We have "low-level"/"raw" methods in our components that are unopinionated and designed to be used from different programming paradigms blocking/non-blocking, threads/fibers/actors, different error handling mechanisms like exceptions/monads/error-values etc.
For `RateLimiter` such methods are the following: `acquirePermission`, `reservePermission`.
For `Bulkhead`: `tryAcquirePermission`, `acquirePermission`, `releasePermission`, `onComplete` and package-private `tryEnterBulkhead`.

But at the same time, our library provides an opinionated set of decorators and helpers that are "high-level". But users aren't forced to use these, and here we can throw exceptions or cancel Futures and so on.
And `RateLimiter.waitForPermission` is the static helper methods created for blocking decorators and others.

RobWin: ok, I understand now why the interrupt flag must be set again, but still we should add something like

```
throw new AcquirePermissionCancelledException("Thread was interrupted during permission wait");
```

storozhukBM: As for `acquirePermission` we also consider it a low-level method and actual exception throwing we offload to decorators. 
Notice that `SempahoreBulkhead.acquirePermission` is a method defined by each implementation of `Bulkhead` interface, and `RateLimiter.waitForPermission` is a static helper method defined in actual `RateLimiter` interface.

RobWin: `SempahoreBulkhead.tryAcquirePermission` is returning a boolean and is offloading exception throwing to the client. Comparable to `Semaphore.tryAcquire`.

But `SempahoreBulkhead.acquirePermission` is throwing a BulkheadFullException, if `tryAcquirePermission` is returning false. Means it is throwing a `BulkheadFullException`, even if the Thread was interrupted. I'm saying tat it should be an `AcquirePermissionCancelledException` instead of `BulkheadFullException`.

storozhukBM:
> Can we please move the message into the default constructor? And please change it to "Thread was interrupted while waiting for a permission"

Done)

Issue #235, opened 2018-06-05
TimeLimiter as rx operator


edudar:
Timeouts and time limiters are essential for resiliency and one of the primary reasons we use Hystrix in the first place. `TimeLimiterOperator`s are really missing in `-rxjava2` and `-reactor` packages.

RobWin:
Hi,

the RxJava and Reactor timeout operator is not sufficient for you?

http://reactivex.io/documentation/operators/timeout.html and https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#timeout-java.time.Duration-

edudar:
I can't say whether it's sufficient as a whole or not but I can make it work in some cases for sure. Same as I can make it with `debounce` and `retry`, for example, yet there are convenient operators for them wired into the whole resilience system. With only time limiter missing there, it feels inconsistent but if that's a design decision, fine.

PR #624, opened 2019-09-16
TimeLimiter as rx transformer (#235)

hexmind:
There is my proposal (only `Maybe` transformer so far) for a reactive extension of `TimeLimiter`. Please give me a feedback:

1. Is it what is expected?
2. Do we still need an operator, isn't transformer sufficient?


RobWin:
Hi,

yes.
Could you mock the TimeLimtier and also verify that `onSuccess` or `onError` are invoked when the stream completes or fails.

Could you use the [TestScheduler](http://reactivex.io/RxJava/javadoc/io/reactivex/schedulers/TestScheduler.html) to mock the clock and move the the Scheduler's clock to a particular moment in time.

hexmind:
The changes are ready but I need to rebase on #631 merged.

Issue #634, opened 2019-09-18
Circuitbreaker metrics is outdated


quaff:
I've created #632 to fix:
1. "buffered_max" is obsoleted
2. missing "slow" and "slow_call_rate"

PR #640, opened 2019-09-19
Add slow call count metrics

dlsrb6342:
### Issue #634 
Added slow call count metrics on micrometer and prometheus.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
@@ -545,7 +545,7 @@ public interface CircuitBreaker {
          *
          * @return the current number of successful calls which were slower than a certain threshold
          */
-        int getNumberOfSlowSuccessCalls();
+        int getNumberOfSlowSuccessfulCalls();
 
         /**
          * Returns the current number of failed calls which were slower than a certain threshold.
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
@@ -541,6 +541,20 @@ public interface CircuitBreaker {
         int getNumberOfSlowCalls();
 
         /**
+         * Returns the current number of successful calls which were slower than a certain threshold.
+         *
+         * @return the current number of successful calls which were slower than a certain threshold
+         */
+        int getNumberOfSlowSuccessCalls();
+
+        /**
+         * Returns the current number of failed calls which were slower than a certain threshold.
+         *
+         * @return the current number of failed calls which were slower than a certain threshold
+         */
+        int getNumberOfSlowFailedCalls();
+
+        /**
          * Returns the current total number of buffered calls in the ring buffer.
          *
          * @return he current total number of buffered calls in the ring buffer
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/MetricNames.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/MetricNames.java
@@ -5,6 +5,8 @@ public class MetricNames {
     public static final String SUCCESSFUL = "successful";
     public static final String FAILED = "failed";
     public static final String SLOW = "slow";
+    public static final String SLOW_SUCCESS = "slow_successful";
+    public static final String SLOW_FAILED = "slow_failed";
     public static final String NOT_PERMITTED = "not_permitted";
     public static final String BUFFERED = "buffered";
     public static final String STATE = "state";
--- a/resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/Snapshot.java
+++ b/resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/Snapshot.java
@@ -41,7 +41,21 @@ public interface Snapshot {
      *
      * @return the current number of calls which were slower than a certain threshold
      */
-    int getNumberOfSlowCalls();
+    int getTotalNumberOfSlowCalls();
+
+    /**
+     * Returns the current number of successful calls which were slower than a certain threshold.
+     *
+     * @return the current number of successful calls which were slower than a certain threshold
+     */
+    int getNumberOfSlowSuccessCalls();
+
+    /**
+     * Returns the current number of failed calls which were slower than a certain threshold.
+     *
+     * @return the current number of failed calls which were slower than a certain threshold
+     */
+    int getNumberOfSlowFailedCalls();
 
     /**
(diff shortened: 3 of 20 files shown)


resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/AbstractCircuitBreakerMetrics.java
```diff
@@ -59,6 +60,11 @@ protected void addMetrics(MeterRegistry meterRegistry, CircuitBreaker circuitBre
                 .tag(TagNames.NAME, circuitBreaker.getName())
                 .tag(TagNames.KIND, KIND_SUCCESSFUL)
                 .register(meterRegistry).getId());
+        idSet.add(Gauge.builder(names.getBufferedCallsMetricName(), circuitBreaker, cb -> cb.getMetrics().getNumberOfSlowCalls())
```
RobWin: Unfortunately this doesn't work.
Our sliding window stores successful and failed calls. A successful and a failed call can be fast or slow.
Means: totalNumberOfCalls = numberOfFailedCalls + numberOfSuccessfulCalls

When you visualize the metric `resilience4j.circuitbreaker.buffered.calls` without a tag filter in Grafana, it should show the total number of calls. If we add  the number of slow calls to it, we falsify the total number of calls.

The current number of slow calls must either be a separate metric like `resilience4j.circuitbreaker.slow.calls` or we have to change our sliding window and store `successful`, `successful_slow`, `failed` and `failed_slow`.

dlsrb6342: Oh I understand. I think store `successful`, `successful_slow`, `failed` and `failed_slow` is better. What do you think?

RobWin: Now it seems to be smarter. I wanted to reduce the amount of integer additions/substractions to the minimum in order to improve performance in [AbstractAggregation](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/AbstractAggregation.java) and  [TotalAggregation](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/TotalAggregation.java).

Actually even the number of successful calls isn't stored. Only `totalNumberOfCalls`, `numberOfFailedCalls` and `numberOfSlowCalls`. If we store `successful`, `successful_slow`, `failed` and `failed_slow`, you even always have to add `successful_slow + failed_slow` on every call to calculate the slow_call_rate. Another integer addition which I tried to avoid.
Questions is now. What is more important to us. Metrics or Performance?

dlsrb6342: I think Performance is more important. 
but.. what do you think count like below? 

```java
class AbstractAggregation {
    long totalDurationInMillis = 0;
    int numberOfSlowCalls = 0;
    int numberOfFailedCalls = 0;
    int numberOfCalls = 0;

    // Two integer added.
    int numberOfSuccessSlowCalls = 0;
    int numberOfFailedSlowCalls = 0;

}
```

We don't have to calculate `successful_slow + failed_slow ` on every calls. It is reasonable for performance I think.

RobWin: I think it would be sufficient to add `numberOfFailedSlowCalls`.
The performance to calculate the `slowCallRate` and `failureRate` in [SnapshotImpl](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/SnapshotImpl.java) would be unchanged.

And we can calculate `numberOfSuccessSlowCalls = numberOfSlowCalls - numberOfFailedSlowCalls`, only when needed.

dlsrb6342: Oh.. You are right. It would be better. I'll change that.

RobWin: But still, if we change that. How should the metrics look like?
If we change the names of the tag, it's not backward-compatible anymore.
What is your suggestion?

RobWin: Should we introduce a new metric like `resilience4j.circuitbreaker.slow.calls` with tag `KIND` and values:  `successful` and `failed`?

dlsrb6342: Yes. I think it is good. But `successful` can cause misunderstanding. It can be thought as NOT slow calls. Full naming is better like `SLOW_SUCCESS` and `SLOW_FAILED`.

dlsrb6342:
@RobWin 
Added slow failed and success metrics.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -540,6 +540,20 @@ public String toString(){
          */
         int getNumberOfSlowCalls();
 
+        /**
+         * Returns the current number of successful calls which were slower than a certain threshold.
+         *
+         * @return the current number of successful calls which were slower than a certain threshold
+         */
+        int getNumberOfSlowSuccessCalls();
```
RobWin: Please rename to `getNumberOfSlowSuccessfulCalls()`

resilience4j-core/src/main/java/io/github/resilience4j/core/metrics/Snapshot.java
```diff
@@ -41,7 +41,21 @@
...
      */
-    int getNumberOfSlowCalls();
+    int getTotalNumberOfSlowCalls();
+
+    /**
+     * Returns the current number of successful calls which were slower than a certain threshold.
+     *
+     * @return the current number of successful calls which were slower than a certain threshold
+     */
+    int getNumberOfSlowSuccessCalls();
```
RobWin: `getNumberOfSlowSuccessCalls()`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/MetricNames.java
```diff
@@ -5,6 +5,8 @@
     public static final String SUCCESSFUL = "successful";
     public static final String FAILED = "failed";
     public static final String SLOW = "slow";
+    public static final String SLOW_SUCCESS = "slow_successful";
```
RobWin: I think after renaming the core metric name to `.slow.calls` we could reuse the existing tags like `failed` and `slow`.

PR #654, opened 2019-10-01
WIP: Add tag API

siepkes:
Adds a tag API which can be used to add tags to various metric collecting backends like Prometheus, micrometer, etc.

Based upon discussion in issue #509.

### DIFF
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
@@ -23,6 +23,8 @@ import io.github.resilience4j.ratelimiter.RateLimiter;
 import io.github.resilience4j.ratelimiter.RateLimiterConfig;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnFailureEvent;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnSuccessEvent;
+import io.vavr.collection.HashMap;
+import io.vavr.collection.Map;
 import io.vavr.control.Option;
 
 import java.time.Duration;
@@ -49,6 +51,7 @@ public class SemaphoreBasedRateLimiter implements RateLimiter {
     private final ScheduledExecutorService scheduler;
     private final Semaphore semaphore;
     private final SemaphoreBasedRateLimiterMetrics metrics;
+    private final Map<String, String> tags;
     private final RateLimiterEventProcessor eventProcessor;
 
     /**
@@ -58,7 +61,18 @@ public class SemaphoreBasedRateLimiter implements RateLimiter {
      * @param rateLimiterConfig The RateLimiter configuration.
      */
     public SemaphoreBasedRateLimiter(final String name, final RateLimiterConfig rateLimiterConfig) {
-        this(name, rateLimiterConfig, null);
+        this(name, rateLimiterConfig, HashMap.empty());
+    }
+
+    /**
+     * Creates a RateLimiter.
+     *
+     * @param name              the name of the RateLimiter
+     * @param rateLimiterConfig The RateLimiter configuration.
+     * @param tags              tags to assign to the RateLimiter
+     */
+    public SemaphoreBasedRateLimiter(final String name, final RateLimiterConfig rateLimiterConfig, Map<String, String> tags) {
+        this(name, rateLimiterConfig, null, tags);
     }
 
     /**
@@ -70,10 +84,24 @@ public class SemaphoreBasedRateLimiter implements RateLimiter {
      */
     public SemaphoreBasedRateLimiter(String name, RateLimiterConfig rateLimiterConfig,
                                      @Nullable ScheduledExecutorService scheduler) {
+        this(name, rateLimiterConfig, scheduler, HashMap.empty());
+    }
+
+    /**
+     * Creates a RateLimiter.
+     *
+     * @param name              the name of the RateLimiter
+     * @param rateLimiterConfig The RateLimiter configuration.
+     * @param scheduler         executor that will refresh permissions
+     * @param tags              tags to assign to the RateLimiter
+     */
+    public SemaphoreBasedRateLimiter(String name, RateLimiterConfig rateLimiterConfig,
+                                     @Nullable ScheduledExecutorService scheduler, Map<String, String> tags) {
         this.name = requireNonNull(name, NAME_MUST_NOT_BE_NULL);
         this.rateLimiterConfig = new AtomicReference<>(requireNonNull(rateLimiterConfig, CONFIG_MUST_NOT_BE_NULL));
 
         this.scheduler = Option.of(scheduler).getOrElse(this::configureScheduler);
+        this.tags = tags;
         this.semaphore = new Semaphore(this.rateLimiterConfig.get().getLimitForPeriod(), true);
         this.metrics = this.new SemaphoreBasedRateLimiterMetrics();
 
@@ -199,6 +227,11 @@ public class SemaphoreBasedRateLimiter implements RateLimiter {
                 + "name='" + name + '\''
                 + ", rateLimiterConfig=" + rateLimiterConfig
                 + '}';
+    }
+
+    @Override
+    public Map<String, String> getTags() {
+        return tags;
     }
 
     /**
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
@@ -24,6 +24,8 @@ import io.github.resilience4j.retry.internal.RetryImpl;
 import io.vavr.CheckedFunction0;
 import io.vavr.CheckedFunction1;
 import io.vavr.CheckedRunnable;
+import io.vavr.collection.HashMap;
+import io.vavr.collection.Map;
 import io.vavr.control.Either;
 import io.vavr.control.Try;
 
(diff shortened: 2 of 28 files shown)


resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -350,6 +376,13 @@ static Runnable decorateRunnable(Retry retry, Runnable runnable) {
 	 */
 	RetryConfig getRetryConfig();
 
+	/**
+	 * Returns an unmodifiable map with tags assigned to this Retry.
+	 *
+	 * @return the tags assigned to this Retry in an unmodifiable map
+	 */
+	Map<String, String> getTags();
```
RobWin: Could we return an immutable Map from Vavr instead?

Comparable to `Seq<CircuitBreaker> getAllCircuitBreakers();`

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -55,9 +59,10 @@
     private final LongAdder succeededWithoutRetryCounter;
     private final LongAdder failedWithoutRetryCounter;
 
-	public RetryImpl(String name, RetryConfig config) {
+	public RetryImpl(String name, RetryConfig config, Map<String, String> tags) {
 		this.name = name;
 		this.config = config;
+		this.tags = Collections.unmodifiableMap(tags);
```
RobWin: Can we use a immutable Vavr map instead?

resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
siepkes: You mean `io.vavr.collection.Map`? Sounds good to me. Am I right to assume you want to use that everywhere? For example also in the `static Retry of(String name, RetryConfig retryConfig, Map<String, String> tags)` methods and such?

RobWin: Yes

siepkes:
@RobWin I've added the ability to add tags to a ` AbstractRegistry` and implemented it in the `InMemoryRetryRegistry` to show what that would look like. I've attempted to maintain backwards compatibility. All map instances are now (immutable) vavr map's.

Is this the (API) design you folks are looking for?

dlsrb6342:
I think `registryTags` should be in the config class. In this way, we can easily add tags in registry with properties file in spring-boot. And we can add unique tags on each instance.
@RobWin @siepkes What do you think?

siepkes:
@dlsrb6342 You mean the ability to specify the tags in the `RetryConfig` class instead of ` AbstractRegistry` and registries directly? That thought crossed my mind too. I implemented it as it currently is because I interpreted the feedback as a specific request to implement it this way. But i'm open to using the `RetryConfig`. It would actually make a lot of sense I think, unless I'm missing something?

RobWin:
We can't add tags to the RetryConfig, because a RetryConfig can be shared by different instances, but still a specific instance should be able to have different tag values. 

If we add tags to the RetryConfig, an instance can't automatically inherit the tags from the global RetryConfig of the AbstractRegistry anymore.

dlsrb6342:
I understand that.
What if not override tags of default config but add instance config's tags to default conifg's tag?

In below method, we can add `InstanceProperties`'s tags to `Builder`'s tags.

https://github.com/resilience4j/resilience4j/blob/8cea8c50df782ea16b7b8bdca2d88d453192a5c1/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/configuration/CircuitBreakerConfigurationProperties.java#L69

RobWin:
Let's ignore the Spring Boot modules for now. We need to provide the basic functionality in the root resilience4j modules. And the root resilience4j modules allow to share configs.

RobWin:
> Let's ignore the Spring Boot modules for now. We need to provide the basic functionality in the root resilience4j modules. And the root resilience4j modules allow to share configs.
I think sharing a config and sharing tags are two different topics.

dlsrb6342:
Oh.. yeah you are right. I understand. Then let's do this as your comments.

siepkes:
@RobWin Is the way I extended the `AbstractRegistry`'s API the way you had in mind? If so I can complete the whole thing in the spirit of this PR.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/InMemoryRetryRegistry.java
```diff
@@ -95,23 +124,38 @@ public Retry retry(String name) {
 	 */
 	@Override
 	public Retry retry(String name, RetryConfig config) {
-		return computeIfAbsent(name, () -> Retry.of(name, Objects.requireNonNull(config, CONFIG_MUST_NOT_BE_NULL)));
+		return retry(name, config, HashMap.empty());
+	}
+
+	@Override
+	public Retry retry(String name, RetryConfig config, io.vavr.collection.Map<String, String> tags) {
+		return computeIfAbsent(name, () -> Retry.of(name, Objects.requireNonNull(config, CONFIG_MUST_NOT_BE_NULL), registryTags.merge(tags)));
```
RobWin: Could you please add a `Objects.requireNonNull(tags, TAGS_MUST_NOT_BE_NULL)` here.
And maybe we should change it to `tags.merge(registryTags)`. 
The [JavaDoc ](https://static.javadoc.io/io.vavr/vavr/0.9.0/io/vavr/collection/Map.html#merge-io.vavr.collection.Map-) says: `Creates a new map which by merging the entries of this map and that map. If collisions occur, the value of this map is taken.`.
I think it's better if the instance-specific tags overwrite the registryTags, if a collision occors..

siepkes: Good catch on the `merge`!

siepkes:
I've added some tests and made some overall API improvements. I've also cemented the way instance tags should override registry tags in both tests and docs.

The API does feel like it could benefit from a "Builder"-treatment to replace all the different constructors which are needed to cope with all situations like: "I want to specify a name and tags but not a retry config". But thats probably something for a different PR.

I assume the following modules need to receive this treatment:

* Retry
* CircuitBreaker
* Fallback
* Ratelimiter
* Bulkhead

Is that a correct assumption?

RobWin:
Yes :) But there is no fallback module.

RobWin:
Do you have time to finalize the PR?

siepkes:
@RobWin Yeah I do! I have currently done half of the other modules. Hoping to finish the PR today.

RobWin:
Awesome. Thanks

siepkes:
@RobWin I'm seeing non-deterministic test failures in multiple test scenarios. For example in `io.github.resilience4j.metrics.TimeLimiterMetricsTest` the test `shouldRecordTimeouts` sometimes fails with:

```
org.junit.ComparisonFailure: expected:<[0]L> but was:<[2]L>
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at io.github.resilience4j.metrics.assertion.CounterAssert.hasValue(CounterAssert.java:21)
	at io.github.resilience4j.metrics.TimeLimiterMetricsTest.shouldRecordTimeouts(TimeLimiterMetricsTest.java:108)
-----8<-------
```

If I run it 4 times in IntellIJ it fails 2 times. When building from the command line it fails most of the times. But there are also other tests that exhibit this behavior.

With vanilla master I seem not to run in to these issues oddly enough. But I fail to see how my modifications can have this effect since I haven't even touched those area's. Let alone that I have made changes that effect timing. Does this seem familiar to you? Something I'm missing?

RobWin:
I will have a look at the tests.

RobWin:
Yes, I will fix the test

RobWin:
Done. I pushed a fix to master

siepkes:
Alright all the grunt work is done. I now need to verify everything and do some cleanup. Also noticed a test failure.

I must admit the whole thing kind of makes my head spin; There are just so many constructors, calling so many other constructors and methods with slightly different arguments. I find it kind of hard to keep an overview.

RobWin:
Yes, we are doing CDP, constructor-driven programming 😜

siepkes:
@RobWin I'm currently still cleaning and checking however I've bumped in a test failure. Not a non-deterministic one this time but quite persistent. However it is about ratpack and I'm not familiar with ratpack. I've looked into it but I don't see anything weird. It is only with the bulkhead so I've double checked my changes in the bulkhead code but I don't see anything weird with it. 

All the normal unit tests (ie. the non ratpack Groovy integration-style tests) also pass.

The test dies with:
```
Condition not satisfied:

bulkheadConfig.present
|              |
Optional.empty false

	at io.github.resilience4j.ratpack.Resilience4jModuleSpec.test shared configs are added to each type registry(Resilience4jModuleSpec.groovy:704)
```

This is with the current state of the PR. Do you have any pointers for me?

RobWin:
Might be the JDK version. Which JDK are you using locally?

RobWin:
Sry, I was  on vacation. I fix the test tomorrow

siepkes:
No sweat, hope you had a good vacation!

RobWin:
Could you please rebase the latest changes in Master

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -182,6 +210,11 @@ public RateLimiterConfig getRateLimiterConfig() {
         return this.rateLimiterConfig.get();
     }
 
+    @Override
+    public Map<String, String> getTags() {
+        return null;
```
RobWin: return tags;

siepkes: Fixed!

siepkes:
@RobWin I've rebased on `master` but I'm still seeing problems with ratpack.

I've also squashed the commits in the PR btw.

RobWin:
Ok, I found the bug.
The bug was in the constructor `public InMemoryBulkheadRegistry(Map<String, BulkheadConfig> configs)`

PR #658, opened 2019-10-02
Kotlin Coroutines flow support

marcoferrer:
This PR introduces support for a circuit breaker operator for cold streams in kotlin coroutines. 

Based on feedback from this PR I can continue implement equivalent operators for the remaining resilience primitives (retry, bulkhead,...) if desired.

Now that the core Flow API has reached a stable [release](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.3.0), it should be safe to implement these operators without worrying about breaking compatibility in future releases.

### DIFF
--- a/resilience4j-kotlin/build.gradle
+++ b/resilience4j-kotlin/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.3.31'
+    id 'org.jetbrains.kotlin.jvm' version '1.3.50'
 }
 
 dependencies {
@@ -19,12 +19,22 @@ dependencies {
     testImplementation(project(':resilience4j-timelimiter'))
 }
 
+def experimentalFlags = [
+    "-Xuse-experimental=kotlin.Experimental"
+]
+
 compileKotlin {
-    kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions{
+        jvmTarget = "1.8"
+        freeCompilerArgs += experimentalFlags
+    }
 }
 
 compileTestKotlin {
-    kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions{
+        jvmTarget = "1.8"
+        freeCompilerArgs += experimentalFlags
+    }
 }
 
 ext.moduleName='io.github.resilience4j.kotlin'
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
@@ -21,6 +21,8 @@ package io.github.resilience4j.kotlin.bulkhead
 import io.github.resilience4j.bulkhead.Bulkhead
 import io.github.resilience4j.bulkhead.BulkheadConfig
 import io.github.resilience4j.kotlin.isCancellation
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import kotlin.coroutines.coroutineContext
 
 /**
@@ -29,19 +31,17 @@ import kotlin.coroutines.coroutineContext
  * If [BulkheadConfig.maxWaitTime] is non-zero, *blocks* until the max wait time is reached or permission is obtained.
  * For this reason, it is not recommended to use this extension function with Bulkheads with non-zero max wait times.
  */
-suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T {
-    acquirePermission()
-    return try {
-        block().also { onComplete() }
-    } catch (e: Throwable){
-        if(isCancellation(e, coroutineContext)){
+suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T =
+    try {
+        acquirePermissionSuspend()
+        block()
+    } finally {
+        if(isCancellation(coroutineContext)){
             releasePermission()
         }else{
             onComplete()
         }
-        throw e
     }
-}
 
 /**
  * Decorates the given suspend function [block] and returns it.
@@ -51,4 +51,20 @@ suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T {
  */
 fun <T> Bulkhead.decorateSuspendFunction(block: suspend () -> T): suspend () -> T = {
     executeSuspendFunction(block)
+}
+
+/**
+ * Try to immediately acquire permission from the bulkhead when it is not expected to block.
+ *
+ * If a wait duration is configured on the bulkhead, then attempt to acquire permission within
+ * the confines of a dispatcher specialized for blocking calls.
+ *
+ */
+internal suspend fun Bulkhead.acquirePermissionSuspend(){
+    // Fast path. Avoid dispatch context switch.
+    if(bulkheadConfig.maxWaitDuration.isZero){
+        acquirePermission()
+    }else{
+        withContext(Dispatchers.IO){ acquirePermission() }
+    }
 }
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
@@ -20,6 +20,8 @@ package io.github.resilience4j.kotlin.bulkhead
 
 import io.github.resilience4j.bulkhead.Bulkhead
 import io.github.resilience4j.bulkhead.BulkheadConfig
+import io.github.resilience4j.kotlin.isCancellation
+import kotlin.coroutines.coroutineContext
 
 /**
  * Decorates and executes the given suspend function [block].
@@ -30,9 +32,14 @@ import io.github.resilience4j.bulkhead.BulkheadConfig
 suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T {
     acquirePermission()
     return try {
-        block()
-    } finally {
-        onComplete()
+        block().also { onComplete() }
+    } catch (e: Throwable){
+        if(isCancellation(e, coroutineContext)){
+            releasePermission()
+        }else{
+            onComplete()
+        }
+        throw e
     }
 }
 
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
@@ -31,18 +31,19 @@ import kotlin.coroutines.coroutineContext
  * If [BulkheadConfig.maxWaitTime] is non-zero, *blocks* until the max wait time is reached or permission is obtained.
  * For this reason, it is not recommended to use this extension function with Bulkheads with non-zero max wait times.
  */
-suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T =
-    try {
-        acquirePermissionSuspend()
-        block()
-    } finally {
-        if(isCancellation(coroutineContext)){
+suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T {
+    acquirePermissionSuspend()
+    return try {
+        block().also { onComplete() }
+    } catch (e: Throwable) {
+        if (isCancellation(coroutineContext,e)) {
             releasePermission()
-        }else{
+        } else {
             onComplete()
         }
+        throw e
     }
-
+}
 /**
  * Decorates the given suspend function [block] and returns it.
  *
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
@@ -35,7 +35,7 @@ suspend fun <T> CircuitBreaker.executeSuspendFunction(block: suspend () -> T): T
         onSuccess(durationInNanos, TimeUnit.NANOSECONDS)
         return result
     } catch (exception: Throwable) {
-        if(isCancellation(exception, coroutineContext)){
+        if(isCancellation(coroutineContext, exception)){
             releasePermission()
         }else{
             val durationInNanos = System.nanoTime() - start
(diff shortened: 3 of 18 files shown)


resilience4j-kotlin/build.gradle
```diff
@@ -19,12 +19,22 @@ dependencies {
     testImplementation(project(':resilience4j-timelimiter'))
 }
 
+def experimentalFlags = [
+    "-Xuse-experimental=kotlin.Experimental"
+]
```
marcoferrer: The experimental flag is only needed for the usage of  https://github.com/resilience4j/resilience4j/blob/05d45eaad00ccd1088e7d30b4e84ca997f233ac5/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreaker.kt#L51

The `UseExperiemental` annotation allows the usage of the `onCompletion()` flow operator.  This lets us guarantee the method contract to library consumers without requiring them to also annotate usages of the circuit breaker operator.

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreaker.kt
```diff
@@ -0,0 +1,96 @@
...
+ *     .collect { println(it) } // Throws CallNotPermittedException
+ * ```
+ *
+ * Coroutine cancellation (_normal_ and _exceptional_) doe not record any events on the circuit breaker
+ * and the acquired permission is released.
+ *
+ */
+@UseExperimental(ExperimentalCoroutinesApi::class)
+fun <T> Flow<T>.circuitBreaker(circuitBreaker: CircuitBreaker): Flow<T> =
+    flow {
```
Dominaezzz: I think you can use [`onStart`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/on-start.html) here.
```
onStart {
    if(circuitBreaker.tryAcquirePermission()) {
        val start = System.nanoTime()
        val source = this@circuitBreaker.onCompletion { e ->
            when {
                isCancellation(e, coroutineContext) -> circuitBreaker.releasePermission()
                e == null -> circuitBreaker.onSuccess(System.nanoTime() - start, TimeUnit.NANOSECONDS)
                else -> circuitBreaker.onError(System.nanoTime() - start, TimeUnit.NANOSECONDS, e)
            }
        }
    } else {
        throw CallNotPermittedException.createCallNotPermittedException(circuitBreaker)
    }
}
```

marcoferrer: The completion handler won’t be subscribed to since its being appended in the onstart. Val source goes used here.

Dominaezzz: This is what I meant.
```
fun <T> Flow<T>.circuitBreaker(circuitBreaker: CircuitBreaker): Flow<T> = onStart {
    if(circuitBreaker.tryAcquirePermission()) {
        val start = System.nanoTime()
        val source = this@circuitBreaker.onCompletion { e ->
            when {
                isCancellation(e, coroutineContext) -> circuitBreaker.releasePermission()
                e == null -> circuitBreaker.onSuccess(System.nanoTime() - start, TimeUnit.NANOSECONDS)
                else -> circuitBreaker.onError(System.nanoTime() - start, TimeUnit.NANOSECONDS, e)
            }
        }
    } else {
        throw CallNotPermittedException.createCallNotPermittedException(circuitBreaker)
    }
}
```

The [implementation](https://github.com/Kotlin/kotlinx.coroutines/blob/7f0da424ed98e3a30e0a7ca2daff33f9a9cdbf0c/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt#L71) is the same, except with some optimisations.

Dominaezzz: Ah I see. Using `onCompletion` in there is a bit awkward to read I think. Perhaps just using try /catch /finally would be clearer?

marcoferrer: The 'onCompletion' operator will return a new instance of flow not the original receiver. Running the tests against this implementation fails. To use 'onStart' we would need to structure the method a little differently so that the completion operator is still chained.

```kotlin
fun <T> Flow<T>.circuitBreaker(circuitBreaker: CircuitBreaker): Flow<T> {
    var start: Long = 0
    return onStart {
        start = System.nanoTime()
        if (!circuitBreaker.tryAcquirePermission()) {
            throw CallNotPermittedException.createCallNotPermittedException(circuitBreaker)
        }
    }.onCompletion { e ->
        when {
            isCancellation(e, coroutineContext) -> circuitBreaker.releasePermission()
            e == null -> circuitBreaker.onSuccess(System.nanoTime() - start, TimeUnit.NANOSECONDS)
            else -> circuitBreaker.onError(System.nanoTime() - start, TimeUnit.NANOSECONDS, e)
        }
    }
}
``` 

But I think this implementation would go against the patterns defined in the stdlib for implementing [opertators](https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-core/common/src/flow/operators). 

Most operators that transform their receiver do so by encapsulating their implementation in a flow builder ie.`flow{}`. With the example above we could potentially see issues with the setting of the start val and reading in the completion handler

Dominaezzz:
```
fun <T> Flow<T>.circuitBreaker(circuitBreaker: CircuitBreaker): Flow<T> = flow {
    if (circuitBreaker.tryAcquirePermission()) {
        val start = System.nanoTime()
        try {
            emitAll(this@circuitBreaker)
            
            circuitBreaker.onSuccess(System.nanoTime() - start, TimeUnit.NANOSECONDS)
        } catch (e: Throwable) {
            if (isCancellation(e, coroutineContext)) {
                circuitBreaker.releasePermission()
                throw e
            } else {
                circuitBreaker.onError(System.nanoTime() - start, TimeUnit.NANOSECONDS, e)
            }
        }
    } else {
        throw CallNotPermittedException.createCallNotPermittedException(circuitBreaker)
    }
}
```

Dominaezzz: After reading the doc for `releasePermission`, I'm not sure it should be called based on cancellation, since you don't really know if cancellation happened before use or not.

marcoferrer: If the flow is already completed then the on complete block wont be invoked again. Meaning that if a cancellation were to occur after the circuit breaker was used, it would not affect completed flows.
 
As for using a try / catch block, the flow documentation explicitly advises against this. Since it violates [Exception Transparency](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/#exception-transparency)
>Flow implementations never catch or handle exceptions that occur in downstream flows. From the implementation standpoint it means that calls to emit and emitAll shall never be wrapped into try { ... } catch { ... } blocks.

Dominaezzz: I guess there's no nice way to write this then.

resilience4j-kotlin/src/test/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreakerTest.kt
```diff
@@ -0,0 +1,202 @@
...
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.toList
+import org.assertj.core.api.Assertions
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import java.lang.IllegalStateException
+import java.util.concurrent.Phaser
+
+class FlowCircuitBreakerTest {
```
RobWin: Could you please add a test which verifies that the producer is not invoked, when a CircuitBreaker is open.
This test should verify that the `CallNotPermittedException` is thrown, before (in RX terms) a subscriber subscribes to the publisher.

RobWin: This test could help you to decide if `onStart` or `flow` is the correct way to implement the circuitbreaker operator.

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreaker.kt
```diff
@@ -0,0 +1,96 @@
...
+        if(circuitBreaker.tryAcquirePermission()) {
```
RobWin: You can just use `circuitBreaker.acquirePermission();` 
The method throws a `CallNotPermittedException`

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreaker.kt
RobWin: > After reading the doc for releasePermission, I'm not sure it should be called based on cancellation, since you don't really know if cancellation happened before use or not.

Your implementation is fine. The method `circuitBreaker.releasePermission()` should be invoked, when the flow is cancelled, but neither `circuitBreaker.onSuccess` nor `circuitBreaker.onError` can be called.

resilience4j-kotlin/src/test/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreakerTest.kt
marcoferrer: This case is covered in the `should not collect when open` test but I can see that its not immediately clear. Ill add another test to make it more explicit that the case is covered

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/FlowCircuitBreaker.kt
marcoferrer: The current implementation will only ever call one of `releasePermission`, `onSuccess`, or `onError`. There are also tests that verify proper permission handling by exercising both normal and exceptional coroutine cancellation.

RobWin: That's correct

marcoferrer: @RobWin Even if we catch and rethrow the exception in the flow body, that will only allow us to handle success / error cases. Using the onCompletion operator ensures that our event reporting is still invoked even in the case of cancellation.

RobWin: Your current implementation looks good to me

RobWin:
Implementation looks fine to me. Would you like to add flow support for RateLimiter, Bulkhead an Retry?

marcoferrer:
Yeah I can submit operator impls for the rest of the resilience primitives. At first I wasn’t sure they if they should all be in one PR, or each have their own. 

One thing I’m realizing now that missing from this PR is updated docs. Is there anything special I need to know about them?

RobWin:
You can add a documentation to a readme in the kotlin module. I will copy it to readme.io

RobWin:
If you prefer, we can make multiple PRs out of it. 
I just want to prevent that we merge a new feature only for one primitive and then the contributor doesn't have time to finalize the other primitives anymore.

marcoferrer:
If thats the case, then I can keep it all in this PR. Just in case there needs to be a patch / hotfix released before I'm finished. 👍

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
```diff
@@ -43,5 +49,5 @@ suspend fun <T> CircuitBreaker.executeSuspendFunction(block: suspend () -> T): T
  * Decorates the given *suspend* function [block] and returns it.
  */
 fun <T> CircuitBreaker.decorateSuspendFunction(block: suspend () -> T): suspend () -> T = {
-    executeSuspendFunction { block() }
+    executeSuspendFunction(block)
```
marcoferrer: Avoid unnecessary lambda instantiation

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
```diff
@@ -28,14 +32,18 @@ import io.github.resilience4j.bulkhead.BulkheadConfig
...
-        onComplete()
+        block().also { onComplete() }
+    } catch (e: Throwable) {
+        if (isCancellation(coroutineContext,e)) {
+            releasePermission()
+        } else {
+            onComplete()
+        }
+        throw e
     }
```
marcoferrer: impl has been updated to properly handle cancellations

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/bulkhead/Bulkhead.kt
```diff
@@ -44,4 +52,20 @@ suspend fun <T> Bulkhead.executeSuspendFunction(block: suspend () -> T): T {
...
+ * the confines of a dispatcher specialized for blocking calls.
+ *
+ */
+internal suspend fun Bulkhead.acquirePermissionSuspend(){
+    // Fast path. Avoid dispatch context switch.
+    if(bulkheadConfig.maxWaitDuration.isZero){
+        acquirePermission()
+    }else{
+        withContext(Dispatchers.IO){ acquirePermission() }
+    }
```
marcoferrer: This is being used to mitigate the risks associated with `acquirePermission` being a blocking API

marcoferrer:
flow operators / tests added for the remaining resilience primitives. Documentation still needs to be updated, but at least the implementations can be discussed in the mean time.

RobWin:
Thx. I've a look at it on Monday

marcoferrer:
No worries. It can wait, I noticed you're on vacation.

RobWin:
PR looks good to me.
Thank you very much.

Issue #590, opened 2019-08-26
Error processing condition on io.github.resilience4j.bulkhead.autoconfigure.BulkheadMetricsAutoConfiguration.registerBulkheadMetrics


dhawal9035:
Trying to use resilience4j with spring boot 1.5.3 and 1.5.22 both. Getting the below error

`START: Application startup failed java.lang.IllegalStateException: Error processing condition on io.github.resilience4j.bulkhead.autoconfigure.BulkheadMetricsAutoConfiguration.registerBulkheadMetrics`

I couldnt find any guides on spring boot 1. 

Dependencies:
```
<dependency>
	<groupId>io.github.resilience4j</groupId>
	<artifactId>resilience4j-spring-boot</artifactId>
	<version>0.17.0</version>
</dependency>
```

RobWin:
Hi,

please have a look at the demo:
 https://github.com/resilience4j/resilience4j-spring-boot-demo/blob/master/build.gradle

Please also have a look at the Spring Boot 2 demo which is more up-to-date.

RobWin:
Could you show me your pom file?

RobWin:
Seems you want to use Dropwizard Metrics instead of Micrometer.
Did you add 

```
<dependency>
	<groupId>io.github.resilience4j</groupId>
	<artifactId>resilience4j-metrics</artifactId>
	<version>0.17.0</version>
</dependency>
```

dhawal9035:
After adding the metrics dependecy, getting the below error. 

```
2019-09-03T10:20:46,844 [local] [ERROR] [restartedMain:SpringApplication:reportFailure:771] [{}] START: Application startup failed org.springframework.context.ApplicationContextException: Unable to start embedded container; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'tomcatEmbeddedServletContainerFactory' defined in class path resource [org/springframework/boot/autoconfigure/web/EmbeddedServletContainerAutoConfiguration$EmbeddedTomcat.class]: Initialization of bean failed; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'serverProperties' defined in class path resource [org/springframework/boot/autoconfigure/web/ServerPropertiesAutoConfiguration.class]: Initialization of bean failed; nested exception is java.lang.NoClassDefFoundError: org/hibernate/validator/internal/engine/DefaultClockProvider
[...]

PR #660, opened 2019-10-03
Remove hibernate-validator dependency

dlsrb6342:
### Issue #590 
Remove hibernate-validator dependency from framework-common module


Romeh:
looks fine for me !

PR #665, opened 2019-10-04
Add exponential backoff functionality to circuit breakers

mlegore:
I added functionality to introduce variable wait times for Circuit breakers in the open state. The wait time can be configured using an IntervalFunction (as is used in resilience-retry). This is functionality we needed within our organization, that I thought might be useful to others.


RobWin:
Thx, could you please create an issue as well and describe your solution?

RobWin:
Do you need an option to reset the attempts counter? 
For example after 10 attempts, reset to the initial wait interval? Or should this be the task of the intervalFunction?

RobWin:
Is the attempts counter set to 0 again, when the CircuitBreaker was closed after it was half open?

RobWin:
Yes, the counter is set to 0. Very good.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -60,6 +61,7 @@
     private boolean writableStackTraceEnabled = DEFAULT_WRITABLE_STACK_TRACE_ENABLED;
```
RobWin: We could remove `waitDurationInOpenState`  and replace it by a constant waitIntervalFunction

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -60,6 +61,7 @@
...
     private Duration waitDurationInOpenState = Duration.ofSeconds(DEFAULT_WAIT_DURATION_IN_OPEN_STATE);
```
RobWin: We could remove `waitDurationInOpenState` and replace it by using a constant default `waitIntervalFunctionInOpenState`

mlegore:
https://github.com/resilience4j/resilience4j/issues/671 created issue

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
mlegore: That would be a breaking change to remove `getWaitDurationInOpenState()` from this class

mlegore: I can do that if you want, up to you

RobWin: No, its fine. I can adjust it.

mlegore:
@RobWin do you have more changes you need? or did I miss something

storozhukBM:
I also did a quick review, and it looks good for me.
Unfortunately, I don't see the results of static code analysis and test coverage.
I am going to kick TravisCI once again.

RobWin:
@mlegore No, I will merge it when I'm back from vacation. I will add my changes after it has been merged.

RobWin:
@mlegore it would be nice if we could add this feature to the Spring Boot modules as well.

Issue #642, opened 2019-09-21
Weight support in RateLimiter


walec51:
Some services define weights of each API call. An example of such is one of the biggest cryptocurrency exchanges: Binance.

Most endpoints have weight = 1. Ones with bigger responses have weight = 5.
Binances API documentation: https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md

There are even endpoints which weight is calculated based on their input parameters.
See `depth` on: https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#order-book

A quick and dirty solution one can currently implement on top of resilience4j is:

```
public interface WeighedRateLimiter {

  static <T> CheckedFunction0<T> decorateCheckedSupplier(
      RateLimiter rateLimiter, int weight, CheckedFunction0<T> supplier) {
    return () -> {
      waitForPermission(rateLimiter);
      IntStream.range(1, weight).forEach(i -> rateLimiter.reservePermission());
      return supplier.apply();
    };
  }
}
```

in this mechanism I actually reserve most of the weight after making the API call. So if a service allows 1200 "points" per minute and I've already user 1170 "points" then I will be still allowed one call with weight 50 for example. This is in some cases acceptable. But there are also services for I which would have to acquire those points before making this call - this is harder to implement and requires changes to resiliance4j it self.

Any suggestions how support for weights could be added to the existing RateLimiter API?

RobWin:
I think we could add a `RateLimiter.acquirePermissions(n)` and `RateLimiter.reservePermissions(n)` to the interface.

RobWin:
Would you like to contribute it?

walec51:
yes I will prepare a PR, just give me some time

RajvanshR:
@RobWin  - I am trying to use Resilience4j with weight. If this is implemented; is there any sample code or documentation I can look at?

PR #672, opened 2019-10-08
Weight support in RateLimiter

walec51:
Implementation of https://github.com/resilience4j/resilience4j/issues/642

### DIFF
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
@@ -323,18 +323,18 @@ public interface RateLimiter {
 	}
 
 	/**
-	 * Acquires a permission of a given weight from this rate limiter, blocking until one is available, or the thread is interrupted.
+	 * Acquires the given amount of permissions from this rate limiter, blocking until one is available, or the thread is interrupted.
 	 * Maximum wait time is {@link RateLimiterConfig#getTimeoutDuration()}
 	 *
 	 * <p>If the current thread is {@linkplain Thread#interrupt interrupted}
 	 * while waiting for a permit then it won't throw {@linkplain InterruptedException},
 	 * but its interrupt status will be set.
 	 * 
-	 * @param weight weight of permission - use for systems where 1 call != 1 permission
+	 * @param permits number of permissions - use for systems where 1 call != 1 permission
 	 * @return {@code true} if a permit was acquired and {@code false}
 	 * if waiting timeoutDuration elapsed before a permit was acquired
 	 */
-	boolean acquirePermission(int weight);
+	boolean acquirePermission(int permits);
 
 	/**
 	 * Reserves a permission from this rate limiter and returns nanoseconds you should wait for it.
@@ -348,14 +348,14 @@ public interface RateLimiter {
 	}
 
 	/**
-	 * Reserves a permission of a given weight from this rate limiter and returns nanoseconds you should wait for it.
+	 * Reserves a number permission from this rate limiter and returns nanoseconds you should wait for it.
 	 * If returned long is negative, it means that you failed to reserve permission,
 	 * possibly your  {@link RateLimiterConfig#getTimeoutDuration()} is less then time to wait for permission.
 	 *
-	 * @param weight weight of permission - use for systems where 1 call != 1 permission
+	 * @param permits number of permission - use for systems where 1 call != 1 permission
 	 * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
 	 */
-	long reservePermission(int weight);
+	long reservePermission(int permits);
 
 	/**
 	 * Get the name of this RateLimiter
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
@@ -318,7 +318,23 @@ public interface RateLimiter {
 	 * @return {@code true} if a permit was acquired and {@code false}
 	 * if waiting timeoutDuration elapsed before a permit was acquired
 	 */
-	boolean acquirePermission();
+	default boolean acquirePermission() {
+		return acquirePermission(1);
+	}
+
+	/**
+	 * Acquires a permission of a given weight from this rate limiter, blocking until one is available, or the thread is interrupted.
+	 * Maximum wait time is {@link RateLimiterConfig#getTimeoutDuration()}
+	 *
+	 * <p>If the current thread is {@linkplain Thread#interrupt interrupted}
+	 * while waiting for a permit then it won't throw {@linkplain InterruptedException},
+	 * but its interrupt status will be set.
+	 * 
+	 * @param weight weight of permission - use for systems where 1 call != 1 permission
+	 * @return {@code true} if a permit was acquired and {@code false}
+	 * if waiting timeoutDuration elapsed before a permit was acquired
+	 */
+	boolean acquirePermission(int weight);
 
 	/**
 	 * Reserves a permission from this rate limiter and returns nanoseconds you should wait for it.
@@ -327,7 +343,19 @@ public interface RateLimiter {
 	 *
 	 * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
 	 */
-	long reservePermission();
+	default long reservePermission() {
+		return reservePermission(1);
+	}
+
+	/**
+	 * Reserves a permission of a given weight from this rate limiter and returns nanoseconds you should wait for it.
+	 * If returned long is negative, it means that you failed to reserve permission,
+	 * possibly your  {@link RateLimiterConfig#getTimeoutDuration()} is less then time to wait for permission.
+	 *
+	 * @param weight weight of permission - use for systems where 1 call != 1 permission
+	 * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
+	 */
+	long reservePermission(int weight);
 
 	/**
 	 * Get the name of this RateLimiter
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
@@ -323,14 +323,14 @@ public interface RateLimiter {
 	}
 
 	/**
-	 * Acquires the given amount of permissions from this rate limiter, blocking until one is available, or the thread is interrupted.
+	 * Acquires the given number of permits from this rate limiter, blocking until one is available, or the thread is interrupted.
 	 * Maximum wait time is {@link RateLimiterConfig#getTimeoutDuration()}
 	 *
 	 * <p>If the current thread is {@linkplain Thread#interrupt interrupted}
 	 * while waiting for a permit then it won't throw {@linkplain InterruptedException},
 	 * but its interrupt status will be set.
 	 * 
-	 * @param permits number of permissions - use for systems where 1 call != 1 permission
+	 * @param permits number of permits - use for systems where 1 call != 1 permit
 	 * @return {@code true} if a permit was acquired and {@code false}
 	 * if waiting timeoutDuration elapsed before a permit was acquired
 	 */
@@ -341,19 +341,19 @@ public interface RateLimiter {
 	 * If returned long is negative, it means that you failed to reserve permission,
 	 * possibly your  {@link RateLimiterConfig#getTimeoutDuration()} is less then time to wait for permission.
 	 *
-	 * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
+	 * @return {@code long} amount of nanoseconds you should wait for reserved permissions. if negative, it means you failed to reserve.
 	 */
 	default long reservePermission() {
 		return reservePermission(1);
 	}
 
 	/**
-	 * Reserves a number permission from this rate limiter and returns nanoseconds you should wait for it.
+	 * Reserves the given number permits from this rate limiter and returns nanoseconds you should wait for it.
 	 * If returned long is negative, it means that you failed to reserve permission,
 	 * possibly your  {@link RateLimiterConfig#getTimeoutDuration()} is less then time to wait for permission.
 	 *
-	 * @param permits number of permission - use for systems where 1 call != 1 permission
-	 * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
+	 * @param permits number of permits - use for systems where 1 call != 1 permit
+	 * @return {@code long} amount of nanoseconds you should wait for reserved permissions. if negative, it means you failed to reserve.
 	 */
 	long reservePermission(int permits);
 
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/AbstractRateLimiterEvent.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/AbstractRateLimiterEvent.java
@@ -23,16 +23,16 @@ import java.time.ZonedDateTime;
 public abstract class AbstractRateLimiterEvent implements RateLimiterEvent {
 
     private final String rateLimiterName;
-    private final int weight;
+    private final int numberOfPermits;
     private final ZonedDateTime creationTime;
 
     public AbstractRateLimiterEvent(String rateLimiterName) {
         this(rateLimiterName, 1);
     }
 
-    public AbstractRateLimiterEvent(String rateLimiterName, int weight) {
+    public AbstractRateLimiterEvent(String rateLimiterName, int numberOfPermits) {
         this.rateLimiterName = rateLimiterName;
-        this.weight = weight;
+        this.numberOfPermits = numberOfPermits;
         creationTime = ZonedDateTime.now();
     }
 
@@ -42,8 +42,8 @@ public abstract class AbstractRateLimiterEvent implements RateLimiterEvent {
     }
 
     @Override
-    public int getWeight() {
-        return weight;
+    public int getNumberOfPermits() {
+        return numberOfPermits;
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -144,13 +145,35 @@ public boolean acquirePermission() {
...
+    /**
+     * Reserving permissions is not supported in the spemaphore based
+     * implementation. Semaphores are totally blocking by it's nature. So this
+     * non-blocking API isn't supported. Use {@link #acquirePermission()}
+     *
+     * @throws UnsupportedOperationException always for this implementation
      */
     @Override
     public long reservePermission() {
-        return -1;
```
walec51: PS. I changed this behaviour to `throw new UnsupportedOperationException` because I think its more developer friendly. No one should call `reservePermissions` in a `SemaphoreBasedRateLimiter`. If some one is doing it he is not aware it is not supported and returning -1 will not explain that to him at first glance.

RobWin: This is a non-backward compatible change. But I like the idea to throw a `UnsupportedOperationException`. What do you think @storozhukBM

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -144,13 +145,35 @@ public boolean acquirePermission() {
...
-     * SemaphoreBasedRateLimiter is totally blocking by it's nature. So this non-blocking API isn't supported.
-     * It will return negative numbers all the time.
+     * Acquiring permissions of a specified weight is not supported in the
+     * spemaphore based implementation. Use {@link #acquirePermission()}
+     *
+     * @throws UnsupportedOperationException always for this implementation
+     */
+    @Override
+    public boolean acquirePermission(int weight) {
+        throw new UnsupportedOperationException("Acquiring permissions of a specified weight is not supported in the spemaphore based implementation");
```
RobWin: We should be able to support it with a Semaphore as well.

RobWin: We can use `tryAcquire(int permits)` and `release(int permits)

walec51: done

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
walec51: I can revert this but I doubt that any properly working code is depending on the current behaviour :)

RobWin: I could live with the change.

walec51:
one test from the `resilience4j-feign` module does not produce repeatable results,
it passed on my local machine but it failed on the CI server

RobWin:
I will fix that.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
storozhukBM: I think this case is already covered in our documentation `amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve`
I'm not against such change in principle, but throwing exception can potentially eliminate some JIT optimizations, so I'd ask @walec51 to perform and post benchmarks and check if there are any regressions.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
```diff
@@ -201,25 +202,26 @@ private State calculateNextState(final long timeoutInNanos, final State activeSt
...
      * @param currentCycle         current {@link AtomicRateLimiter} cycle    @return nanoseconds to wait for the next permission
      */
-    private long nanosToWaitForPermission(final long cyclePeriodInNanos, final int permissionsPerCycle,
+    private long nanosToWaitForPermission(final int weight, final long cyclePeriodInNanos, final int permissionsPerCycle,
                                           final int availablePermissions, final long currentNanos, final long currentCycle) {
-        if (availablePermissions > 0) {
+        if (availablePermissions >= weight) {
             return 0L;
         }
         long nextCycleTimeInNanos = (currentCycle + 1) * cyclePeriodInNanos;
```
storozhukBM: Unfortunately, you can't just update a happy path for this function. With `weight` being introduced to this function, you will need to wait longer, potentially even more than one cycle to accumulate the required amount of permissions.

walec51: I don't think that having weights > limitForPeriod is an actual use case that exists in the wild. One could argue if a `aquirePermission` in this situation should wait multiples of limitRefreshPeriod or should it always timeout in this situation. Having a `aquirePermission` that block even on a fresh cycle seams strange to me.

I would suggest to address this issue either by:
a) throwing an exception the passed weight is weights > limitForPeriod
b) adding some documentation that which ever weight you pass we will not wait more then one limitRefreshPeriod

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
walec51: After reading the documentation I think most people will assume that `failed to reserve` means that it failed because the calculated wait time was more then the timeout. Calling an unsupported function is a different calibre of failure to me. The first type of failure is expected to happen in normal production code, the second is not.

As for testing performance you mean that I should test if the performance of calling `AtomicRateLimiter#reservePermission` if affected by this change?

Because as for `SemaphoreBasedRateLimiter#reservePermission` I do not think performance is any issue because it should never be called in production code and the whole point of throwing the suggested exception is to prevent that as early in development as possible.

RobWin: > After reading the documentation I think most people will assume that failed to reserve means that it failed because the calculated wait time was more then the timeout. Calling an unsupported function is a different calibre of failure to me. The first type of failure is expected to happen in normal production code, the second is not.

I agree

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
walec51: so, in which direction do we want to take this?
should we have an `aquirePermission` that can wait even on a fresh cycle or should we do (a) or (b) from my proposals?
personally I'm in favour of (a)

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
walec51: I tried running the JMH benchmarks but their Gradle configuration doesn't work on my machine.

```
cd resilience4j-ratelimiter
../gradlew --configure-on-demand -w -x check jmh
Configuration on demand is an incubating feature.
Processing 1 classes from /mnt/reszta/Projekty/WALCZAK IT/scr/resilience4j/resilience4j-ratelimiter/build/classes/java/jmh with "reflection" generator
Writing out Java source to /mnt/reszta/Projekty/WALCZAK IT/scr/resilience4j/resilience4j-ratelimiter/build/jmh-generated-sources and resources to /mnt/reszta/Projekty/WALCZAK IT/scr/resilience4j/resilience4j-ratelimiter/build/jmh-generated-resources
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.openjdk.jmh.util.Utils (file:/home/walec51/.gradle/caches/modules-2/files-2.1/org.openjdk.jmh/jmh-core/1.21/442447101f63074c61063858033fbfde8a076873/jmh-core-1.21.jar) to field java.io.PrintStream.charOut
WARNING: Please consider reporting this to the maintainers of org.openjdk.jmh.util.Utils
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
# JMH version: 1.21
# VM version: JDK 11.0.4, OpenJDK 64-Bit Server VM, 11.0.4+11
# VM invoker: /usr/lib/jvm/java-11-openjdk-11.0.4.11-0.fc30.x86_64/bin/java
# VM options: <none>
# Warmup: 10 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 2 threads, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: io.github.resilience4j.ratelimiter.RateLimiterBenchmark.atomicPermission

# Run progress: 0,00% complete, ETA 00:40:00
# Fork: 1 of 2
Error: Could not find or load main class org.openjdk.jmh.runner.ForkedMain
Caused by: java.lang.ClassNotFoundException: org.openjdk.jmh.runner.ForkedMain
<forked VM failed with exit code 1>
<stdout last='20 lines'>
</stdout>
<stderr last='20 lines'>
Error: Could not find or load main class org.openjdk.jmh.runner.ForkedMain
Caused by: java.lang.ClassNotFoundException: org.openjdk.jmh.runner.ForkedMain
</stderr>
```

any suggestions what could be the cause of this?

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
storozhukBM: I believe that we should not change the contract and implement current semantic "if an `aquirePermission` in this situation should wait for a multiple of `limitRefreshPeriod's`". The client already can pass the expected timeout, so if `nanosToWaitForPermission` returns too big duration, we won't reserve these permissions.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
RobWin: I have a look at it. We did a major gradle update a while ago.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
walec51: so...
even if the rate limiter did do nothing for the past dozens of cycles,
if I call `aquirePermission(6)` and my `limitForPeriod = 2`,
then you want me to wait for 3 cycles?

this behaviour is absurd but if you want it I'll implement it
from my perspective its an error to have weight > limitForPeriod and I would never allow it in my application code nor does any API service known to me impose such bizarre rules - they would be even impossible to enforce from the APIs perspective because you'll wait even before making your first call to a function with a weight > limitForPeriod ...

storozhukBM: Hi @walec51

First of all, I want to say that I'm very happy and thankful that you decided to contribute such a change to our library. I think that this is a good idea and it will be valuable for some of our users.
But I want to kindly ask you to try to avoid any subjective thoughts and messages. Things like "so..." and "this behavior is absurd" can be considered rude by other contributors.

As for your point about, "its an error to have `weight` > `limitForPeriod` and I would never allow it in my application code". I totally get it, especially if we consider only "blocking" APIs.
But current AtomicRateLimiter is essentially an implementation of a super simple "fixed window rate limiting" algorithm with non-blocking API. And in our implementation, it is OK for `State.activePermissions` to be a negative `int`, where it means that there is a thread or multiple threads that reserved some permissions for future calls and are waiting for their turn or doing some other work while waiting. 
So when we take this into consideration, there can be a situation where
`weight` is less than `limitForPeriod`, but you'll need to wait for more than one cycle for your turn to perform a request.

Hopefully, this makes sense for you, and please correct me if I'm wrong.

walec51: ok, sorry for my attitude, please take a look at the current implementation

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
```diff
@@ -201,50 +202,56 @@ private State calculateNextState(final long timeoutInNanos, final State activeSt
...
         long nanosToNextCycle = nextCycleTimeInNanos - currentNanos;
-        int fullCyclesToWait = (-availablePermissions) / permissionsPerCycle;
-        return (fullCyclesToWait * cyclePeriodInNanos) + nanosToNextCycle;
+        int permissionsToTakeAfterCurrencyCycle = weight - availablePermissions;
+        int fullCyclesToWaitCount = Math.floorDiv(permissionsToTakeAfterCurrencyCycle, permissionsPerCycle);
+        if (permissionsToTakeAfterCurrencyCycle % permissionsPerCycle == 0) {
+            // will take all permissions from the last cycle but can reserve them without waiting till its end
+            fullCyclesToWaitCount -= 1;
+        }
+        return nanosToNextCycle + (fullCyclesToWaitCount * cyclePeriodInNanos);
```
storozhukBM: I'd suggest such simplification
```java
int permissionsAtTheStartOfNextCycle = availablePermissions + permissionsPerCycle;
int fullCyclesToWait = -(permissionsAtTheStartOfNextCycle-weight) / permissionsPerCycle;
return (fullCyclesToWait * cyclePeriodInNanos) + nanosToNextCycle;
```
I'm not totally sure about all corner cases, but all tests appear to work OK.
If I missed something, please correct me.

walec51: ok, done

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
```diff
@@ -318,7 +318,23 @@ static void waitForPermission(final RateLimiter rateLimiter) {
...
+	 *
+	 * <p>If the current thread is {@linkplain Thread#interrupt interrupted}
+	 * while waiting for a permit then it won't throw {@linkplain InterruptedException},
+	 * but its interrupt status will be set.
+	 * 
+	 * @param weight weight of permission - use for systems where 1 call != 1 permission
+	 * @return {@code true} if a permit was acquired and {@code false}
+	 * if waiting timeoutDuration elapsed before a permit was acquired
+	 */
+	boolean acquirePermission(int weight);
```
RobWin: Can we rename `weight` to `permits`. Analogous to `https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html#acquire(int)`?

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
```diff
@@ -327,7 +343,19 @@ static void waitForPermission(final RateLimiter rateLimiter) {
...
+
+	/**
+	 * Reserves a permission of a given weight from this rate limiter and returns nanoseconds you should wait for it.
+	 * If returned long is negative, it means that you failed to reserve permission,
+	 * possibly your  {@link RateLimiterConfig#getTimeoutDuration()} is less then time to wait for permission.
+	 *
+	 * @param weight weight of permission - use for systems where 1 call != 1 permission
+	 * @return {@code long} amount of nanoseconds you should wait for reserved permission. if negative, it means you failed to reserve.
+	 */
+	long reservePermission(int weight);
```
RobWin: Can we rename `weight` to `permits`. Analogous to `https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html#acquire(int)`?

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/AbstractRateLimiterEvent.java
```diff
@@ -35,6 +41,11 @@ public String getRateLimiterName() {
         return rateLimiterName;
     }
 
+    @Override
+    public int getWeight() {
```
RobWin: Can we rename it to `getNumberOfPermits`?

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/AbstractRateLimiterEvent.java
```diff
@@ -23,10 +23,16 @@
...
 
     private final String rateLimiterName;
+    private final int weight;
     private final ZonedDateTime creationTime;
 
     public AbstractRateLimiterEvent(String rateLimiterName) {
+        this(rateLimiterName, 1);
+    }
+
+    public AbstractRateLimiterEvent(String rateLimiterName, int weight) {
```
RobWin: Can we rename `weight` to `permits`.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/RateLimiterEvent.java
```diff
@@ -28,6 +28,8 @@
 
     Type getEventType();
 
+    int getWeight();
```
RobWin: Can we rename it to getNumberOfPermits?

RobWin:
PR looks good to me.
@storozhukBM Your final review?

storozhukBM:
Hi @RobWin and @walec51 
Looks good for me too. So, I'll proceed with the merge. 
@walec51 thank you once again for this contribution and your patience, good job.

Issue #677, opened 2019-10-09
BDD mockito expectation and comments are at variance in SupplierRetryTest.java 


mohanr:
I looked at `resilience4j/resilience4j-retry/src/test/java/io/github/resilience4j/retry/internal/SupplierRetryTest.java`. 
The comments in the test code to 'retry' mention it should be retried a certain number of times. But the BDD mockito is expecting the methods to be called a different number of times.
But I may have misunderstood the code.

Romeh:
@mohanr can you please share which test case in that unit test class you have confusion about ?

Romeh:
@mohanr 
Please ignore the comments as it is maybe not updated properly during the copy past for the reported the test cases , just follow the unit test code which explain the flow .
we will update the comments and test cases name to reflect properly the actual test logic

hexmind:
**testDecorateSupplier**
Look at 
```
given(helloWorldService.returnHelloWorld())
  .willThrow(new HelloWorldException())
  .willReturn("Hello world");
```
Then `should(times(2)` because 2nd call is successful (no more retry needed).


**testDecorateSupplierAndInvokeTwice**
 Look at 
```
given(helloWorldService.returnHelloWorld())
  .willThrow(new HelloWorldException())
  .willReturn("Hello world")
  .willThrow(new HelloWorldException())
  .willReturn("Hello world");
```
and 
```
String result = supplier.get();
String result2 = supplier.get();
```
Then `should(times(4)` because result retry once (2 calls)
and result2 also retry once (next 2 calls).

PR #680, opened 2019-10-12
BDDMockito clean up (#677)

hexmind:
* add static imports (it is DSL for tests - readability is important)
* replace verify with then
* remove redundant (sometimes outdated) comments that repeat what is expressed line below (only spaces  are replaces by dots :)


RobWin:
Could you please rebase the latest changes in Master

Issue #657, opened 2019-10-02
Future not supported in BulkheadAspect


pulkitmehra:
Hi, resiliency4j team, I was looking into `BulkheadAspect` code in `resiliency4j-spring` module. I see for `ThreadPool` mode we only support `CompletionStage`. Is it possible to support plain java `Futures` as well?

See code [here](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java#L218-L229)
```
 if (CompletionStage.class.isAssignableFrom(returnType)) {
            ....
        } else {
            throw new IllegalStateException("ThreadPool bulkhead is only applicable for completable futures ");
        }
```

Can you suggest an alternative route if we want to support java Future. Thanks in advance!

RobWin:
Hi,
would you like to propose a solution/PR?

Romeh:
@pulkitmehra  can you please explain more why you still need to use the plain java future and can not use the completable future ?

pulkitmehra:
@Romeh , @RobWin 

The following are the reason for supporting this.

1)For application who wants to use declarative style(Annotation-based) with Future return type Resiliency4j features, they do have any option other than upgrader as Aspects simply throw an exception for an incompatible type like here

https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java#L227

2) CompletableFuture extends Future and CompletionStage. There is no connection between Future and CompStage, so returning CompetionStage supports only CompletableFuture and fail for Future as it's in a separate inheritance hierarchy.

3)There are lots of legacy applications that are on Java8 but do not support the CompletableFuture callback model. This forces them to upgrade legacy code which could impact the adaptability of the Resiliency4j framework. This is a minimum migration path for Java8 + Future + Declarative style implementation.

5) We are trying to migrate from Hystix to Resiliency4j but as Hysrix was not supporting CompletableFuture, we have 5k+ Java+Spring based APIs that are using the Future types. We want to give them simpler migration path by supporting Future + Declarative style annotation.

RobWin:
I think it's only possible to decorate a Future. A Bulkhead, CircuitBreaker or Retry decorator must know if an exceptions happened or when a call succeeded. That means the decorator must actually invoke `future.get` to know the result. But then the decorator can't return a decorated Future, but only a result or the exception.

[...]

PR #689, opened 2019-10-15
WIP: Support java Future in Resiliency4j Aspects.

pulkitmehra:
PR issue discussion https://github.com/resilience4j/resilience4j/issues/657

Background:

1. For application who wants to use declarative style(Annotation-based) with Future return type Resiliency4j features, they do have any option other than upgrader as Aspects simply throw an exception for an incompatible type like here
```
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java#L227
```

2. CompletableFuture extends Future and CompletionStage. There is no connection between Future and CompStage, so returning CompetionStage supports only CompletableFuture and fail for Future as it's in a separate inheritance hierarchy.

3. There are lots of legacy applications that are on Java8 but do not support the CompletableFuture callback model. This forces them to upgrade legacy code which could impact the adoptability of the Resiliency4j framework. This is a minimum migration path for Java8 + Future + Declarative style implementation.

4. We are trying to migrate from Hystix to Resiliency4j but as Hysrix was not supporting CompletableFuture, we have 5k+ Java+Spring based APIs that are using the Future types. We want to give them simpler migration path by supporting Future + Declarative style annotation.


Romeh:
Hi @pulkitmehra 

Many thanks for your PR and your effort  :) , please find out our first remarks round :

1. We do not understand why you wrap future in a completion stage in your future API introduction? what is the benefit of that ?!  loop till the future it is done and then wrapping it is own result in Completion stage is just not right :) .

2.  Decorating Future for resileience4j supported operations (ex Retry..ect) require invoking the future itself to be able to detect if it is successfully invoked or not to call the related onSuccess , onError hooks that actually execute the target required handling of retry or circuitbreaker...ect 

3. the ways you handle aspect invocation is not right , for example :

```
private Object handleJoinPointFuture(ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.circuitbreaker.CircuitBreaker circuitBreaker) {
        return circuitBreaker.executeSupplier(() -> {
            try {
                return (Future<?>) proceedingJoinPoint.proceed();
                throw new CompletionException(throwable);
            }
        });
    }

```
Code invokes a supplier method which returns a Future.
But when the Future itself is done and throws an exception, the exception is not caught by the CircuitBreaker and recorded to invoked the right action on it accordingly

pulkitmehra:
First of all, thank you for your comments and please pardon my naive implementation, you can take this as starting point where we can have a further discussion on the approach and design. I am committed to incorporate any feedback provided and would enjoy learning along the way! 

> We do not understand why you wrap future in a completion stage in your future API introduction? what is the benefit of that ?! loop till the future it is done and then wrapping it is own result in Completion stage is just not right :) .

The background is that `CompletionStage` and `Future` interfaces lies in different inheritance hierarchy and CompletableFuture implements both interfaces. Following were my motivation to decorate `Future` with `CS`

1. I want to create an adapter which convert the `Future` to `CompletableFuture` so that we can reuse the existing code and stick to Functional paradigm of Java8 otherwise slowly there would be while's and timed wait's leaked into all part of the codebase for handling Futures. I could be wrong in this assumption and open to take a different route.

2. I wanted to reuse the [AsycnRetryContext](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java#L584-L640) for retry logic which only works with `CompletionStage`. We can create a new class `AsycnFutureRetryContext` which exclusively handle `Future` type.

Above are top two motivation to wrap `Future` into `CompletionStage`. I would appreciate if you guide me with an approach which you think would be best for the framework. 

> Decorating Future for resileience4j supported operations (ex Retry..ect) require invoking the future itself to be able to detect if it is successfully invoked or not to call the related onSuccess , onError hooks that actually execute the target required handling of retry or circuitbreaker...ect

> he ways you handle aspect invocation is not right , for example : .... Code invokes a supplier method which returns a Future.
But when the Future itself is done and throws an exception, the exception is not caught by the CircuitBreaker and recorded to invoked the right action on it accordingly

I think the above 2 comments from you are pointing to the same problem. I think I understand what I missed here, but to re-confirm I should do something similar like below for `Future` type implementation. I can incorporate that change.

https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java#L631-L666

Overall, I would appreciate if you share some implementation notes/design that you would like me to incorporate in the PR.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -620,6 +629,34 @@ public String toString(){
...
+     * Returns a supplier of type Future which is decorated by a CircuitBreaker.
+     *
+     * @param circuitBreaker the CircuitBreaker
+     * @param supplier       the original supplier
+     * @param <T>            the type of the returned Future's result
+     * @return a supplier which is decorated by a CircuitBreaker.
+     */
+    static <T> Supplier<Future<T>> decorateFuture(CircuitBreaker circuitBreaker, Supplier<Future<T>> supplier) {
+       return () -> {
+           if (!circuitBreaker.tryAcquirePermission()) {
```
RobWin: I think when I invoke a method which returns a Future (and which does not throw any checked exceptions), I wouldn't expect to get a RuntimeException like `CallNotPermittedException`. I would expect to get the exception inside of an `ExecutionException` when I call `future.get`.  We should direclty return a Future completed exceptionallym when the acquire permission was rejected.

```
final CompletableFuture<T> promise = new CompletableFuture<>();
promise.completeExceptionally(e);
return promise;
```

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -983,7 +1020,102 @@ static CircuitBreaker of(String name, Supplier<CircuitBreakerConfig> circuitBrea
...
+        }
+
+        @Override
+        public boolean isDone() {
+            return future.isDone();
+        }
+
+        @Override
+        public T get() throws InterruptedException, ExecutionException {
+            if (!isPermissionAcquired && !circuitBreaker.tryAcquirePermission()) {
```
RobWin: Why do we need to check the permission again? 
The permission was already checked even before the method which returns a Future was invoked.

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
```diff
@@ -152,6 +155,19 @@ private Object handleJoinPointCompletableFuture(ProceedingJoinPoint proceedingJo
 		});
 	}
 
+	/**
+	 * handle the Future return types AOP based into configured circuit-breaker
+	 */
+	private Object handleJoinPointFuture(ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.circuitbreaker.CircuitBreaker circuitBreaker) {
+		return circuitBreaker.executeSupplier(() -> {
```
clgroft: Shouldn't `CircuitBreaker` have an `executeFuture()` method to call here?

While I'm here anyway, why does this code catch `Throwable` rather than `Exception`? I know this just copies behavior from nearby, but it seems troublesome.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
RobWin: The decorator itself should not spawn a thread. The method which returns the future controls the thread management. Our decorator should just do the permission check and call onSuccess or onError.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
```diff
@@ -190,7 +191,7 @@ public void acquirePermission() {
     public void onError(long duration, TimeUnit durationUnit, Throwable throwable) {
         // Handle the case if the completable future throws a CompletionException wrapping the original exception
         // where original exception is the the one to retry not the CompletionException.
-        if (throwable instanceof CompletionException) {
+        if (throwable instanceof CompletionException || throwable instanceof ExecutionException) {
```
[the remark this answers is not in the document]
RobWin: True, I didn't see that

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
clgroft: When precisely should the decorator call `onSuccess()` or `onError()`? The first time a success or failure is observed? That could be a much longer interval than the time that the underlying call took to return. Short of pushing the call onto the thread that's already supplying the value of the Future (and how would one do that?), I'm not sure what else can be done here.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
pulkitmehra: Following are thoughts behind the solution. I am flexible here.. 

1. There are certain gotchas with using `CompletableFuture` with `Future` type.  Wrapping exception inside `promise.completeExceptionally(e);` will wrap everything inside `ExecutionException` and it will break `Future` contract for example

```
final CompletableFuture<T> promise = new CompletableFuture<>();
try{
  T v = future.get(5, SECONDS);
  promise.complete(v);
} catch(ExecutionException e){
  promise.completeExceptionally(e.getCause()); //Works
} catch (CancellationException e){
  promise.cancel(true); //Works
} catch (TimeoutException | InterruptedException e){
  promise.completeExceptionally(e); //Breaks Contract as TimeoutException & InterruptedException wraps inside ExecutionException
}
return promise;
```

Also, the downside of this approach is that we have to invoke `get()` to find results where as with `CircuitBreakerFuture` we just decorate future and let caller invoke it.

2. Regarding `CallNotPermittedException` exception there are two scenario here, Supplier fails before `Future` is even returned and Second, invoking `Future` fails. 

::Example: Supplier fails before `Future` is even returned
```
Future<String> work = () -> {
    Supplier<String> t = svc.execute() //this blows up!
    return executorService.submit(t);
}
Future<String> decorateFuture = circuitBreaker.decorateFuture(work);
```
The above scenario would not be count towards CB if we do not handle it outside `Future`

::Example: Invoking `Future` fails: this would handle in `CircuitBreakerFuture::get` method.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
pulkitmehra: After above discussion, I am inching towards adding a `permissionEvaluated` check in `CircuitBreakerFuture` class which bypass the CircuitBreaking logic if evaluated already. Please confirm?

@Clgroft makes a good point here, evaluating `future.get` is at discretion of caller and it could be a much longer interval than the time that the underlying call took to return. Where as in `CompletableFuture` `whenComplete` is called as soon as underlying computation is return. This is downside of creating your own `Future` and decorating it with CircuitBreaking logic. Let us know your thoughts.

pulkitmehra: @RobWin ,This code has `isPermissionAcquired` flag which tells whether permission was acquired before or not. This is for completeness, trying to cover scenario where permission is not acquired before. However, we don't have a use case in the implementation yet.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -620,6 +629,34 @@ public String toString(){
...
+    static <T> Supplier<Future<T>> decorateFuture(CircuitBreaker circuitBreaker, Supplier<Future<T>> supplier) {
+       return () -> {
+           if (!circuitBreaker.tryAcquirePermission()) {
+               throw CallNotPermittedException.createCallNotPermittedException(circuitBreaker);
+           } else {
+               final long start = System.nanoTime();
+               Future<T> val = null;
+               try {
+                   val = supplier.get();
+               } catch (Exception e) {
```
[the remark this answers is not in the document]
pulkitmehra: compiler doing its magic here 👍

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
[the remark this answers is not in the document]
pulkitmehra: It's just an extra safety check. I will remove this piece and pass `e.getCause()` from the caller

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -983,7 +1020,102 @@ static CircuitBreaker of(String name, Supplier<CircuitBreakerConfig> circuitBrea
...
+                    long durationInNanos = System.nanoTime() - start;
+                    circuitBreaker.onSuccess(durationInNanos, TimeUnit.NANOSECONDS);
+                    return v;
+                } catch (ExecutionException e) {
+                    long durationInNanos = System.nanoTime() - start;
+                    circuitBreaker.onError(durationInNanos, TimeUnit.NANOSECONDS, e.getCause());
+                    throw e;
+                } catch (CancellationException e) {
+                    circuitBreaker.releasePermission();
+                    throw e;
```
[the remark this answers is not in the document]
pulkitmehra: @RobWin , should we `InterruptedExceptions` as `CancellationException` by releasing the permission and not counting these calls towards CircuitBreaking as the current thread is cancelled?

resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerAspect.java
pulkitmehra: I will address other parts of code later after we understand how to correctly decorate Futures.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
RobWin: What about this
```
    static <T> Supplier<Future<T>> decorateFuture(CircuitBreaker circuitBreaker, Supplier<Future<T>> supplier) {
        return () -> {
            if (!circuitBreaker.tryAcquirePermission()) {
                CompletableFuture<T> promise = new CompletableFuture<>();
                promise.completeExceptionally(CallNotPermittedException.createCallNotPermittedException(circuitBreaker));
                return promise;
            } else {
                final long start = System.nanoTime();
                try {
                    return new CircuitBreakerFuture<>(circuitBreaker, supplier.get(), start);
                } catch (Exception e) {
                    circuitBreaker.onError(durationInNanos, TimeUnit.NANOSECONDS, e);
                    throw e;
                }
            }
        };
    }

    final class CircuitBreakerFuture<T> implements Future<T> {
        private CircuitBreaker circuitBreaker;
        private Future<T> future;
        private long start;
        private AtomicBoolean processed = new AtomicBoolean(false);

        CircuitBreakerFuture(CircuitBreaker circuitBreaker, Future<T> future, long start) {
            Objects.requireNonNull(future, "Non null Future is required to decorate");
            this.circuitBreaker = circuitBreaker;
            this.future = future;
            this.start = start;
        }

        @Override
        public boolean cancel(boolean mayInterruptIfRunning) {
            return future.cancel(mayInterruptIfRunning);
        }

        @Override
        public boolean isCancelled() {
            return future.isCancelled();
        }

        @Override
        public boolean isDone() {
            return future.isDone();
        }

        @Override
        public T get() throws InterruptedException, ExecutionException {
            try {
                T v = future.get();
                if(processed.compareAndSet(false, true)){
                    circuitBreaker.onSuccess(durationInNanos, TimeUnit.NANOSECONDS);
                }
                return v;
            } catch (CancellationException | InterruptedException e) {
                circuitBreaker.releasePermission();
                throw e;
            } catch (Exception e) {
                if(processed.compareAndSet(false, true)){
                    circuitBreaker.onError(durationInNanos, TimeUnit.NANOSECONDS, e.getCause());
                }
                throw e;
            }
        }

        @Override
        public T get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
            try {
                T v = future.get(timeout, unit);
                circuitBreaker.onSuccess(durationInNanos, TimeUnit.NANOSECONDS);
                return v;
            } catch (CancellationException | InterruptedException e) {
                circuitBreaker.releasePermission();
                throw e;
            } catch (Exception e) {
                if(processed.compareAndSet(false, true)){
                    circuitBreaker.onError(durationInNanos, TimeUnit.NANOSECONDS, e.getCause());
                }
                throw e;
            }
        }
    }
```

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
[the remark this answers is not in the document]
RobWin: Yes

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
RobWin: Since there is not hook like `whenComplete`, we stop the time when `future.get` is invoked. I think that's fine. We can document it in the JavaDoc.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
clgroft: Given that we're OK with stopping the clock at the first `get` if necessary, I like this solution. Would suggest factoring the "only call to the CircuitBreaker once" and "pass in execution time" into their own utility classes, e.g.:

```java
final class OnceSupplier<T> {
    private final T val;
    private final AtomicBoolean hasRun = new AtomicBoolean(false);
    public OnceSupplier(T val) { this.val = val; }
    public void supplyOnce(Consumer<T> consumer) {
        if (hasRun.compareAndSet(false, true)) {
            consumer.accept(val);
        }
    }
}

final class NanosecondTimer {
    private final long startNanos;
    public NanosecondTimer(long startNanos) { this.startNanos = startNanos; }
    public void runWithTime(LongConsumer consumer) {
        consumer.accept(System.nanoTime() - startNanos);
    }
}

final class CircuitBreakerFuture<T> implements Future<T> {
    private Future<T> future;
    private OnceSupplier<CircuitBreaker> onceToCircuitBreaker;
    private NanosecondTimer durationTimer;

    public CircuitBreakerFuture(CircuitBreaker circuitBreaker, Future<T> future, long start) {
        this.future = future;
        this.onceToCircuitBreaker = new OnceSupplier<>(circuitBreaker);
        this.durationTimer = new NanosecondTimer(start);
    }

    ...

    @Override
    public T get() throws InterruptedException, ExecutionException {
        try {
            T v = future.get();
            onceToCircuitBreaker.supplyOnce(circuitBreaker ->
                durationTimer.runWithTime(durationInNanos ->
                    circuitBreaker.onSuccess(durationInNanos, TimeUnit.NANOSECONDS)));
        } catch (CancellationException | InterruptedException e) {
            onceToCircuitBreaker.supplyOnce(CircuitBreaker::releasePermission);
            throw e;
        } catch (Exception e) {
            onceToCircuitBreaker.supplyOnce(circuitBreaker ->
                durationTimer.runWithTime(durationInNanos ->
                    circuitBreaker.onError(durationInNanos, TimeUnit.NANOSECONDS, e)));
            throw e;
        }
    }

    ...
}
```

pulkitmehra:
@RobWin @Romeh , I am planning to close this PR and likely to issue small PRs for each Interface. Here is the PR to support Future in CircuitBreaker https://github.com/resilience4j/resilience4j/pull/714/files

The following PR will likely to follow after this.
1. Retry support Future
2. RateLimiter support Future
3. Bulkhead support Future
4. Support Future in Spring Aspects

cc: @clgroft

Issue #691, opened 2019-10-16
Exceptions thrown from fallback methods shouldn't be wrapped


olovandersson:
Currently any exception thrown from a fallback method is wrapped in an InvocationTargetException by the FallbackMethod class. Fallback methods should be able to throw regular exceptions that calling code should be able to catch.

RobWin:
Thanks

PR #692, opened 2019-10-16
Unwrap fallback exception

olovandersson:
Fix for issue #691


RobWin:
Could you also please verify how it behaves if it is a checked Exception?

RobWin:
Don't worry about the failing test. I fix the test in the master branch. The bug was introduces by a different PR.

PR #698, opened 2019-10-17
Issue 687 retry on result on success() proper handling

Romeh:
The implementation will be like the following :

Rather than just throwing Max retries exceeded exception when on result check fail > max retries which can beak existing semantic for people who use Retry API for now plus Rx and Reactor operators  , so onSuccess will be tuned : 

1- if result check succeeded < max retries , onSuccess will do the current logic which include increase the related counters and submitting RetryOnSuccessEvent  
2- if result check > max tries , onSuccess will  generate RetryOnErrorEvent with last reported exception otherwise if none it will be MaxRetriesExceeded exception plus increase failedAfterRetryCounter

the current Retry APIs will keep the current behavior which return the last calculated result


RobWin:
@KAbakumovAtl What would you expect? An exception or the result ?

KAbakumovAtl:
If my operation was failed to be retried I'd expect the overall procedure fail in the same way as operation itself. If operation failed because of exception I'd expect the same exception to be thrown. If operation failed because result did not match isResultSuccessful predicate I'd expect to get the result as is. (no exception in that case)

From the events point of view I'd expect to get some event which will indicate that we have an error and that maximum amount of retries reached. OnError events seems to fit here.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -139,12 +139,18 @@ private ContextImpl() {
...
 			int currentNumOfAttempts = numOfAttempts.get();
-			if (currentNumOfAttempts > 0) {
+			if (currentNumOfAttempts > 0 && currentNumOfAttempts < maxAttempts) {
 				succeededAfterRetryCounter.increment();
 				Throwable throwable = Option.of(lastException.get()).getOrElse(lastRuntimeException.get());
 				publishRetryEvent(() -> new RetryOnSuccessEvent(getName(), currentNumOfAttempts, throwable));
 			} else {
-				succeededWithoutRetryCounter.increment();
+				if (currentNumOfAttempts >= maxAttempts) {
+					failedAfterRetryCounter.increment();
```
KAbakumovAtl: Looks a bit weird that there is a branch inside onSuccess method which emits error event. It would be more consistent to rename onSuccess to onOperationFinished or onRetriesStopped etc.

RobWin: Yes, we are also discussing it internally.
Unfortunately we don't want to introduce a non-backward compatible change. The method `onSuccess` is part of the public API.

RobWin: The onResult handling was added at a later point in time. That's why it doesn't really fit into the public API.

Romeh: So what is the final conclusion here :) ?

RobWin: We could also just create a new `onResult` and move the logic, which is currently in the decorators, into that method.
The code is actually duplicated everywhere:

```
T result = supplier.get();
				final boolean validationOfResult = context.onResult(result);
				if (!validationOfResult) {
					context.onSuccess();
					return result;
				}
```

Romeh:
> We could also just create a new onResult and move the logic, which is currently in the decorators, into that method.
> The code is actually duplicated everywhere:
> 
> ```
> T result = supplier.get();
> final boolean validationOfResult = context.onResult(result);
> if (!validationOfResult) {
>     context.onSuccess();
>     return result;
> }
> ```
> 
> `onSuccess` could be deprecated and only used internally.
@RobWin 
you mean adding new API(onResult) on the Retry interface level not on Retry context level ?

RobWin:
No, you are right.
I meant `Retry.Context` and `Retry.AsyncContext`. We could deprecate `onSuccess` and move the common logic into an improved `onResult`.

RobWin:
Could you please check if it also works for our RxJava2 and Reactor operator?

resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -510,9 +510,16 @@ default void executeRunnable(Runnable runnable) {
...
 		 * Records a successful call.
+		 * @deprecated since 1.2.0
 		 */
+		@Deprecated
 		void onSuccess();
 
+		/**
+		 * Records a successful call or retryable call with the needed generated retry events.
+		 */
+		void onFinish();
```
RobWin: One last request. Can we rename it to `onComplete`? :)

Romeh: Sure i suck in naming :)

resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -510,9 +510,16 @@ default void executeRunnable(Runnable runnable) {
 
 		/**
 		 * Records a successful call.
+		 * @deprecated since 1.2.0
 		 */
+		@Deprecated
 		void onSuccess();
 
+		/**
+		 * Records a successful call or retryable call with the needed generated retry events.
```
RobWin: Could you please enhance this JavaDoc and describe what happens on onComplete to the user.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -539,9 +546,17 @@ default void executeRunnable(Runnable runnable) {
...
 		/**
 		 * Records a successful call.
+		 * @deprecated since 1.2.0
 		 */
+		@Deprecated
 		void onSuccess();
 
+
+		/**
+		 * Records a successful call or retryable call with the needed generated retry events.
```
RobWin: Could you please enhance this JavaDoc and describe what happens on onComplete to the user.

Romeh:
all is set based into last review comments :)

resilience4j-retry/src/main/java/io/github/resilience4j/retry/utils/MaxRetriesExceeded.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+package io.github.resilience4j.retry.utils;
```
RobWin: Can we move the exception into package `io.github.resilience4j.retry`. The exception is part of the public API.

PR #700, opened 2019-10-17
TimeLimiter with CompletionStage

dlsrb6342:
Decorating and Executing `CompletionStage` with `TimeLimiter`.

### DIFF
--- a/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
+++ b/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
@@ -74,9 +74,18 @@ public interface TimeLimiter {
         return timeLimiter.decorateFutureSupplier(futureSupplier);
     }
 
+    /**
+     * Decorate a CompletionStage supplier which is decorated by a TimeLimiter
+     * @param timeLimiter the TimeLimiter
+     * @param scheduler execution service to use to schedule timeout
+     * @param supplier the original CompletionStage supplier
+     * @param <T> the type of the returned CompletionStage's result
+     * @param <F> the CompletionStage type supplied
+     * @return a CompletionStage supplier which is decorated by a TimeLimiter
+     */
     static <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(
-            TimeLimiter timeLimiter, Supplier<F> supplier) {
-        return timeLimiter.decorateCompletionStage(supplier);
+            TimeLimiter timeLimiter, ScheduledExecutorService scheduler, Supplier<F> supplier) {
+        return timeLimiter.decorateCompletionStage(scheduler, supplier);
     }
 
     String getName();
@@ -101,8 +110,17 @@ public interface TimeLimiter {
         return decorateFutureSupplier(this, futureSupplier).call();
     }
 
-    default <T, F extends CompletionStage<T>> CompletionStage<T> executeCompletionStage(Supplier<F> supplier) {
-        return decorateCompletionStage(this, supplier).get();
+    /**
+     * Decorates and executes the CompletionStage Supplier
+     * @param scheduler execution service to use to schedule timeout
+     * @param supplier the original CompletionStage supplier
+     * @param <T> the type of the returned CompletionStage's result
+     * @param <F> the CompletionStage type supplied
+     * @return the decorated CompletionStage
+     */
+    default <T, F extends CompletionStage<T>> CompletionStage<T> executeCompletionStage(
+            ScheduledExecutorService scheduler, Supplier<F> supplier) {
+        return decorateCompletionStage(this, scheduler, supplier).get();
     }
 
     /**
@@ -115,7 +133,16 @@ public interface TimeLimiter {
      */
     <T, F extends Future<T>> Callable<T> decorateFutureSupplier(Supplier<F> futureSupplier);
 
-    <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(Supplier<F> supplier);
+    /**
+     * Decorate a CompletionStage supplier which is decorated by a TimeLimiter
+     * @param scheduler execution service to use to schedule timeout
+     * @param supplier the original CompletionStage supplier
+     * @param <T> the type of the returned CompletionStage's result
+     * @param <F> the CompletionStage type supplied
+     * @return a CompletionStage supplier which is decorated by a TimeLimiter
+     */
+    <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(
+            ScheduledExecutorService scheduler, Supplier<F> supplier);
 
     /**
      * Returns an EventPublisher which can be used to register event consumers.
--- a/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
+++ b/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
@@ -8,8 +8,7 @@ import io.github.resilience4j.timelimiter.event.TimeLimiterOnTimeoutEvent;
 import io.github.resilience4j.timelimiter.internal.TimeLimiterImpl;
 
 import java.time.Duration;
-import java.util.concurrent.Callable;
-import java.util.concurrent.Future;
+import java.util.concurrent.*;
 import java.util.function.Supplier;
 
 /**
@@ -75,6 +74,11 @@ public interface TimeLimiter {
         return timeLimiter.decorateFutureSupplier(futureSupplier);
     }
 
+    static <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(
+            TimeLimiter timeLimiter, Supplier<F> supplier) {
+        return timeLimiter.decorateCompletionStage(supplier);
+    }
+
     String getName();
 
     /**
@@ -97,6 +101,10 @@ public interface TimeLimiter {
         return decorateFutureSupplier(this, futureSupplier).call();
     }
 
+    default <T, F extends CompletionStage<T>> CompletionStage<T> executeCompletionStage(Supplier<F> supplier) {
+        return decorateCompletionStage(this, supplier).get();
+    }
+
     /**
      * Creates a Callback that is restricted by a TimeLimiter.
      *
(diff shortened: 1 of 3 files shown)


resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -55,6 +55,34 @@ public TimeLimiterImpl(String name, TimeLimiterConfig timeLimiterConfig) {
         };
     }
 
+    @Override
+    public <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(Supplier<F> supplier) {
+        return () -> CompletableFuture.supplyAsync(() -> {
+            CompletableFuture<T> future = supplier.get().toCompletableFuture();
+            try {
+                return future.get(getTimeLimiterConfig().getTimeoutDuration().toMillis(), TimeUnit.MILLISECONDS);
```
RobWin: I think we should avoid to do a blocking call.

RobWin: Please have a look at the delayer in this example.
http://iteratrlearning.com/java9/2016/09/13/java9-timeouts-completablefutures.html
What do you think?

hexmind: `executor.setRemoveOnCancelPolicy(true);` can be useful
 https://medium.com/@nilesh_mahajan/java-concurrency-pitfalls-of-scheduledexecutorservice-d7298b66e59f

RobWin:
I think we shouldn't wait for Java9. When we migrate to Java9, we can internally replace our implementaiton by `.orTimeout(...)`

dlsrb6342:
@Romeh
yeah. we need to use `Executor` not default. We have to add config like `ThreadPoolBulkheadConfig` for `executor`.
And for timeout, I'll check @RobWin 's example and jdk9's implementation.
Thanks for all reviews.

dlsrb6342:
@RobWin @Romeh @hexmind 
I changed with `delayer`. Please check. 
And one question, `CompletableFuture` cannot cancel running future like `Future`. If we call `completableFuture.cancel(true)`, it completes this CompletableFuture with CancellationException not cancel job.
What should we do with `cancelRunningFuture` in `CompletableFuture`?

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -137,4 +140,38 @@ private void publishEvent(TimeLimiterEvent event) {
             LOG.warn("Failed to handle event {}", event.getEventType(), e);
         }
     }
+
+    /**
+     * Completes CompletableFuture with {@link TimeoutException}.
+     */
+    static final class Timeout {
+
+        private static final ScheduledThreadPoolExecutor delayer;
```
hexmind: Shouldn't be  non-static ScheduledThreadPoolExecutor taken from TimeLimiterConfig? Optional parameter with default value like this one here.

dlsrb6342: `delayer` just completes the future exceptionally without doing anything. So I think it is ok with one thread. 
But, I found warnings about `CompletableFuture` in @Romeh 's [link](https://crondev.blog/2017/01/23/timeouts-with-java-8-completablefuture-youre-probably-doing-it-wrong/). Subsequent executions can be run on `delayer` if timeout occurs. 
In jdk9 implementation, it doesn't care about this and leave it to the users. And in our library, users can compose `TimeLimiter` and `ThreadPoolBulkhead` as @RobWin said in slack. so I think we don't have to worry about it either. What do you think?

hexmind: Maybe some test could show that it can work parallel without loss, not like a "synchronized" one at once?

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -75,6 +74,11 @@ static TimeLimiter of(Duration timeoutDuration) {
         return timeLimiter.decorateFutureSupplier(futureSupplier);
     }
 
+    static <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(
```
RobWin: Could you please add some JavaDoc

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -97,6 +101,10 @@ static TimeLimiter of(Duration timeoutDuration) {
         return decorateFutureSupplier(this, futureSupplier).call();
     }
 
+    default <T, F extends CompletionStage<T>> CompletionStage<T> executeCompletionStage(Supplier<F> supplier) {
```
RobWin: Could you please add some JavaDoc

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -55,6 +55,37 @@ public TimeLimiterImpl(String name, TimeLimiterConfig timeLimiterConfig) {
         };
     }
 
+    @Override
+    public <T, F extends CompletionStage<T>> Supplier<CompletionStage<T>> decorateCompletionStage(Supplier<F> supplier) {
+        CompletableFuture<T> future = supplier.get().toCompletableFuture();
```
RobWin: Line 60 already executes the Supplier (which returns a CompletionStage).
The Supplier should not be invoked in `decorateCompletionStage`. But instead a decorated Supplier should be returned.

You can have a look at `CircuitBreaker.decorateCompletionStage` and check where `supplier.get()` is invoked.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -55,6 +55,37 @@ public TimeLimiterImpl(String name, TimeLimiterConfig timeLimiterConfig) {
...
+        ScheduledFuture<?> timeoutFuture =
```
RobWin: The ScheduledFuture should not be created when `decorateCompletionStage` is invoked, but when the returned, decorated Supplier is invoked.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
RobWin: We could solve it like: `Retry.decorateCompletionStage`.
The method has a `ScheduledExecutorService scheduler` parameter which allows the user (per call) to configure a scheduler which performs the async retry.

dlsrb6342: Changed to getting `ScheduledExecutorService` from parameter like `Retry.decorateCompletionStage`.

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -55,6 +55,41 @@ public TimeLimiterImpl(String name, TimeLimiterConfig timeLimiterConfig) {
...
+                if (result != null) {
+                    if (!timeoutFuture.isDone()) {
+                        timeoutFuture.cancel(false);
+                    }
+                    onSuccess();
+                }
+
+                // exceptionally
+                if (throwable != null) {
+                    if (throwable instanceof ExecutionException) {
```
RobWin: Don't we also have to handle a `CompletionException`.
Like in https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java#L193

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -109,4 +144,33 @@ private void publishEvent(TimeLimiterEvent event) {
...
+                CompletableFuture<?> future, ScheduledExecutorService scheduler, long delay, TimeUnit unit) {
+            return scheduler.schedule(() -> {
+                if (future != null && !future.isDone()) {
+                    future.completeExceptionally(new TimeoutException());
+                }
+            }, delay, unit);
+         }
+    }
+
+    static final class DaemonThreadFactory implements ThreadFactory {
```
RobWin: Can we delete this class?

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
dlsrb6342: Yeah right. I added. Thanks

resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
dlsrb6342: Deleted!

resilience4j-timelimiter/src/test/java/io/github/resilience4j/timelimiter/TimeLimiterTest.java
```diff
@@ -8,41 +8,66 @@
...
+        assertThat(decoratedResult.getCause()).isInstanceOf(TimeoutException.class);
 
-        verify(mockFuture).cancel(true);
+        then(mockFuture).should().cancel(true);
     }
 
     @Test
-    public void shouldThrowTimeoutExceptionAndNotInvokeCancel() throws InterruptedException, ExecutionException, TimeoutException {
+    public void shouldThrowTimeoutExceptionWithCompletionStage() throws Exception {
+        Duration timeoutDuration = Duration.ofSeconds(1);
```
RobWin: Can we somehow speed up this test? Maybe reduce the timeoutduration?

dlsrb6342: Mocking is most fast way. But for credible test, I'll reduce duration.

PR #702, opened 2019-10-18
Determine whether Circuit Breaker permits calls

athlan:
Currently the only way to determine whether Circuit Breaker permits calls is to invoke `acquirePermission` or `tryAcquirePermission`.

However, `tryAcquirePermission` have some side effects. Especially in HALF-OPEN state where counters are decremented - they might be decremented twice when check is performed and then call invoked.

This PR introduces `CircuitBreakerUtil.isCallPermitted()` method to determine status of Circuit Breaker without side effects.

Use case: Client side round-robin over instances behind Circuit Breaker not picking up not operative services to not loose retires.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/CircuitBreakerUtil.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/CircuitBreakerUtil.java
@@ -1,0 +1,20 @@
+package io.github.resilience4j.circuitbreaker.utils;
+
+import io.github.resilience4j.circuitbreaker.CircuitBreaker;
+import io.github.resilience4j.circuitbreaker.CircuitBreaker.State;
+
+import static io.github.resilience4j.circuitbreaker.CircuitBreaker.State.*;
+
+public final class CircuitBreakerUtil {
+
+    /**
+     * Indicates whether Circuit Breaker allows any calls or not.
+     *
+     * @param circuitBreaker to test
+     * @return call is permitted
+     */
+    public static boolean isCallPermitted(CircuitBreaker circuitBreaker) {
+        State state = circuitBreaker.getState();
+        return state == CLOSED || state == HALF_OPEN || state == DISABLED;
+    }
+}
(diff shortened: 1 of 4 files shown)


resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/utils/CircuitBreakerUtilIIsCallPermittedTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collection;
+
+import static io.github.resilience4j.circuitbreaker.utils.CircuitBreakerUtil.isCallPermitted;
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(Parameterized.class)
+public class CircuitBreakerUtilIIsCallPermittedTest {
```
RobWin: Could you please rename it to `CircuitBreakerUtilTest`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/utils/CircuitBreakerUtil.java
```diff
@@ -0,0 +1,20 @@
...
+
+public final class CircuitBreakerUtil {
+
+    /**
+     * Indicates whether Circuit Breaker allows any calls or not.
+     *
+     * @param circuitBreaker to test
+     * @return call is permitted
+     */
+    public static boolean isCallPermitted(CircuitBreaker circuitBreaker) {
```
RobWin: Sry, I don't remember, why we need this method again 😕 
It's basically the same functionality like `circuitBreaker.tryAcquirePermission`, but without emitting an `CircuitBreakerOnCallNotPermittedEvent`. You want to check the state of a CircuitBreaker, but without producing an event, correct?

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/utils/CircuitBreakerUtilIIsCallPermittedTest.java
athlan: It's already taken by another test. junit 4 does not support nested tested classes off shelf. Unfortunately if I'd include test method from `CircuitBreakerUtilTest` it will be executed n times (as per parameters).

RobWin: Ok, I understand.
Can we then fix the typo ? :)
`CircuitBreakerUtilIsCallPermittedTest`

athlan: fixed

PR #704, opened 2019-10-19
rate limiters permits support added to supporting methods

walec51:


hexmind:
Looks good for me. Could you update https://resilience4j.readme.io/docs/ratelimiter after a merge?

PR #713, opened 2019-10-25
Improve event documentation

feliksik:
I made the documentation a bit more explicit. 

Open question: Are we now making promises that are true for the current implementation, but which should not be part of the contract?


RobWin:
Could you please fix the failing test

PR #714, opened 2019-10-27
Support decorating Future in CircuitBreaker Interface

pulkitmehra:
This PR adds functionality to decorate `Future` to `CircuitBreaker` interface. 
Here is the [Issue](https://github.com/resilience4j/resilience4j/issues/657) explaining the need.
There is an existing PR [689](https://github.com/resilience4j/resilience4j/pull/689) that includes all the discussion regarding the solution of decorating Futures.


Romeh:
@pulkitmehra  can u please fix the failing test in :

io.github.resilience4j.circuitbreaker.CircuitBreakerFutureTest > shouldDecorateFutureAndCallerRequestCancelled FAILED

    java.lang.AssertionError at CircuitBreakerFutureTest.java:132

181 tests completed, 1 failed

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerFutureTest.java
```diff
@@ -0,0 +1,204 @@
...
+    public void shouldDecorateFutureAndCallerRequestCancelled() {
+        CircuitBreaker circuitBreaker = CircuitBreaker.ofDefaults("testName");
+
+        final Future<String> future = executor.submit(() -> "Hello World");
+        future.cancel(true);
+
+        CircuitBreakerFuture<String> decoratedFuture = new CircuitBreakerFuture<>(circuitBreaker, future);
+        Throwable thrown = catchThrowable(() -> decoratedFuture.get());
+
+        assertThat(thrown).isInstanceOf(CancellationException.class);
```
RobWin: This test seems to fail sometimes. Maybe you should use a Future mock instead.

resilience4j-core/src/test/java/io/github/resilience4j/core/functions/NanosecondTimerTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Class NanosecondTimerTest.
+ */
+public class NanosecondTimerTest {
+
+    @Test
+    public void shouldPassElapsedTimetoConsumer() {
+        NanosecondTimer timer = NanosecondTimer.now();
+        final long now = System.nanoTime();
```
clgroft: Is there a way to either mock `System.nanoTime()` for the test or (more cleanly) give the `NanosecondTimer` constructor an optional `java.time.Clock` parameter and pass in a mock here? That would give us cleaner and less brittle tests. (I hadn't thought of this in the earlier comments.)

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerFutureTest.java
clgroft: Ooh, yeah, if the operation completes before `cancel(true)` is called, it won't get cancelled.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -997,4 +1014,112 @@ static CircuitBreaker of(String name, Supplier<CircuitBreakerConfig> circuitBrea
...
+                return v;
+            } catch (CancellationException | InterruptedException e) {
+                onceToCircuitbreaker.applyOnce(cb -> cb.releasePermission());
+                throw e;
+            } catch (Exception e) {
+                onceToCircuitbreaker.applyOnce(cb -> nanosecondTimer.elapsed(
+                        time -> cb.onError(time, TimeUnit.NANOSECONDS, e)));
+                throw e;
+            } finally {
+                future.cancel(true);
```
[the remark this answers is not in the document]
RobWin: Good question?

resilience4j-core/src/test/java/io/github/resilience4j/core/functions/NanosecondTimerTest.java
RobWin: We should not introduce a new class like `NanosecondTimer `. We should avoid to create class per call just to measure the time. This avoids GC pressure.

We usually use a MockClock to simulate the time.

Please have a look at https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachineTest.java#L46-L55

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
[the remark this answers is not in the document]
pulkitmehra: It doesn't harm if you request cancel when future is evaluated (either with result/exeception). It become more relevant in `get(timeout)` when timeout occurs, caller decide to walks away and requesting task to stop before leaving. (its more of good practice in evaluating future)

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -997,4 +1014,112 @@ static CircuitBreaker of(String name, Supplier<CircuitBreakerConfig> circuitBrea
...
```
pulkitmehra: In get(timeout) when timeout occurs, caller decide to walks away and requesting task to stop before leaving. (its more of good practice in evaluating future)

RobWin: Can we catch the `TimeoutException`. I think it's missing. And can we just cancel then?

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerFutureTest.java
pulkitmehra: Thanks @clgroft , thats why it was failing intermittently, I will add `Thread.sleep()`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
pulkitmehra: Sure, I will make that change.

resilience4j-core/src/test/java/io/github/resilience4j/core/functions/NanosecondTimerTest.java
pulkitmehra: Sure, will remove `NanosecondTimer`

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
pulkitmehra: We think that we should remove `future.cancel(true);` altogether. Reason: if multiple clients share the same `Future` instance and one says `get(5, SECONDS)`, other `get(30, SECONDS)` and computation completes in 10 seconds than the later one would not get results because of the cancellation done by the first one.

pulkitmehra:
@RobWin , @Romeh , test case is fixed. Pls review!

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -997,4 +1010,105 @@ static CircuitBreaker of(String name, Supplier<CircuitBreakerConfig> circuitBrea
...
+        @Override
+        public boolean isDone() {
+            return future.isDone();
+        }
+
+        @Override
+        public T get() throws InterruptedException, ExecutionException {
+            try {
+                T v = future.get();
+                long durationInNanos = System.nanoTime() - start;
```
RobWin: Can we inline this into the lamda. It should be calculated only once.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -997,4 +1010,105 @@ static CircuitBreaker of(String name, Supplier<CircuitBreakerConfig> circuitBrea
...
+                onceToCircuitbreaker.applyOnce(cb -> cb.onError(durationInNanos, TimeUnit.NANOSECONDS, e));
+                throw e;
+            }
+        }
+
+        @Override
+        public T get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+            try {
+                T v = future.get(timeout, unit);
+                long durationInNanos = System.nanoTime() - start;
```
RobWin: Can we inline this into the lamda. It should be calculated only once.

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerFutureTest.java
```diff
@@ -0,0 +1,205 @@
...
+        } catch (InterruptedException e) {
+            executor.shutdownNow();
+        }
+    }
+
+    @Test
+    public void shouldDecorateFutureAndReturnSuccess() throws Exception {
+        CircuitBreaker circuitBreaker = CircuitBreaker.ofDefaults("testName");
+
+        final Future<String> future = executor.submit(() -> "Hello World");
```
RobWin: Can  we please mock the Future?

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerFutureTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertThat(circuitBreaker.getMetrics().getNumberOfSuccessfulCalls()).isEqualTo(1);
+        assertThat(circuitBreaker.getMetrics().getNumberOfNotPermittedCalls()).isEqualTo(0);
+        assertThat(circuitBreaker.getState()).isEqualTo(CircuitBreaker.State.CLOSED);
+    }
+
+    @Test
+    public void shouldDecorateFutureAndCircuitBreakingLogicApplyOnceOnMultipleFutureEval() throws Exception {
+        CircuitBreaker circuitBreaker = CircuitBreaker.ofDefaults("testName");
+
+        final Future<String> future = executor.submit(() -> "Hello World");
```
RobWin: Can  we please mock the Future?

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerFutureTest.java
```diff
@@ -0,0 +1,205 @@
...
+        assertThat(circuitBreaker.getMetrics().getNumberOfSuccessfulCalls()).isEqualTo(0);
+        assertThat(circuitBreaker.getMetrics().getNumberOfNotPermittedCalls()).isEqualTo(0);
+        assertThat(circuitBreaker.getState()).isEqualTo(CircuitBreaker.State.CLOSED);
+    }
+
+    @Test
+    public void shouldDecorateFutureAndThrowTimeoutException() throws Exception {
+        CircuitBreaker circuitBreaker = CircuitBreaker.ofDefaults("testName");
+
+        final Future<String> future = executor.submit(() -> {
```
RobWin: Can  we please mock the Future?

pulkitmehra:
@RobWin , pls review!

RobWin:
Sry, I think my review request was a bit misleading. What I requested was to replace the ExecutorService and Futures  in all tests by a Future mock.

pulkitmehra:
@RobWin , Pls review!

RobWin:
Thank you.

Issue #719, opened 2019-10-30
Test case failing while gradle build


KrnSaurabh:
while doing gradle build TaggedThreadPoolBulkheadMetricsPublisherTest.shouldReplaceMetrics test fails with below error.
java.lang.IllegalArgumentException: maxThreadPoolSize must be a greater than or equals to coreThreadPoolSize
I am building code on 12 processor machine which sets coreThreadPoolSize to 11. If I set maxThreadPoolSize =12 in test case, test case passes successfully.

Romeh:
yes test setup in that case need to be changed to set max Thread to  Runtime.getRuntime().availableProcessors() +1 to cover different machine core CPUs situations

PR #720, opened 2019-10-30
#719 fix flaky test data update

Romeh:


RobWin:
I guess the test will fail now

Issue #628, opened 2019-09-17
Enhance our gradle build with `checkstyle` plugin


storozhukBM:

Stupremee:
May I implement all "Enhance the gradle build" issues?
If so, which styleguide do you want for Checkstyle?

Romeh:
@storozhukBM  can you please advise @Stupremee  what is needed so he can help if possible .

storozhukBM:
Hi @Stupremee 
Thank you for your interest. I'd prefer to see separate PRs for each of these issues because they will touch basically all modules and can be hard to review if anything gets blended. Maybe you can start  with `.editorconfig` + full reformat of all files according to this new config.
As for the style guide, I don't have anything specific in mind. My initial plan was to scan other popular projects like https://github.com/spring-projects/spring-framework/blob/master/.editorconfig and some others and gather few alternative configs and list them here for @RobWin, @Romeh, and others to discuss.

Stupremee:
Okay. I will add a .editorconfig file. For checkstyle I need some more information of which code style you want to follow. You can have a look at some common codestyles (e.g. [Google Sytel Guide](https://google.github.io/styleguide/javaguide.html), [Twitter](https://github.com/twitter/commons/blob/master/src/java/com/twitter/common/styleguide.md) or [Spring](https://github.com/spring-projects/spring-framework/wiki/Code-Style) ) and modify them if needed.

storozhukBM:
Great suggestion @Stupremee
@Romeh, @RobWin, @hexmind, @dlsrb6342
If you have any preferences, feel free to express them.
I'm pretty familiar with google style guide, but will go through twitter and spring ones, and will get back to you shortly.

hexmind:
I think we need non-invasive style guide that is similar to existing state as much as possible. Revolution in automated reformatting can darkens the git history...

storozhukBM:
As for now I like Twitter formatting Guide except I'd make indent size = 4 columns (spaces).

Stupremee:
Okay. I will wait a few days to get some more opinions and then add a .editorconfig file and try to add some pmd and checkstyle rules

RobWin:
I agree to @hexmind.
We shouldn't reformat too much.

dlsrb6342:
> As for now I like Twitter formatting Guide except I'd make indent size = 4 columns (spaces).

I agree to @storozhukBM. And I think overall reformatting is inevitable because most of our code is not formatted in one style

[...]

PR #721, opened 2019-10-31
Add .editorconfig

Stupremee:
- Add a .editorconfig file
- Format every file with the Twitter Styleguide rules

An editorconfig file was request in #627


RobWin:
Compilation of the tests failed unfortunately.

```
home/travis/build/resilience4j/resilience4j/resilience4j-core/src/test/java/io/github/resilience4j/core/StringUtilsTest.java:9: error: cannot find symbol
        assertThat(StringUtils.isNotEmpty("bla")).isEqualTo(true);
```

RobWin:
AAAAh, this PR is difficult. Whenver we merge some other PR, there is a merge conflict. 
Can you rebase one last time, I will directly merge it afterwards.

Stupremee:
I have no idea why every commit exists two times now, but it can be merged

worldtiki:
Just noticed that this change also committed .idea specific files.

The codeStyleConfig.xml is managed by the editor and after this change it clashes with my configs (and I'm sure many other folks). Intellij automatically adds this line:
`<option name="USE_PER_PROJECT_SETTINGS" value="true" />`.

If I revert it, it will just put it back. This makes it hard to work with this project. Simple things like switching branch are no longer possible because of these local changes.

RobWin:
What is your suggestion? I think we should remove the .idea specific files again.

worldtiki:
> What is your suggestion? I think we should remove the .idea specific files again.

I agree. But maybe someone else can comment on this or suggest an alternative.

dlsrb6342:
I think we can do this as spring-cloud did. 
https://github.com/spring-cloud/spring-cloud-build#intellij-idea
Manage xml files in another directory not .idea and Add intelliJ setup guide on README.md or CONTRIBUTING.md

RobWin:
Good idea

hexmind:
Maybe we can also move more settings from xml to .editorconfig?
> Generic IntelliJ IDEA options that have the ij_ prefix and are applicable to all languages
https://blog.jetbrains.com/idea/2019/06/managing-code-style-on-a-directory-level-with-editorconfig/

worldtiki:
My 2 cents. I like @hexmind's suggestion and the "batteries included" approach.
It's really cool if folks can just clone the project and get on with things rather than following extra manual steps, even if that has a small cost of this approach not being able to define every formatting rule.

dlsrb6342:
I think we have to put all formatting rules.  
we will add checkstyle plugin(#628) and check rules in PR. So if we don't put all rules, contributors have to reformat their code after PR submitted. 
So I think we need to provide configs so that they can check formatting rules in their local env.

Issue #607, opened 2019-09-09
Spring Boot health status switch to DOWN when circuit breaker is opened


jhaeyaert:
Hello,

I'm using circuit breaker in a reactive Spring Boot application and I noticed that there were an odd behavior with global health of my application when a circuit breaker is opened. 

By default, all circuit breakers contribute to the health indicator and when a circuit breaker goes opened, its corresponding health status switch to DOWN, so the global health status too. 

This behavior has a huge repercussion because when a remote service used by the application is in a bad shape, all instances of the application open the circuit breaker. As a consequence all health statuses switch to DOWN so all instances are kicked out of the load balancer and the application became totally unavailable.

We could consider this behavior makes sense but, sometimes, even a non critical functionality can lead to a service interruption. It appears that the circuit breaker supposed to prevent application from being unavailable just do the opposite declaring the whole application as DOWN.

Given that, does it make sense to have such a behavior by default ? Maybe we could consider that status health check must remain UP (or UNKNOWN) whatever the state of the circuit breaker (CLOSED, OPEN, ...) ? Another solution could be disabling auto configuration of health check by default and let the developer enable health check explicitly for each circuit breaker ?

RobWin:
Hi,

I agree. The CircuitBreaker health indicators were developed long before we added Dropwird Metrics, Prometheus and Micrometer support. I would actually like to remove the Health Indicators completely, but we can't, because we want to be backward-compatible.
But you can disable health indicators globally or per CircuitBreaker instance via the configuration.

For example:

```
resilience4j.circuitbreaker:
    configs:
        default:
            registerHealthIndicator: false
```

jhaeyaert:
Thanks for your response.

It seems that even when creating a default configuration it is still necessary to explicitly inherit from for each circuit breaker instance. That's good to me but I think it is not a safe solution at long term cause I think most of developers don't expect this behavior when they implement circuit breaker (I was expecting more resilience of my application and it appears that, by default, adding a circuit breaker increase the probability of unavailability of my application).

Maybe It could be interesting to change the default behavior (or remove health) in a future major version (v1) and provide a migration guide to explicitly detail breaking changes.

Thanks

RobWin:
Yes, but you only have to inherit from default, if you want to overwrite the default properties, otherwise not.

KaiStapel:
Hi. I agree with @jhaeyaert. Not changing health status on open circuits should be the default. This is about increasing resilience, not decreasing it.

[...]

PR #722, opened 2019-10-31
Configurable health indicators impact

worldtiki:
- Configurable health indicator impact in overall status with `allowHealthIndicatorToFail` property.
- Health indicators are now registered by default since there is no longer the danger of them causing the application healthchecks to fail.

Edit: Health indicators still need to manually registered with the `registerHealthIndicator` property


resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/circuitbreaker/monitoring/health/CircuitBreakersHealthIndicator.java
```diff
@@ -61,25 +61,33 @@ public CircuitBreakersHealthIndicator(CircuitBreakerRegistry circuitBreakerRegis
...
-                .collect(Collectors.toMap(CircuitBreaker::getName, CircuitBreakersHealthIndicator::mapBackendMonitorState));
+                .collect(Collectors.toMap(CircuitBreaker::getName, this::mapBackendMonitorState));
 
         return healthAggregator.aggregate(healths);
     }
 
     private boolean isRegisterHealthIndicator(CircuitBreaker circuitBreaker) {
         return circuitBreakerProperties.findCircuitBreakerProperties(circuitBreaker.getName())
                 .map(InstanceProperties::getRegisterHealthIndicator)
+                .orElse(true);
```
dlsrb6342: We decided to disable HealthIndicator by default. Please set `false` the default value.

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/ratelimiter/monitoring/health/RateLimitersHealthIndicator.java
```diff
@@ -55,6 +57,12 @@ public Health health() {
     private boolean isRegisterHealthIndicator(RateLimiter rateLimiter) {
         return rateLimiterProperties.findRateLimiterProperties(rateLimiter.getName())
                 .map(InstanceProperties::getRegisterHealthIndicator)
+                .orElse(true);
```
dlsrb6342: We decided to disable HealthIndicator by default. Please set `false` the default value.

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/ratelimiter/monitoring/health/RateLimitersHealthIndicator.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.github.resilience4j.ratelimiter.monitoring.health;
 
+import io.github.resilience4j.circuitbreaker.CircuitBreaker;
+import io.github.resilience4j.common.circuitbreaker.configuration.CircuitBreakerConfigurationProperties;
```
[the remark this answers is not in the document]
worldtiki: Yeap, will do.

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/circuitbreaker/monitoring/health/CircuitBreakersHealthIndicator.java
worldtiki: I'm happy to do that but from what I understood from #607 they were disabled because they would cause the health status to go down.
Since that's no longer the case I wonder if we can turn them back on by default.

Thoughts?

RobWin: I think we should keep them disabled by default.

dlsrb6342: I agree with @RobWin . That change may cause confusion to users.

worldtiki: Ok, i'll make the change.

RobWin:
I wonder if we can find a property name which describes the purpose better. 

It seems we are reporting a new health status now. The Spring Boot documentation says: 
>In addition to Spring Boot’s predefined Status types, it is also possible for Health to return a custom Status that represents a new system state. In such cases, a custom implementation of the StatusAggregator interface also needs to be provided, or the default implementation has to be configured by using the management.endpoint.health.status.order configuration property.

If a CircuitBreaker reports `CIRCUIT_OPEN`. What is the overall aggregated health status? Is it UP, or is mapped to UNKNOWN, because the new status type is not known to Spring Boot? 
The unit test uses an `OrderedHealthAggregator`. But does it also work in an integration test?

worldtiki:
> I wonder if we can find a different property name than `allowHealthIndicatorToFail` which describes the purpose better.
> 
> It seems we are reporting a new health status now. The Spring Boot documentation says:
> 
> > In addition to Spring Boot’s predefined Status types, it is also possible for Health to return a custom Status that represents a new system state. In such cases, a custom implementation of the StatusAggregator interface also needs to be provided, or the default implementation has to be configured by using the management.endpoint.health.status.order configuration property.
> 
> If a CircuitBreaker reports `CIRCUIT_OPEN`. What is the overall aggregated health status? Is it UP, or is mapped to UNKNOWN, because the new status type is not known to Spring Boot?
> The unit test uses an `OrderedHealthAggregator`. But does it also work in an integration test?

From a top level the name (CIRCUIT_OPEN) doesn't matter much..

I'll put some examples here that may help with the explanation.

This is what you would get if the "allowHealthIndicatorToFail" was set to false:

```json
{
"status": "UP",
  "details": {
    "circuitBreakers": {
    "status": "UNKNOWN",
    "details": {
      "test": {
        "status": "CIRCUIT_OPEN",
        "details": {
          "failureRate": "100.0%",
          "failureRateThreshold": "20.0%",
          "slowCallRate": "0.0%",
          "slowCallRateThreshold": "100.0%",
          "bufferedCalls": 100,
          "slowCalls": 0,
          "slowFailedCalls": 0,
          "failedCalls": 100,
          "notPermittedCalls": 341,
          "state": "OPEN"
          }
        }
      }
    }
  }
}
```

And this is what you would get if the "allowHealthIndicatorToFail" was set to true:

```json
{
"status": "DOWN",
  "details": {
    "circuitBreakers": {
    "status": "DOWN",
    "details": {
      "test": {
        "status": "DOWN",
        "details": {
          "failureRate": "100.0%",
          "failureRateThreshold": "20.0%",
          "slowCallRate": "0.0%",
          "slowCallRateThreshold": "100.0%",
          "bufferedCalls": 100,
          "slowCalls": 0,
          "slowFailedCalls": 0,
          "failedCalls": 100,
          "notPermittedCalls": 341,
          "state": "OPEN"
          }
        }
      }
    }
  }
}
```

The aggregation of status is done in the `OrderedHealthAggregator`.
In short what it does is it goes over the different health status and compares them to DOWN, OUT_OF_SERVICE, UP. If they don't match (eg: CIRCUIT_OPEN) they'll be mapped to UNKNOWN.

What this means is that if we have multiple health indicators, aggregating them will result in the worst possible state.

For the global health indicator, Spring only cares about DOWN or OUT_OF_SERVICE. All others return a healthy status (200).

Examples:
1 -> UP, 2 -> DOWN, 3 -> POTATO. Aggregate result -> DOWN. Global result -> DOWN / 503
1 -> UP, 2 -> POTATO. Aggregate result -> UNKNOWN. Global result -> UP / 200

RobWin:
Thanks, could you rebase the latest changes from master?

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/ratelimiter/configuration/RateLimiterConfigurationProperties.java
```diff
@@ -56,15 +56,15 @@ public RateLimiterConfig createRateLimiterConfig(
...
     private RateLimiterConfig buildConfigFromBaseConfig(InstanceProperties baseProperties,
-        InstanceProperties instanceProperties) {
+                                                        InstanceProperties instanceProperties) {
         ConfigUtils.mergePropertiesIfAny(baseProperties, instanceProperties);
         RateLimiterConfig baseConfig = buildRateLimiterConfig(RateLimiterConfig.custom(),
             baseProperties);
         return buildRateLimiterConfig(RateLimiterConfig.from(baseConfig), instanceProperties);
     }
 
     private RateLimiterConfig buildRateLimiterConfig(RateLimiterConfig.Builder builder,
```
worldtiki: Not entirely sure what happened here.
Autoformatting aligns all the params :/

TwinkleTShah:
Hello,
When will this change be available as a released version of the library? Is there any planned release date?
Thanks.

Romeh:
Hey @TwinkleTShah  , that should be released part of next version : 1.2.0 , we are trying to finish one more item then I guess we will have a release candidate to be released soon !

TwinkleTShah:
> Hey @TwinkleTShah , that should be released part of next version : 1.2.0 , we are trying to finish one more item then I guess we will have a release candidate to be released soon !

Thanks @Romeh

RobWin:
I just release v1.2.0

saisuryakat:
Hi @RobWin, I'm currently using this dependency
```<dependency>
            <groupId>io.github.resilience4j</groupId>
            <artifactId>resilience4j-spring-boot2</artifactId>
            <version>1.1.0</version>
        </dependency>```
I want to use the **allowHealthIndicatorToFail** feature, can I explicitly override the version and use this ?
Please advise...

Issue #651, opened 2019-09-30
Configuration Documentation for Resilience4j-Spring-Boot-2


edwardbeckett:
Just thought I'd leave a comment to add a note in the configuration documentation that `MetricsAutoConfiguration` is required for resilence4j-spring-boot-2. I had `MetricsAutoConfiguration` excluded and it wasn't quite obvious why the application failed to start.

dlsrb6342:
Yes you are right. I think we have to add `MetricsAutoConfiguration ` to `@ConditionalOnClass`.

RobWin:
Hi,
you excluded `org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration`?
You don't need Micrometer metrics? 
If you don't need Micrometer, you don't have to exlude `MetricsAutoConfiguration`. Just remove Micrometer from your classpath.

edwardbeckett:
> Hi,
> you excluded `org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration`?
> You don't need Micrometer metrics?

I added this reslience-4j to an existing spring boot 2 project and had excluded metrics auto configuration for a historical requirement - after removing that exclusion I realized `@MetricsAutoConfiguration` was required.

RobWin:
Yes,  it only if you use `resilience4j-micrometer`

edwardbeckett:
@RobWin ~ I can't validate that assertion... `resilience4j-spring-boot2` has a transitive dependency on `resilience4j-micrometer`. Even if I put an explicit exclusion on `resilience4j-micrometer` the app won't start with `@MetricsAutoConfiguration` excluded.

App Exclusion
---

```
@SpringBootApplication(
        scanBasePackageClasses = {
                Application.class
        },
        exclude = {
                ErrorMvcAutoConfiguration.class,
                MetricsAutoConfiguration.class
        }
)
```

Pom Exclusion on Micrometer
---

```
        <!--Resilience4j-->
        <dependency>
            <groupId>io.github.resilience4j</groupId>
            <artifactId>resilience4j-spring-boot2</artifactId>
            <version>${resilience4j-spring-boot2.version}</version>
            <exclusions>
                  <exclusion>
                      <groupId>io.github.reslience4j</groupId>
                      <artifactId>resilience4j-micrometer</artifactId>
                  </exclusion>
              </exclusions>
        </dependency>
```

Error 
---
```
***************************
APPLICATION FAILED TO START
[...]

PR #725, opened 2019-11-02
WIP: Issue#651 WIP:Added @CondionalOnBean on **MetricsPublisher class

KrnSaurabh:
Added @ConditionalOnBean(MeterRegistry.class) on **MetricsPublisher class as suggested in Issue #651 

**Help Needed
After adding the annotation, **MetricsPublisher bean is not created and test cases in MetricsAutoConfigurationTest class are failing. I could see MeterRegistry Bean is present in BeanFactory but somehow @CondiationOnBean(MeterRegistry.class)  is not matching. I added @AutoConfigureAfter({MeterRegistry.class) to make sure meterRegistry Bean is created beforehand. I am still trying to analyse further and needed some guidance hence I have created WIP PR to get initial feedback.
Pardon me for my mistake, this is my first contribution.


KrnSaurabh:
I changed the `@ConditionalOnBean(MeterRegistry.class)` to  `@ConditionalOnBean(MetricsAutoConfiguration.class)` and unit test cases are success now. I realised MeterRegistry.class is not an auto-configuration class and that was causing issue.

KrnSaurabh:
Hi @dlsrb6342, Thanks for the guidance. I have annotated ** MetricsPublisher classes with @ConditionOnBean(MeterRegistry.class). I have created custom MeterRegistry AutoConfiguration as suggested which will trigger in case MetricAutoConfiguration is excluded.
I have also added unit test cases. Looking forward for your feedback.
Thanks.

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
+    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
```
dlsrb6342: Please remove this option

resilience4j-spring-boot2/src/main/java/io/github/resilience4j/metrics/autoconfigure/CustomMeterRegistryAutoConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Meter Registry AutoConfiguration for resilience4j-metrics when MetricsAutoConfiguration is disabled.
+ */
+@Configuration
+@ConditionalOnMissingBean(MeterRegistry.class)
+@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@AutoConfigureBefore({BulkheadMetricsAutoConfiguration.class, ThreadPoolBulkheadMetricsAutoConfiguration.class, CircuitBreakerMetricsAutoConfiguration.class, RateLimiterMetricsAutoConfiguration.class
+, RetryMetricsAutoConfiguration.class})
+public class CustomMeterRegistryAutoConfiguration {
```
dlsrb6342: We don't need this class. If `MetricsAutoConfiguration` is excluded, we don't need to make any metrics. 
Please remove this class.

resilience4j-spring-boot2/src/main/java/io/github/resilience4j/metrics/autoconfigure/package-info.java
```diff
@@ -0,0 +1,25 @@
+/*
+ *
+ *  Copyright 2018: Clint Checketts
```
dlsrb6342: Could you change to 2019?

resilience4j-spring-boot2/src/main/java/io/github/resilience4j/metrics/autoconfigure/CustomMeterRegistryAutoConfiguration.java
KrnSaurabh: Hi @dlsrb6342 I am little bit confused here as in the previous comment it was suggested to create our own meterRegistry without MetricsAutoConfiguration so I have created this class.
In case we remove this class then @ConditionalOnBean(MeterRegistry.class) does not match even with MetricAutoConfiguration enabled and fails to instantiate **MetricsPublisher class.
I am testing with below code:

```@Bean
    @ConditionalOnBean(MeterRegistry.class)
    @ConditionalOnProperty(value = "resilience4j.circuitbreaker.metrics.legacy.enabled", havingValue = "false", matchIfMissing = true)
    @ConditionalOnMissingBean
    public TaggedCircuitBreakerMetricsPublisher taggedCircuitBreakerMetricsPublisher(MeterRegistry meterRegistry) {
        return new TaggedCircuitBreakerMetricsPublisher(meterRegistry);
    }
```

And my application class does not exclude MetricsAutoConfiguration like below:

```@SpringBootApplication
public class TestApplication {
	public static void main(String[] args) {
		SpringApplication.run(TestApplication.class, args);
	}
}
```

I am getting the same error in this scenario:

```Description:
Parameter 0 of method taggedThreadPoolBulkheadMetricsPublisher in io.github.resilience4j.bulkhead.autoconfigure.ThreadPoolBulkheadMetricsAutoConfiguration required a bean of type 'io.micrometer.core.instrument.MeterRegistry' that could not be found.
```

dlsrb6342: > I think **MetricsPublisher doesn't need MetricsAutoConfiguration but MeterRegistry. We can create our own 'MeterRegistry' without 'MetricsAutoConfiguration'.

Sorry to confuse you. I mean **Users** can create their own not us.
We want to make **MetricsPublisher bean only when MeterRegistry Bean is exists. 

Could you check again your test code? If we have `@ConditionalOnBean(MeterRegistry.class) ` annotation, it should not try to make **MetricsPublisher Bean. And also, error log was occurred in `TaggedThreadPoolBulkheadMetricsPublisher`.  Please check `TaggedThreadPoolBulkheadMetricsPublisher` bean has annotation.

KrnSaurabh: Hi @dlsrb6342 , The error is coming for all **MetricsPublisher classes. I have double checked that```TaggedThreadPoolBulkheadMetricsPublisher```class has the annotation as ```@ConditionalOnBean(MeterRegistry.class)```. As I stated above, it is not able to instantiate any of **MetrcisPublisher bean even when ```MetricsAutoConfiguration``` is enabled. 
As per my analysis, this is due to ```MeterRegistry```class is not an autoConfiguration class and as per Spring docs about @ConditionalOnBean: https://docs.spring.io/ConditionalOnBean.html

> The condition can only match the bean definitions that have been processed by the application context so far and, as such, **it is strongly recommended to use this condition on auto-configuration classes only.**

If I use ```@ConditionalOnBean(MetricAutoConfiguration.class)```instead of ```@ConditionalOnBean(MeterRegistry.class)```then **MetricsPublisher bean gets initialised when ```MetricAutoConfiguration``` is enabled. When ```MetricAutoConfiguration```is disabled, it doesn't initialise **MetricsPublisher bean without any error which is the behaviour we expect. This is because ```MetricAutoConfiguration``` class is an autoConfiguration class and works fine with ```@ConditionOnBean```annotation.

Let me know your views on this.
Thanks.

dlsrb6342: About spring docs, it means, `@ConditionalOnBaen` annotation should be used in auto-configuration class. It does not mean `@ConditionalOnBean`'s parameter should be auto-configuration class. So, it doesn't matter to use `@ConditionalOnBean(MeterRegistry.class)`.

With your branch. I deleted your `CustomMeterRegistryAutoConfiguration` and published in maven local for test.
And I tested with https://github.com/resilience4j/resilience4j-spring-boot2-demo, it doesn't make any error both `MetricsAutoConfiguration` is enabled and disabled.

Could you add your test code in zip file or on github?

KrnSaurabh: Hi @dlsrb6342 , I have removed```CustomMeterRegistryAutoConfiguration```as suggested. I tested with https://github.com/resilience4j/resilience4j-spring-boot2-demo and it doesn't make any error. However it fails the existing unit test case in```MetricsAutoConfigurationTest```class as none of **MetricsPublisher Bean could get initialised. I was talking about this problem in my previous comment as well but I think I couldn't explain well. With failed unit test case, we can see even though ```MetricAutoConfiguration```is enabled, with```@ConditionalOnBean(MeterRegistry.class)```it is not able to initialise any of **MetricsPublisher Bean.
I tried a lot but looks like I am kind of stuck here and need help if I am missing something.
I have added new test cases for scenario when```MetricsAutoConfigurationTest```is disabled in the PR.

dlsrb6342: Hi @KrnSaurabh
I tested with your branch. And there is weird log for `MetricsAutoConfigurationTest`.
```console
Exclusions:
-----------

    org.springframework.boot.autoconfigure.web.servlet.error.ErrorMvcAutoConfiguration

    org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration
```

I think `TestApplication` scanned `TestApplicationWithoutMetricsAutoConfiguration`. So `MetricsAutoConfiguration` is excluded.

Please remove `TestApplicationWithoutMetricsAutoConfiguration` and use `TestApplication` in `MeterRegistryWithoutMetricsAutoConfigurationTest` like below.
```java
@RunWith(SpringJUnit4ClassRunner.class)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = TestApplication.class)
@EnableAutoConfiguration(exclude = {MetricsAutoConfiguration.class, ErrorMvcAutoConfiguration.class})
public class MeterRegistryWithoutMetricsAutoConfigurationTest {
    ...
}
```

And also, we have to add `SimpleMetricsExportAutoConfiguration.class` to `@AutoConfigureAfter` in `**MetricsAutoConfiguration` because we need `meterRegistry` bean. you can refer [`WebFluxMetricsAutoConfiguration`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/reactive/WebFluxMetricsAutoConfiguration.java).

KrnSaurabh: Hi @dlsrb6342 , Apologies for the delayed response, I was occupied with some project work lately. So I tested ```MetricsAutoConfigurationTest``` as suggested above. I removed ```TestApplicationWithoutMetricsAutoConfiguration```and used ```TestApplication```. I found all test cases in ```MetricsAutoConfigurationTest``` failing again same way. Then I removed ```MeterRegistryWithoutMetricsAutoConfigurationTest``` tests as well from the project and ran only existing tests but ```MetricsAutoConfigurationTest``` still fails.

resilience4j-spring-boot2/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/BulkheadMetricsAutoConfiguration.java
```diff
@@ -34,7 +36,7 @@
  */
 @Configuration
 @ConditionalOnClass({MeterRegistry.class, Bulkhead.class, TaggedBulkheadMetricsPublisher.class})
-@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@AutoConfigureAfter({MetricsAutoConfiguration.class, SimpleMetricsExportAutoConfiguration.class})
```
dlsrb6342: I think it works well only with `MetricsAutoConfiguration.class`.

KrnSaurabh: Hi @dlsrb6342 , As per your previous comment I understood we need to add ```SimpleMetricsExportAutoConfiguration.class``` to ```@AutoConfigureAfter``` in **MetricsAutoConfiguration. please refer to the previous comment below:
https://github.com/resilience4j/resilience4j/pull/725#discussion_r353803397
Kindly let me know if I have misunderstood something.

dlsrb6342: Oh Sorry. Please ignore this comment.

KrnSaurabh:
> Could you make same changes on resilience4j-spring-boot module?

Yeah sure, but I have one doubt. In resilience4j-spring-boot module, I don't see ```MetricsAutoConfiguration.class``` in ```@AutoConfigureAfter``` in any of ***MetricsAutoConfiguration classes. Since there is no dependency on ```MetricsAutoConfiguration.class``` in resilience4j-spring-boot module, do we really want to make same changes there ?

dlsrb6342:
In resilience4j-spring-boot module, **MetricsAutoConfiguration is for dropwizard metrics. Please check **MicrometerAutoConfiguration.

KrnSaurabh:
Hi @dlsrb6342 , I have made changes in Resilience4j-spring-boot module as suggested. Kindly review.
On an another note, I tested https://github.com/resilience4j/resilience4j-spring-boot-demo project with MetricAutoConfiguration excluded as below:
```
@SpringBootApplication(
		scanBasePackageClasses = {
				Application.class
		},
		exclude = {
				ErrorMvcAutoConfiguration.class,
				MetricsAutoConfiguration.class
		}
)
public class Application {

	public static void main(String[] args) {
		SpringApplication.run(Application.class, args);
	}

}
```
Application started successfully and did not fail as it failed with ```Resilience4j-Spring-boot2-demo```. so I was thinking again if we need to make same changes in ```Resilience4j-spring-boot``` module too. Let me know your views.

dlsrb6342:
@KrnSaurabh 
Sorry for late response.
Even though it didn't failed, `**MetricsPublisher` needs `MeterRegistry` bean. I think we should add to spring-boot module as same as spring-boot2.

KrnSaurabh:
@dlsrb6342 Sure, I got it. I have made the changes in ```Resilience4j-Spring-boot```, Kindly review.

PR #727, opened 2019-11-04
Support decorating Future with Retry

clgroft:
Partial solution to issue #657, supporting Futures.

cf: #714, #689.

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
@@ -120,7 +120,10 @@ public interface Retry {
 	}
 
 	/**
-	 * Decorates a Future Supplier with Retry.
+	 * Returns a Supplier of type Future which is decorated by Retry.
+	 *
+	 * Any delay which is specified in the retry context is accomplished by calling `Thread.sleep()`.  Thus it is not
+	 * recommended that `executor` be a single-threaded or otherwise fixed-thread-pool executor.
 	 *
 	 * @param retry    the retry context
 	 * @param executor execution service to use to pull and wait for supplied Futures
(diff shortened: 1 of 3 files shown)


resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -119,6 +119,37 @@ static Retry ofDefaults(String name) {
...
+	 * Any delay which is specified in the retry context is accomplished by calling `Thread.sleep()`.  Thus it is not
+	 * recommended that `executor` be a single-threaded or otherwise fixed-thread-pool executor.
+	 *
+	 * @param retry    the retry context
+	 * @param executor execution service to use to pull and wait for supplied Futures
+	 * @param supplier future supplier
+	 * @param <T>      type of future result
+	 * @return decorated supplier
+	 */
+	static <T> Supplier<Future<T>> decorateFuture(Retry retry, ExecutorService executor, Supplier<Future<T>> supplier) {
```
RobWin: Please have a look at `Retry.decorateCompletionStage` and `Retry.AsyncRetryBlock`.
I think we should use a ScheduledExecutorService and schedule a retry attempt. This sync version blocks with a `Thread.sleep`.

clgroft:
PTAL!

clgroft:
That should be the last of it!

resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -658,4 +691,55 @@ private void onResult(T result) {
...
+			this.scheduler = scheduler;
+			this.retryContext = retryContext;
+			this.supplier = supplier;
+			this.promise = promise;
+		}
+
+		@Override
+		public void run() {
+			try {
+				T val = supplier.get().get();
```
RobWin: Unfortunately I'm not sure if this is the correct solution.
Similar to our CircuitBreaker solution, the method `decorateFuture` should not invoke `future.get` internally. The user has not chance to define a timeout value anymore.
I guess we need to rethink this.

clgroft: It's true that this `run()` call can block for as long as the underlying `get()` call does, which is why I initially scheduled the `FutureAsyncRetryBlock` instead of simply `run()`ning it. (I'm still unclear on why we use `run()` directly.)

If we want the call to the underlying `Future` to have a mandatory timeout, that's easy enough to add. If it should be optional, I think that's better handled by the user with a `TimeLimiter`. Either way, I think the fundamental design is OK (once we resolve the schedule vs. run thing).

clgroft:
@RobWin I added code for a few alternatives (schedule, wait without explicit timeout, wait with explicit timeout). I think any of them could make sense in the right circumstances. PTAL and see whether you agree and/or think other alternatives make sense. I'll add tests and refactoring after that quick sanity check.

If this is going too slowly, ping me on Hangouts (you should be able to see my email address) and we'll discuss the design that way.

Romeh:
For me we should only support 2 options: 

1. decorated future as we do in circuit breaker already and let the user execute it by himself 
2-decoreted future with an optional executor for non blocking execution 

and discard completely the blocking option .

Romeh:
@clgroft  you still have time to finish PR :) ?

clgroft:
@RobWin @Romeh thank you for your patience here! I don't think we'll be moving forward on this soon if at all. Incorporating Resilience4j became a much lower priority for our team, and when we do it I believe we'll be adapting our own code to use CompletableFutures rather than needing more capabilities from R4j.

Romeh:
closing due to inactivity :(

PR #728, opened 2019-11-05
reactor: version updated to '3.3.0.RELEASE'

hexmind:


hexmind:
Surprising, but yes. `compose`:
> @deprecated will be removed in 3.4.0, use {@link #transformDeferred(Function)} instead

WIP only because of `Add .editorconfig #721`

RobWin:
Ok, before we update to a new version, we should check which version is used in the latest spring boot version.

hexmind:
> Ok, before we update to a new version, we should check which version is used in the latest spring boot version.

`spring-boot/2.2.0.RELEASE` depends on `reactor-core/3.3.0.RELEASE` :heavy_check_mark: 

https://mvnrepository.com/artifact/io.projectreactor/reactor-core/3.3.0.RELEASE
https://mvnrepository.com/artifact/org.springframework/spring-webflux/5.2.1.RELEASE
https://mvnrepository.com/artifact/org.springframework.boot/spring-boot/2.2.0.RELEASE

RobWin:
Hi, could you resolve the conflicts? I would like to merge the PR

PR #729, opened 2019-11-05
Add Future support to Bulkhead Interface

pulkitmehra:
This PR adds functionality to decorate Future in Bulkhead interface.
Here is the Issue explaining the need https://github.com/resilience4j/resilience4j/issues/657

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/Bulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/Bulkhead.java
@@ -25,6 +25,7 @@ import io.github.resilience4j.bulkhead.event.BulkheadOnCallRejectedEvent;
 import io.github.resilience4j.bulkhead.internal.SemaphoreBulkhead;
 import io.github.resilience4j.core.EventConsumer;
 import io.github.resilience4j.core.exception.AcquirePermissionCancelledException;
+import io.github.resilience4j.core.functions.OnceConsumer;
 import io.vavr.CheckedConsumer;
 import io.vavr.CheckedFunction0;
 import io.vavr.CheckedFunction1;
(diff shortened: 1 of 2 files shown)


resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/Bulkhead.java
```diff
@@ -575,4 +601,54 @@ static Bulkhead of(String name, Supplier<BulkheadConfig> bulkheadConfigSupplier,
...
+        public boolean isDone() {
+            return future.isDone();
+        }
+
+        @Override
+        public T get() throws InterruptedException, ExecutionException {
+            try {
+                return future.get();
+            }  finally {
+                onceToBulkhead.applyOnce(bh -> bh.onComplete());
```
pulkitmehra: `cancel` or `timeout` exception should take it as Bulkead complete operation as permission was successfully acquired or should we release permission()?

clgroft: It seems to me we asked the same question over `CircuitBreakerFuture`, and the answer we came to was to assume that `get()` or `get(long, TimeUnit)` would be called on the returned `Future` at some point. So what you have here should be fine. If it's not, then we should revisit that question for the circuit breaker case as well.

RobWin: The permission was acquired and the supplier which returns the Future was invoked. The permission must not be released, if the future is canceled afterwards.

RobWin: Good point. If nobody invokes `future.get` the sempahore is never released. I would assume this is a bug in the application of the user then. 
Yes cancel and timout exceptions should also release the permission.

resilience4j-bulkhead/src/test/java/io/github/resilience4j/bulkhead/BulkheadFutureTest.java
```diff
@@ -0,0 +1,212 @@
...
+        helloWorldService = mock(HelloWorldService.class);
+        future = mock(Future.class);
+        config = BulkheadConfig.custom()
+                .maxConcurrentCalls(1)
+                .build();
+    }
+
+    @Test
+    public void shouldDecorateSupplierAndReturnWithSuccess() throws Exception {
+        Bulkhead bulkhead = Bulkhead.of("test", config);
```
RobWin: I think it would be better to use a Bulkhead mock in every test  and verify that `onComplete` is invoked.

PR #733, opened 2019-11-06
Support decorating Future with RateLimiter

clgroft:
Partial solution to https://github.com/resilience4j/resilience4j/issues/657.


clgroft:
@pulkitmehra, @RobWin, @Romeh, PTAL!

PR #738, opened 2019-11-07
Support futures in Bulkhead Aspect.

pulkitmehra:


resilience4j-spring/src/main/java/io/github/resilience4j/fallback/FutureFallbackDecorator.java
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+        }
+
+        @Override
+        public T get(long timeout, TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+
+            //Split the time between actual get and fallback
+            Duration timeoutDivBy2 = ofNanos(unit.toNanos(timeout)).dividedBy(2);
```
pulkitmehra: I don't know how to handle this in best way. With `CompletableFuture` we execute fallback in `whenComplete` method, which runs as soon as underlying future is completes but here we execute fallback when caller actually invoke `get`. Please feel free to suggest any alternate way!

RobWin: I guess we need another FutureFallbackDecorator class.

pulkitmehra: Can you please elaborate more. In the PR, I already added `FallbackFutureDecorator`. Do you mean we add one more FutureDecorator for fallback [here](https://github.com/resilience4j/resilience4j/pull/738/files#diff-9ffbbd66fa85e7bb2df8a5e95f8e3459R58)? If yes, how will we return two futures together?

RobWin: Oh sry, I missed that.

pulkitmehra: I feel that we might have to start a new thread here which block on `get` and run fallback if main operation fails and return a promise to the caller. Should we go that route? if yes, we need a global timeout like in hystrix here

https://github.com/Netflix/Hystrix/wiki/Configuration#execution.isolation.thread.timeoutInMilliseconds

pulkitmehra: The other option would be to add an interface with another overloaded method to supply timeout for fallback

```
interface FallbackFuture<T> extends Future<T> {
  T get(long timeout, TimeUnit unit, long timeoutFallback, TimeUnit unitFallback);  
}
```
This will give opportunity to the caller to set up timeout for fallback as well. For normal get operation we use same timeout for fallback as well. We can document this behavior in Java doc. Please let us know your thoughts on this

```
@Override
        public T get() throws InterruptedException, ExecutionException {
            try {
                return future.get();
            } catch (CancellationException | InterruptedException e) {
                throw e;
            } catch (Throwable e) {
                try {
                    return ((Future<T>) fallbackMethod.fallback(e)).get();
                    return sneakyThrow(throwable);
                }
            }
        }

        @Override
        public T get(long timeout, TimeUnit unit)
            throws InterruptedException, ExecutionException, TimeoutException {

            try {
                return future.get(timeout,unit);
            } catch (CancellationException | InterruptedException e) {
                throw e;
            } catch (Throwable e) {
                try {
                    return ((Future<T>) fallbackMethod.fallback(e)).get(timeout,unit);
                    return sneakyThrow(throwable);
                }
            }
        }

        public T get(long timeout, TimeUnit unit, long timeoutFallback, TimeUnit unitFallback)
            throws InterruptedException, ExecutionException, TimeoutException {

            try {
                return future.get(timeout,unit);
            } catch (CancellationException | InterruptedException e) {
                throw e;
            } catch (Throwable e) {
                try {
                    return ((Future<T>) fallbackMethod.fallback(e)).get(timeoutFallback, unitFallback);
                    return sneakyThrow(throwable);
                }
            }
        }

        static <T extends Throwable, R> R sneakyThrow(Throwable t) throws T {
            throw (T) t;
        }
    }
```

pulkitmehra:
@RobWin need your thoughts/ideas on this issue. Thanks!

pulkitmehra:
@dlsrb6342 @RobWin @Romeh see if you have any ideas here. If we can move forward on this PR?

resilience4j-spring/src/main/java/io/github/resilience4j/fallback/FutureFallbackDecorator.java
Romeh: Fallback usually used to provide default response in case of failure so i do not feel it is right to split wait time between future call and its fallback and accordingly i do not feel it is needed to provide wait time for fallback , usually if the user want to invoke the service call in case of timeout he will enable retry in that case

Romeh:
For Aspects support for futures , for me it is a best practice to configure executor to be used to execute the future logic and do not block the main thread which execute the aspect  , same story for example when spring implement @Async.

I am completely not in favor of executing futures in blocking mode.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/Bulkhead.java
```diff
@@ -557,6 +557,17 @@ default void executeRunnable(Runnable runnable) {
         return decorateCompletionStage(this, supplier).get();
     }
 
+    /**
+     * Decorates and executes the decorated Future.
+     *
+     * @param supplier the original Future
+     * @param <T>      the type of results supplied by this supplier
+     * @return the decorated CompletionStage.
```
Romeh: I do not feel we should do that at all , it is either the wrapped future and let the user execute it or we should provide an optional executor to be used for that

Romeh:
@pulkitmehra you still have time to finish PR :) ?

Romeh:
closing due to inactivity :(

Issue #643, opened 2019-09-21
Better annotations proposal


walec51:
Problems with current annotations:

- you cannot specify more then one rate limiter
- you cannot specify the order of decorators (for example specify if rate limiter should take precedent before retries to not count them or should it be after the retryier to count the retries.

Proposal for new annotations:

```
@Retention(value = RetentionPolicy.RUNTIME)
@Target(value = {ElementType.METHOD})
@Documented
public @interface ResilienceDecorators {

  public Resilience[] value();
}

@Retention(value = RetentionPolicy.RUNTIME)
@Target(value = {ElementType.METHOD})
@Repeatable(ResilienceDecorators.class)
@Documented
public @interface Resilience {

  Retry[] retry() default {};

  RateLimiter[] rateLimiter() default {};

  @interface Retry {

    public static final String DEFAULT_CONFIG = "default";

    public String name();

    public String baseConfig() default DEFAULT_CONFIG;
  }

  @interface RateLimiter {

    static String FIXED_WEIGHT = "__FIXED_WEIGHT__";

    String name();

    int weight() default 1;

    String weightCalculator() default FIXED_WEIGHT;
  }
}

```

Usage examples on a JAX-RS interface for a REST Proxy Client framework like retrofit or si.mazi.rescu:

```
@Path("")
@Produces(MediaType.APPLICATION_JSON)
public interface Binance {

  @GET
  @Path("api/v1/exchangeInfo")
  @Resilience(retry = @Retry(name = "exchangeInfo"))
  @Resilience(rateLimiter = @RateLimiter(name = REQUEST_WEIGHT_RATE_LIMITER))
  BinanceExchangeInfo exchangeInfo() throws IOException;

  @GET
  @Path("api/v1/depth")
  @Resilience(retry = @Retry(name = "depth"))
  @Resilience(
      rateLimiter =
          @RateLimiter(name = REQUEST_WEIGHT_RATE_LIMITER, weightCalculator = "depthWeight"))
  BinanceOrderbook depth(@QueryParam("symbol") String symbol, @QueryParam("limit") Integer limit)
      throws IOException, BinanceException;

  public static int depthWeight(String symbol, Integer limit) {
    if (limit <= 100) {
      return 1;
    } else if (limit <= 500) {
      return 5;
    } else if (limit <= 1000) {
      return 10;
    }
    return 50;
  }

  @GET
  @Path("api/v1/aggTrades")
[...]

PR #744, opened 2019-11-13
WIP: New annotations to combine different Annotations #643

walec51:
Issue: #643


resilience4j-spring/src/main/java/io/github/resilience4j/decorators/configure/DecoratorsAspect.java
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * This Spring AOP aspect intercepts all methods which are annotated with a {@link Decorator}
+ * annotation. The aspect will handle methods that return a RxJava2 reactive type, Spring Reactor
+ * reactive type, CompletionStage type, or value type.
+ *
+ * The RetryRegistry is used to retrieve an instance of a Retry for a specific name.
+ *
+ * Given a method like this:
+ * <pre><code>
+ *     {@literal @}Decorator(retry = @Retry(name = "readTimeouts"))
```
RobWin: How can we guarantee that the order of the `@Decorator` annotations is applied correctly.
I think it's not ensured that the Aspects executes them in that order, or?

walec51: there is only one aspect that handles the `Decorator` annotations: `DecoratorsAspect`
it extracts an order list of `Decorator` annotations on a given method (or class) and decorates the proceeding call with respect to the order in which those 

see `DecoratorsAspect` implementations

RobWin: Yes, but as far as I remember: When a method has same annotation multiple times, Spring AOP does not guarantee that the Aspect processes them exactly in the order they are specified.

RobWin: I might be wrong. We have to test it.
That's why I always thought we have to use one annotation to combine them

```
@Decorators(value = [
    @Retry(name = "tickerAllPrices"), 
    @CircuitBreaker(name = "tickerAllPrices"),
    @RateLimiter(name = "tickerAllPrices")    
])
```

walec51: to my understanding Spring AOP checks if this criteria is true:

```
    @Pointcut(
        value = "@within(decorator) || @annotation(decorator)",
        argNames = "decorator")
    public void matchAnnotatedClassOrMethod(Decorator decorator) {
    }

    @Around(
        value = "matchAnnotatedClassOrMethod(foundDecoratorAnnotation)",
        argNames = "proceedingJoinPoint, foundDecoratorAnnotation")
```

and if it is true then it will call `decorateAroundAdvice` once for the given method call even if it has many `@Decorator` annotations.

I'll test this today in the debugger and add a junit test later.

As for you proposal

```
@Decorators(value = [
    @Retry(name = "tickerAllPrices"), 
    @CircuitBreaker(name = "tickerAllPrices"),
    @RateLimiter(name = "tickerAllPrices")    
])
```

I don't think this is doable because Java does not support annotation inheritance and `@Retry`, `@CircuitBreaker` and `@RateLimiter` would have to have a common super class to make this work.

RobWin: Ok, I missed out `ProceedingJoinPointHelper`.
Could you please check if the Apsect is only invoked once or multiple times?

walec51: guess what, when I've added a second `@Decorator` to a method it got called zero times :o)

if I add a second `@Decorator` the AOP does not see it as a `Decorator` but a `Decorators` annotation holding those two `Decorator` instances.

Pushed a commit that fixes this and now this:

```
@Component
public class RetryDummyService implements TestDummyService {

    @Override
    @Decorator(rateLimiter = @RateLimiter(name = BACKEND))
    @Decorator(retry = @Retry(name = BACKEND, fallbackMethod = "recovery"))
    public String sync() {
        return syncError();
    }
```

results in `decorateAroundAdvice` getting called once.

I'll refactor `DecoratorsAspect` to a single pointcuts. However in current form it proves that my annotation scheme is implementable.

RobWin: Very nice. Thanks

walec51:
Is the general direction this PR is taking ok?
If yes then I will proceed to polish it and add unit test.

RobWin:
Yes, the direction is fine.

resilience4j-spring/src/main/java/io/github/resilience4j/decorators/configure/DecoratorsAspect.java
```diff
@@ -0,0 +1,165 @@
...
+        List<Decorator> decoratorAnnotations = joinPointHelper.getMethodAnnotations(
+            Decorator.class);
+        if (decoratorAnnotations == null) {
+            decoratorAnnotations = joinPointHelper.getClassAnnotations(Decorator.class);
+        }
+        if (decoratorAnnotations == null) { //because annotations wasn't found
+            return proceedingJoinPoint.proceed();
+        }
+        Collections.reverse(decoratorAnnotations);
+        for (Decorator decoratorAnnotation : decoratorAnnotations) {
```
RobWin: Currently the `@Decorator` annotation allows a user to specify a Retry, CircuitBreaker and Bulkhead in one annotation. The `DecoratorsAspect` can handle it, but it should do it in the order as it is now:  `Retry(CircuitBreaker(RateLimiter(Bulkhead(Function))))`

```
Supplier<String> decoratedSupplier = Decorators.ofSupplier(supplier) 
  .withBulkhead(Bulkhead.ofDefaults("name"))     
  .withRateLimiter(RateLimiter.ofDefaults("name"))
  .withCircuitBreaker(CircuitBreaker.ofDefaults("name"))
  .withRetry(Retry.ofDefaults("name"))
```

resilience4j-annotations/src/main/java/io/github/resilience4j/decorators/annotation/Decorator.java
```diff
@@ -0,0 +1,48 @@
...
+@Target(value = {ElementType.METHOD, ElementType.TYPE})
+@Repeatable(Decorators.class)
+@Documented
+public @interface Decorator {
+
+    Retry[] retry() default {};
+
+    RateLimiter[] rateLimiter() default {};
+
+    Bulkhead[] bulkhead() default {};
```
RobWin: Can we add `ThreadBoolkBulkhead`

RobWin:
Would you like to merge the latest changes, so that we can merge your PR?

RobWin:
I tested another option which seems to work.
A user can created custom composed annotations and only add the annotations he needs:

```
@Decorators
@CircuitBreaker(name = "")
@Retry(name = "")
@Retention(value = RetentionPolicy.RUNTIME)
@Target(value = {ElementType.METHOD, ElementType.TYPE})
@Documented
public @interface ComposedDecorator {

    /**
     * Name of the decorator.
     *
     * @return the name of the decorator
     */
    String name();
}
```

The `@Decorators` annotation is used as a marker to find the custom composed annotation.
Spring has support for composed an meta-annotations. It merge the attribute `name` into the other annotations.

A user can use his custom composed annotation as follows:

```
@ComposedDecorator(name = BACKEND_A)
public String method() {
  ....
}
```

Internally a DecoratorsInterceptors applies the Annotations in the order has they are specified in the composed annotation:

```
return Decorators.ofCheckedSupplier(invocation::proceed).
  .withCircuitBreaker(circuitBreakerRegistry.circuitBreaker(circuitBreakerAnnotation.name()))
  .withRetry(retryRegistry.retry(retryAnnotation.name())
  .get()
```

walec51:
sorry for the long break, I didn't have time to polish this PR
I will resolve conflicts and add more tests next week

RobWin:
I created a little playground to test another approach. Would you like to have a look at it and add your thoughts? I described the approach above your comment.

 https://github.com/resilience4j/resilience4j-spring-boot2-demo/tree/composed-annotation/src/main/java/io/github/robwin/annotation

walec51:
as for the annotation you proposed I do not see how could I add two rate limiters or decide when ether the rate limiter should take precedence over a retry or not <- being able to do things like this is the whole point of this PR

please look at the use cases I gave at #643 and private an example solution in your convention

RobWin:
The order of the annotations is important:

```
@Decorators
@CircuitBreaker(name = "")
@Retry(name = "")
```

results in

```
return Decorators.ofCheckedSupplier(invocation::proceed).
  .withCircuitBreaker(circuitBreakerRegistry.circuitBreaker(circuitBreakerAnnotation.name()))
  .withRetry(retryRegistry.retry(retryAnnotation.name())
  .get()
```

vs

```
@Decorators
@Retry(name = "")`
@CircuitBreaker(name = "")
```

results in

```
return Decorators.ofCheckedSupplier(invocation::proceed).
   .withRetry(retryRegistry.retry(retryAnnotation.name())
  .withCircuitBreaker(circuitBreakerRegistry.circuitBreaker(circuitBreakerAnnotation.name()))  
  .get()
```

But you are right I forgot that you want to add multiple RateLimiters.
We could make each annotations repeatable so that you can do

```
@Decorators
@RateLimiter(name = "")`
@RateLimiter(name = "")`
@CircuitBreaker(name = "")
```

The advantage of composed annotations is that users could define their custom annotations and even add Spring annotations.

@Decorators
@CircuitBreaker(name = "")
@Retry(name = "")
@Retention(value = RetentionPolicy.RUNTIME)
@Target(value = ElementType.TYPE)
@Documented
@Service
@Transactional
public @interface ProtectedService {

}

RobWin:
In the first step we could modify the RateLimiterAspect to allow repeatable RateLimiter annotations, e.g.:

@RateLimiter(name = "perSecondLimit")
@RateLimiter(name = "perDayLimit")
@Component
public class ProtectedService {
}

In the second step we could allow custom composed annotations, e.g.:

```
@Decorators
@CircuitBreaker(name = "")
@RateLimiter(name = "perSecondLimit")
@RateLimiter(name = "perDayLimit")
@Retention(value = RetentionPolicy.RUNTIME)
@Target(value = ElementType.TYPE)
@Documented
@Component
@Transactional
public @interface ProtectedService {

    @AliasFor(annotation = CircuitBreaker.class, attribute = "name")
    String name();
}
```

walec51:
i need my solution to not be restricted only to Spring

for my annotations I have already a working implementation of an interceptor for https://github.com/mmazi/rescu for example

is there a library that can handle these type of annotations in your convetion without dependencies to Spring ?

besides that, in your convention, will I be able to do something like this:

```
interface SomeClient {

  @Retry(name = "failedCall")
  @RateLimiter(name = "callsPerHour")
  @Retry(name = "failedConnection")
  void callApi();
}
```

which is equivalent to:

```
return Decorators.ofCheckedSupplier(someClient::callApi).
  .withRateLimiter(retryRegistry.retry("failedCall"))
  .withRateLimiter(rateLimiterRegistry.rateLimiter("callsPerHour"))
  .withCircuitBreaker(circuitBreakerRegistry.circuitBreaker("failedConnection"))
  .get()
```

will I be forced to do add a custom annotation to accomplish this?

RobWin:
Yes the first step is to support repeatable annotations (except for `@Bulkhead`). A lot of your code can be used. So your work is not for nothing, even if we would implementation it a little bit differently.
Spring's AnnotationUtil and AnnotationElementUtil could potentially be used in any other framework to allow composed annotations.

walec51:
I have looked at your proposal and I do not see any way how to determine the order of different repeatable annotations

you do not seam to understand what I'm trying to achieve

I want this 

```
interface SomeClient {

  @Retry(name = "failedCall")
  @RateLimiter(name = "callsPerHour")
  @Retry(name = "failedConnection")
  void callApi();
}
```

to be equivalent to this:

```
return Decorators.ofCheckedSupplier(someClient::callApi).
  .withRetry(retryRegistry.retry("failedCall"))
  .withRateLimiter(rateLimiterRegistry.rateLimiter("callsPerHour"))
  .withRetry(retryRegistry.retry("failedConnection"))
  .get()
```

in the code you suggested I see no way to get a list of different annotations in the order they annotated a method

`AnnotationUtils` can only get a list for one given annotation type, I see no method which for the given example above would return a list like: [failedCallRetry, callsPerHourRateLiiter, failedConnectionRetry]

if you do not want to support the use case I have given above I will not waste any more time on this

I have stated from the start that this is crucial for me and implementation that does not support it is useless for me

thomas-mc-work:
It would be a pity to quit this nice PR just due to a misconception.

@RobWin Is it really required to have repeatable annotations? Can your active PR about repeatable annotations be a key for this problem?

@walec51 Robert has stated that a lot of your code can be used. Maybe it's a step forward to break the problem down into parts?

It would be lovely to see this problem solved in the end!

walec51:
@thomas-mc-work it is impossible reimplement this PR using repeatable annotations without a loss o functionality 

I have demonstrated a use case with is impossible to be achieved with them

but @RobWin seams to not care or understand the use case / feature I've stated I need to support event before I started work on this code which resulted in a waste of my time

PR #749, opened 2019-11-15
Issue 509 spring support for custom tags 

Romeh:
add the spring boot support for configuring metrics custom tags , right now we merge all tags definition inside base config and instance configuration before registering related meters as micrometer underneath use Prometheus registry which does not allow adding extra customs tags to an already added gauge .


resilience4j-core/src/main/java/io/github/resilience4j/core/Registry.java
```diff
@@ -75,6 +76,11 @@
      */
     C getDefaultConfig();
 
+    /**
+     * @return global configured registry tags
+     */
+    Map<String, String> getRegistryTags();
```
RobWin: I would prefer to call it `getTags` as well

Romeh: done

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/circuitbreaker/autoconfigure/AbstractCircuitBreakerConfigurationOnMissingBean.java
```diff
@@ -56,12 +56,18 @@ public CircuitBreakerRegistry circuitBreakerRegistry(
...
                 circuitBreakerRegistryEventConsumer);
 
-        // Register the event consumers
-        circuitBreakerConfiguration
-            .registerEventConsumer(circuitBreakerRegistry, eventConsumerRegistry);
+		// Register the event consumers
+		circuitBreakerConfiguration.registerEventConsumer(circuitBreakerRegistry, eventConsumerRegistry);
+		io.vavr.collection.HashMap<String, String> mergedTags = io.vavr.collection.HashMap.empty();
+		circuitBreakerProperties.getInstances().values().forEach(instanceProperties -> {
+			instanceProperties.getTags().forEach(mergedTags::put);
```
dlsrb6342: `io.vavr.collection.HashMap.put()` returns new `HashMap`.
https://github.com/vavr-io/vavr/blob/b19ce1065a74d2364569f1edf0c9773c59e0ff26/src/main/java/io/vavr/collection/HashMap.java#L781-L784

So `mergedTags` is still empty after `forEach`.
I think it have to be changed like `mergedTags = mergedTags.put(key, value)`.

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/AbstractMetrics.java
```diff
@@ -37,4 +41,10 @@ void removeMetrics(MeterRegistry registry, String name) {
         }
         meterIdMap.remove(name);
     }
+
+    List<Tag> mapToTagsList(Map<String, String> tagsMap) {
```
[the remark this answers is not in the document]
Romeh: what will be the added value :)? Tags is an array of Tag at the end ;)

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/circuitbreaker/autoconfigure/AbstractCircuitBreakerConfigurationOnMissingBean.java
Romeh: a very good catch man ! let me tackle it then

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/TaggedCircuitBreakerMetrics.java
```diff
@@ -63,7 +63,7 @@ public static TaggedCircuitBreakerMetrics ofCircuitBreakerRegistry(MetricNames m
     @Override
     public void bindTo(MeterRegistry registry) {
         for (CircuitBreaker circuitBreaker : circuitBreakerRegistry.getAllCircuitBreakers()) {
-            addMetrics(registry, circuitBreaker);
+            addMetrics(registry, circuitBreaker, circuitBreakerRegistry);
```
dlsrb6342: Is it needed? 
If instance(CircuitBreaker) is made from registry, instance has not only instance's tag but also registry's tags. So I think it can be `addMetrics(registry, circuitBreaker)`.

What do you think?

RobWin: How do we proceed?

Romeh: it is not needed , code cleaned that was leftovers from first solution approach to merge all tags

Issue #742, opened 2019-11-12
Propagate distributed tracing information across Thread pool bulkhead


bmorris591:
Using using Spring Cloud Sleuth distributed tracing, one needs to take care to handover trace information across threads.

The way Sleuth does this is generally by wrapping beans in the application context with tracing ones. Specifically in the case [`TraceableExecutorService`][1].

Currently in `resilience4j-spring-boot2` I see no way in `ThreadPoolBulkheadConfiguration` to provide an `ExexcutorService` factory or decorator

Resilience4j version: 1.1.0

Java version: 11

Problem description:

Spring Cloud Sleuth propagates tracing by default using `ThreadLocal`. When handing over tasks to a thread pool care must be taken to create a new `Span` based on the current `Span` such that this action is recorded in the trace.

The usual approach to this is to [decorate the `Runnable` or `Callable`] with a tracing aware option that grabs the current context on construction and wraps execution in a scoped `Span`.

The Resilience4j Spring Boot 2 integration using `@Bulkhead` wraps method calls in an aspect, which submits the execution to an `ExecutorService` in `FixedThreadPoolBulkhead`. This is bootstrapped through `ThreadPoolBulkheadConfiguration`. As the `ExecutorService` is created in the bulkhead's construction there is no way to wrap the `ExecutorService` in a trace aware one. There is also no mechanism to supply a custom `Callable` wrapper to `BulkheadAspect#proceedInThreadPoolBulkhead`.

This precludes the use of a thread pool based bulkhead in a Spring Sleuth environment.

[1]: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceableExecutorService.java
[2]: https://cloud.spring.io/spring-cloud-sleuth/reference/html/#runnable-and-callable

bmorris591:
Suggestions:

- An `ExeuctorServiceCustomizer` in the Spring style, which is applied to created `ExecutorService` and allows for customization or wrapping; or
- A `CallableCustomizer` in the Spring style, which is applied to created `Callable` and allows for customization or wrapping
- A `ThreadPoolBulkheadCustomizer` in the Spring style, which is applied to created `ThreadPoolBulkhead` and allows for customization or wrapping

The first may be fragile as it allows for creating an `ExecutorService` that doesn't conform to the contracts of thread pool bulkheading (fixed queue size).

[...]

Issue #565, opened 2019-08-14
Pass Threadlocal across "@Bulkhead" annotation


pulkitmehra:
Hi team,

We are trying to migrate from Hystrix to Resiliency4j. We have context value set in ThreadLocal. Whenever we go across `@Bulkhead` annotation the ThreadLocal is empty

```
    @Bulkhead(name = "backendA", type = Type.THREADPOOL)
    public CompletableFuture<String> futureSuccessReqCtx() { 
        //empty thread local here
    }
```

In Hystrix we were copying thread local using `HystrixConcurrencyStrategy` . Do we have anything equivalent in Resiliency4j. If not, please suggest any other approach.

RobWin:
Hi, there is no equivalent yet.
May I ask what is stored in the threadLocal?

You could use Reactor and Context which is supported. https://projectreactor.io/docs/core/release/reference/#context

pulkitmehra:
We have some business entities that is stored in ThreadLocal. Unfortunately, we are using SpringBoot2 and do not support reactive paradigm. 

What is the best way of tacking this?  If you tell us approach we can contribute the solution to Resiliency4j.

RobWin:
You can't forward them as a parameter of the method signature?

RobWin:
Do you need to run the code in a thread pool? Can you use the semaphore-based bulkhead?

pulkitmehra:
> You can't forward them as a parameter of the method signature?

that will pollute our signatures with unnecessary parameters.So not this option.

>  Do you need to run the code in a thread pool? Can you use the semaphore-based bulkhead?

We don't want to limit the team or restrict them.  

We can do something similar to `HystrixConcurrencyStrategy` in class `FixedThreadPoolBulkhead`

Romeh:
I am not in favor to add support of Thread Local management to the bulkhead as its own issues are far more than its own advantages and it can easily go wrong in multi threading executions if no proper management and cleanup in place.

adrilo:
+1 to this improvement.

In our microservices codebase and for distributed tracing, we pass a correlation id from service to service to be able to follow a request. This correlation id is stored in a `ThreadLocal` (and we don't want to pass it explicitly as method arguments for obvious reason). When switching threads, we need to transfer this data to not break our tracing.

[...]

PR #761, opened 2019-12-03
Thread Local Context Propagator Support

pulkitmehra:
PR for issue https://github.com/resilience4j/resilience4j/issues/565 and related to https://github.com/resilience4j/resilience4j/issues/742


pulkitmehra:
@RobWin , please review.

pulkitmehra:
@RobWin , @Romeh , please review the PR. thanks!

RobWin:
Yes, I will do it.

RobWin:
Implementation looks fine to me.
@Romeh @dlsrb6342 What do you think?

pulkitmehra:
@dlsrb6342 , Thanks for the review. Multiple options are possible here

1. User can always extend ContextPropagator to make a `CompositeContextPropagator` which has a list of `ContextPropagator` inside it. We can document this.
2. We can take multiple `ContextPropagator` as input from properties and run each of them.

Let me know which one is more preferred and I will make that change. Thanks!

dlsrb6342:
I prefer to first one. We can make it like [`CompositeRegistryEventConsumer.java`](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-core/src/main/java/io/github/resilience4j/core/registry/CompositeRegistryEventConsumer.java).
@RobWin @Romeh  What do you think?

RobWin:
Yes, I also prefer option 1)

Romeh:
Option 1 for me as well !

pulkitmehra:
@RobWin @dlsrb6342, @RobWin  , When I started implementation I realized that `CompositeContextPropagator` not possible because of following reasons

1. We are taking ContextPropagator class as input from the user and instantiating default constructor via Reflection. So there is no way of passing constructor argument from YAML file.
 
```
    backendD:
      maxThreadPoolSize: 1
      coreThreadPoolSize: 1
      queueCapacity: 1
      contextPropagator:
        - io.github.resilience4j.TestThreadLocalContextPropagator
        - io.github.resilience4j.TestThreadLocalContextPropagatorTwo
```
2. ContextPropagator methods has to be called in certain way. First, retrieve() should be called , response from it should be passed to copy() and then later to clear(). This interdependency is not possible in CompositeContextPropagator, It will lead to change in interface, return a List and accept list.

Let me know if you guys have a better way than what I have currently. Thanks!

dlsrb6342:
What about taking `ContextPropagator` instance not class?

In spring, we can autowire `ContextPropagator ` beans. 
https://github.com/resilience4j/resilience4j/blob/da4ec93e7ffbeb7e06c27eafc59788e8d9f058af/resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerConfiguration.java#L67-L73

Is there any reason taking a class?

RobWin:
How do we proceed?

pulkitmehra:
We can use `ContextPropagator` spring bean instead of passing class from properties( instantiating a no arg constructor on it). 

High level changes I foresee here is, In `resilience4J-spring` module, we add a bean in `ThreadPoolBulkheadConfiguration` class

```
 @Bean
    @Primary
    public ContextPropagator threadPoolContextPropagator(Optional<List<ContextPropagator>> contextPropagators){
        return new CompositeContextPropagator(contextPropagators);
    }
```

We forward this bean to `ThreadPoolBulkheadConfigurationProperties` in `resilience4j-framework-common` module, which further pass it to `ThreadPoolBulkheadConfig`. In `ThreadPoolBulkheadConfig.Builder`, we will have a setter method for adding a bean instead of Class. For example,
```
public final Builder contextPropagator(ContextPropagator... contextPropagators) {
            this.contextPropagators = contextPropagators != null
                ? contextPropagators
                : null
            return this;
        }
```
instead of current.
```
public final Builder contextPropagator(
            @Nullable Class<? extends ContextPropagator>... contextPropagatorClasses) {
            this.contextPropagatorClasses = contextPropagatorClasses != null
                ? contextPropagatorClasses
                : new Class[0];
            return this;
        }
````
 Once we have bean in `ThreadPoolBulkheadConfig`, then we can always use it in `FixedThreadPoolBulkhead` class.

In above setup, We won't be having any properties based set up for adding a `ContextPropagator`, like we have it for `ignoreException` in CircuitBreaking. It depends upon bean present in the Spring context or not. 

Let me know what you guys think and I can make these changes.

**Comment by RobWin** [comment]
Thank you for your proposal.
Your last point is crucial. I guess users might want to use different ContextPropagators für different ThreadPoolBulkheads. I guess it's not good if the ContextPropagator beans are applied to all ThreadPoolBulkheads by default. 

Another solution could be to allow the customization of ThreadPoolBulkhead instances via Customizers.
In order to support a customization  of ThreadPoolBulkhead instances, we could add an alternative which is comparable to how Spring Cloud CircuitBreaker works. Please see their [example](https://cloud.spring.io/spring-cloud-circuitbreaker/reference/html/#specific-circuit-breaker-configuration).
A user can define Customizer beans which allow to configure ThreadPoolBulkhead properties (besides the config file). This would allow also a mixture of configuration via code and config file.
The following is just a draft/suggestion:

Either:

```
@Bean
public ThreadPoolBulkheadConfigCustomizer customizer() {
        return (bulkheadName, builder) -> {
            if (bulkheadName.equals("yourBulkheadName")) {
                builder.contextPropagator(....);
            }
            return builder;
        };
}
```

or

```
@Bean
public ThreadPoolBulkheadConfigCustomizer yourBulkheadNameCustomizer() {
        return builder -> {
            builder.contextPropagator(....);
            return new Tuple2<>("bulkheadName", builder);
        };
}
```

**Comment by pulkitmehra** [comment]
@RobWin , added a Customizer interface as suggested but only implemented in ThreadPoolBulkhead. Please review

File: resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadConfig.java [hunk 1]
```diff
@@ -82,18 +91,23 @@ public int getMaxThreadPoolSize() {
...
         return writableStackTraceEnabled;
     }
 
-    public static class Builder {
+    public List<? extends ContextPropagator> getContextPropagator() {
+        return contextPropagators;
+    }
 
-        private ThreadPoolBulkheadConfig config;
+    public void setContextPropagators(List<? extends ContextPropagator> contextPropagators) {
```
pulkitmehra: @RobWin , We have to add a setter so that Customizer can update config via bean. I only generated setter for contextPropagators.

pulkitmehra: If you think this design is good than I can add `RegistryCustomizerXXX` in other part of R4J in a separate PR(so that we have better context in reviewing).

RobWin: I'm not sure. I've to think about it. All Configs are usually immutable by design.

pulkitmehra: For `Customizer` interface to work we have to provide a way to mutate Configs.

RobWin: Yes, or the Customizer is applied before the ThreadPoolBulkheadConfig is even created. 
The customizers could be invoked in `ThreadPoolBulkheadConfigurationProperties.createThreadPoolBulkheadConfig`.

And a Customizer bean could look as follows: 

```
@Bean
public ThreadPoolPropertiesCustomizer customizer() {
        return (String instanceName, InstanceProperties properties) -> {
            if (instanceName.equals("bla")) {
                properties.setContextPropagator(TestThreadLocalContextPropagator.class);
            }
            return builder;
        };
}
```

RobWin: As an alternative we could add it in `ThreadPoolBulkheadConfiguration.threadPoolBulkheadRegistry`.
The method `bulkheadConfigurationProperties.createThreadPoolBulkheadConfig(name)` could return a builder instead of the final ThreadPoolBulkheadConfig. 
And the correct Customizer is applied to this builder.

pulkitmehra: This could be done, I can try that!

pulkitmehra: @RobWin , supporting Builders now. Please review!

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/bulkhead/configuration/ThreadPoolBulkheadConfigurationProperties.java
```diff
@@ -54,34 +57,45 @@ public InstanceProperties getBackendProperties(String backend) {
...
+
+    public ThreadPoolBulkheadConfig createThreadPoolBulkheadConfig(String backend,
+        Optional<Customizer<ThreadPoolBulkheadConfig.Builder>> customizer) {
+        return createThreadPoolBulkheadConfig(getBackendProperties(backend),backend, customizer);
     }
 
     public ThreadPoolBulkheadConfig createThreadPoolBulkheadConfig(
-        InstanceProperties instanceProperties) {
+        InstanceProperties instanceProperties, String backendName,
+        Optional<Customizer<ThreadPoolBulkheadConfig.Builder>> customizer) {
```
Romeh: Minor remark , can we use https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/lang/Nullable.html for the optional spring dependency injection as using Optional as method parameter type is really bad practice .

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/AbstractBulkheadConfigurationOnMissingBean.java
```diff
@@ -104,11 +106,19 @@ public ReactorBulkheadAspectExt reactorBulkHeadAspectExt() {
...
 
         return threadPoolBulkheadConfiguration.threadPoolBulkheadRegistry(
             threadPoolBulkheadConfigurationProperties, bulkheadEventConsumerRegistry,
-            threadPoolBulkheadRegistryEventConsumer);
+            threadPoolBulkheadRegistryEventConsumer,compositeThreadPoolBulkheadBuilderCustomizer);
+    }
+
+    @Bean
+    @Primary
+    public Customizer<Builder> compositeThreadPoolBulkheadBuilderCustomizer(
```
Romeh: what if the user want to override that via custom bean injection , here were forcing this bean as primary bean all the time , can not we just use @ConditionalOnMissingBean ?

pulkitmehra: This is similar to design of `threadPoolBulkheadRegistryEventConsumer` https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/AbstractBulkheadConfigurationOnMissingBean.java#L114-L120

We are collecting all beans of type 'Customizer` present in the Context in `CompositeCustomizer`, composing into a single wrapper bean(delegate pattern https://github.com/resilience4j/resilience4j/pull/761/files#diff-fd0197063f47bb044a0902723f6f248eR13) and then using a single bean in framework internals. If we use `@ConditionalOnMissingBean ` then 
1. everywhere we have to loop through customizers present in context(https://github.com/resilience4j/resilience4j/pull/761/files#diff-bc5195d76b00f5bb9fc1eb3d24534e11R97) 
2. Use Optional in DI because bean could be absent in the context. 

Let me know if you still think we should change this.

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/bulkhead/configuration/ThreadPoolBulkheadConfigurationProperties.java
pulkitmehra: DI is not happening here, its a helper method which is public because it is called by other configuration. But I removed `Optional` from the Customizer beans which is getting DI. See latest commit.

Romeh: yes that i what i mentioned in my comment for AbstractBulkheadConfigurationOnMissingBean class ;)  , and still u do not need Optional in the helper methods as a method parameters , just resilience4j @Nullable which indicate it can be be null and do ur null check inside the last method itself or ofNullable whatever u like

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/AbstractBulkheadConfigurationOnMissingBean.java
Romeh: @ConditionalOnMissingBean means it will be injected all the time as long as no one is overriding it as we do for for all resilience4j managed beans so for your points mentioned above :

- if no overridden bean is introduced , your confitional bean will be injected so u will not face the issues u mentioned above 
- if the user override it , then he have to respect it is own purpose as it is being shown in ur code and u will not face the issues u mentioned as well , same goes for most of other resilience managed beans 

Make sense ? or i am missing something ? please let know and thx a ot for this nice PR 👍

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/bulkhead/configuration/ThreadPoolBulkheadConfigurationProperties.java
pulkitmehra: I will make this change!

resilience4j-spring-boot-common/src/main/java/io/github/resilience4j/bulkhead/autoconfigure/AbstractBulkheadConfigurationOnMissingBean.java
pulkitmehra: Thank you for the explanation!

Application can provide multiple customizer beans so we have to make this code a List<Customizer> instead of single Customizer (we smartly fit delegate pattern using CompositeCustomizer) https://github.com/resilience4j/resilience4j/pull/761/files#diff-118f95f880f6d107c67d522fd9712617R112

Now, this list either contains
1. A composite customizer bean (if no override bean is available in the context)
2. Or list of application provided Customizer bean.

The thing that I don't like here is that if no bean override happen we have a List<Customizer> which has only one bean of type CompositeCustomizer. So, essentially a list within list.

I can make that change but want to call out this scenario. Pls confirm!

Romeh: Yes you are correct  ,  that is why maybe your composite interface should not be of a type of Customizer , your composite delegate should be the one that we use in our internal logic once it has  the list of customizers injected into it either by our bean or the overridden composite if any by the users .

So users :

1. can implement as many customizes as they wish
2. your unique composite delegate bean will collect them via spring List injection 
3. then we pass the composite delegate bean to our internal initialization logic which will enable the list of Customizers access 

Now what i do not feel optimal is we have to stream all customizers per configured instance to execute everyone of them to see if it will match the name parameter .

What if we can extend the Customizer interface by having extra method ex:  instanceName() , then through DI you can have a map where the key will be the instance name and the value will the Customizer instance and should get by name from the map if any and in that case u will not need even if the composite and it will be more optimal as you get only the configured Customizer by name and u will not need to stream all of them every-time per instance .

What do u think ?

RobWin: @romeh I agree to your points

pulkitmehra:
@RobWin @Romeh , please review.
(cc: Thanks Romeh for adding Customizer logic 👍 )

RobWin:
Could you please rebase the latest changes. Sry for that :/

RobWin:
Might be even faster to create a new PR?

RobWin:
Do you have time to work on this PR again? :)

pulkitmehra:
@RobWin , pls review. Updated after merging code with master.

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadConfig.java
```diff
@@ -188,6 +221,16 @@ public ThreadPoolBulkheadConfig build() {
                 throw new IllegalArgumentException(
                     "maxThreadPoolSize must be a greater than or equals to coreThreadPoolSize");
             }
+            if (contextPropagatorClasses.length > 0) {
+                config.contextPropagators = stream(contextPropagatorClasses)
+                    .map(c -> instantiateClassDefConstructor(c))
+                    .collect(toList());
+            }
+            //setting bean of type context propagator overrides the class type.
+            if (contextPropagators.size() > 0){
```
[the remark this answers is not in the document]
pulkitmehra: make sense, merged the list. pls review!

RobWin:
@Romeh I think this PR is ready to be merged. Do you have any findings/comments?

Issue #769, opened 2019-12-13
Duplicated auto complete support in IDEA


dlsrb6342:
Resilience4j version: 
* resilience4j-spring-cloud2:1.2.0

Java version: 
* jdk11

Problem description:
There are duplicated spring configuration metadata. 
![image](https://user-images.githubusercontent.com/19386038/70771958-5bd28c80-1db6-11ea-9b2b-e6170bb41c92.png)

RobWin:
Merged

PR #771, opened 2019-12-13
Issue #769 Remove duplicated auto complete support in IDEA

dlsrb6342:
I tested this in local. There is no duplicated auto complete support.
image

### DIFF
--- a/resilience4j-spring-boot-common/build.gradle
+++ b/resilience4j-spring-boot-common/build.gradle
@@ -3,8 +3,8 @@ dependencies {
 
     compileOnly(libraries.spring_boot2_aop)
     compileOnly(libraries.spring_boot2_actuator)
-    annotationProcessor(libraries.spring_boot2_config_processor)
-    annotationProcessor(libraries.spring_boot2_autoconfigure_processor)
+    compileOnly(libraries.spring_boot2_config_processor)
+    compileOnly(libraries.spring_boot2_autoconfigure_processor)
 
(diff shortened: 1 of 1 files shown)


resilience4j-spring-boot-common/build.gradle
```diff
@@ -3,8 +3,8 @@ dependencies {
 
     compileOnly(libraries.spring_boot2_aop)
     compileOnly(libraries.spring_boot2_actuator)
-    annotationProcessor(libraries.spring_boot2_config_processor)
-    annotationProcessor(libraries.spring_boot2_autoconfigure_processor)
+    compileOnly(libraries.spring_boot2_config_processor)
```
RobWin: Why does that change behavior? Do you know it?

dlsrb6342: If we add `spring-boot-autoconfigure-processor` as annotationProcessor, it makes `spring-autoconfigure-metadata.json` and `spring-configuration-metadata.json`.
image

The one produced by the spring-boot-common module and the one created by the downstream module(like spring-boot, spring-boot2) are duplicated. So I changed spring-boot-common module not to make those files.

RobWin: But can't we then delete the dependencies?

dlsrb6342: Yes you are right. I removed them

Issue #775, opened 2019-12-15
Getters for waitInterval of RetryOnRetryEvent


slawekjaranowski:
Resilience4j version: 1.2.0

Class `RetryOnRetryEvent` has private field `waitInterval`.
I would like to access to this information.
In standard `toString` method it is used, but in my handler I can't use it. 

We can implement it by added getters for `RetryOnRetryEvent` or extends `RetryEvent` interface of it.
When `RetryEvent` will be extended all other implementation must return some default value like zero.

Which one way is acceptable for you?

RobWin:
Hi, we should add a `getter` to RetryOnRetryEvent.
Would you like to create a PR?

slawekjaranowski:
Ok, I will create PR today or tomorrow.

PR #778, opened 2019-12-16
Getters for waitInterval of RetryOnRetryEvent

slawekjaranowski:
fix: #775

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/event/RetryOnErrorEvent.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/event/RetryOnErrorEvent.java
@@ -18,13 +18,15 @@
  */
 package io.github.resilience4j.retry.event;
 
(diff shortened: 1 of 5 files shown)


RobWin:
Could you please add a test to [RetryEventTest](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-retry/src/test/java/io/github/resilience4j/retry/event/RetryEventTest.java) :)

slawekjaranowski:
yes of course

slawekjaranowski:
I see some inconsistent in `AbstractRetryEvent` field  `lastThrowable` is annotated with @ Nullable, but in child class `RetryOnSuccessEvent`, `RetryOnErrorEvent`, `RetryOnIgnoredErrorEvent` this annotation is missed  on constructor parameter.
`RetryOnRetryEvent` has @ Nullable in constructor on this field.

I can fix it - by the way

RobWin:
Thanks

Issue #773, opened 2019-12-13
Use named configuration with `@RateLimiter` annotation


brempusz:
Resilience4j version: 1.2.0
Java version: 11.0.5

Problem description:

There is no way to pragmatically customize rate limiter applied with `@RateLimiter` annotation.

I tried:

```java
@Configuration
class RateLimiterConfiguration {
    @Bean
    public RateLimiterRegistry rateLimiterRegistry() {
        RateLimiterConfig defaultConfiguration = RateLimiterConfig.custom()
                                                                  // set some parameters
                                                                  .build();
        RateLimiterConfig customConfiguration  = RateLimiterConfig.custom()
                                                                  // set some parameters
                                                                  .build();

        RateLimiterRegistry registry = RateLimiterRegistry.of(defaultConfiguration);
        registry.addConfiguration("custom", customConfiguration);

        // tried...
        // registry.replace("custom", RateLimiter.of("custom", customConfiguration));

        return registry;
    }
}

@Component
class RateLimitedComponent {
    @RateLimiter(name = "custom")
    public void limitMe() {
    }
}
```
but because of this code:
https://github.com/resilience4j/resilience4j/blob/6d2f870e6bcb4a09d346b9888adf25dd8a09b9f8/resilience4j-spring/src/main/java/io/github/resilience4j/ratelimiter/configure/RateLimiterAspect.java#L109-L110

and then https://github.com/resilience4j/resilience4j/blob/6d2f870e6bcb4a09d346b9888adf25dd8a09b9f8/resilience4j-spring/src/main/java/io/github/resilience4j/ratelimiter/configure/RateLimiterAspect.java#L142-L143

and then
https://github.com/resilience4j/resilience4j/blob/6d2f870e6bcb4a09d346b9888adf25dd8a09b9f8/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/InMemoryRateLimiterRegistry.java#L137-L138

there is no way to customize RateLimiter in code (I know, it is still possible with properties).

What is your opinion on:

1) Add parameter to `@RateLimiter` annotation - `configurationName` and use it to retrieve `RateLimiter` instance
or
2) Use `name` from `@RateLimiter` annotation and pass it to `RateLimiterRegistry` as configuration name in `getOrCreateRateLimiter`:
```
rateLimiterRegistry
            .rateLimiter(name, name);
```
Some additional logic is needed because missing configuration will throw:
[...]

PR #784, opened 2019-12-19
Issue #773: Add `configurationName` to `@RateLimiter`

brempusz:
Just the implementation - without tests.

I tried to find proper place to test this but then I noticed something in  `RateLimiterRegistry` API (and `InMemoryRateLimiterRegistry` implementation):

https://github.com/resilience4j/resilience4j/blob/4cce868397ff6bc046a1bb70892d5b652671b6c7/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/InMemoryRateLimiterRegistry.java#L200-L204 

1st call:  `rateLimiterRegistry.rateLimiter("name", "configName1")` will cache limiter instance created and configured with `configName1`.
2nd call: `rateLimiterRegistry.rateLimiter("name", "configName2")` will silently ignore `"configName2"` and just return limiter configured with `"configName1"`.

I am not sure how the cache is implemented but it may get really ugly in multi-threaded scenario where order of calls to registry is different on each application run.

With `configurationName` parameter added to `@RateLimiter` it will be easy to introduce subtle bugs in client applications.

Maybe it would be good idea to check it like this (from top of my head):

```java
        RateLimiterConfig rateLimiterConfig = getConfiguration(configName)
            .orElseThrow(() -> new ConfigurationNotFoundException(configName));
        RateLimiter rateLimiter = computeIfAbsent(name, () -> RateLimiter.of(name, rateLimiterConfig, getAllTags(tags)));
        if(rateLimiter.getRateLimiterConfig() != rateLimiterConfig) {
            // throw or log warning?
        }
        return rateLimiter;
```

### DIFF
--- a/resilience4j-annotations/src/main/java/io/github/resilience4j/ratelimiter/annotation/RateLimiter.java
+++ b/resilience4j-annotations/src/main/java/io/github/resilience4j/ratelimiter/annotation/RateLimiter.java
@@ -36,6 +36,13 @@ public @interface RateLimiter {
     String name();
 
     /**
+     * Name of the configuration
+     *
+     * @return the name of the configuration
+     */
+    String configurationName() default "";
+
+    /**
      * fallbackMethod method name.
      *
      * @return fallbackMethod method name.
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/InMemoryRateLimiterRegistry.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/InMemoryRateLimiterRegistry.java
@@ -199,7 +199,13 @@ public class InMemoryRateLimiterRegistry extends
     @Override
     public RateLimiter rateLimiter(String name, String configName,
         io.vavr.collection.Map<String, String> tags) {
-        return computeIfAbsent(name, () -> RateLimiter.of(name, getConfiguration(configName)
-            .orElseThrow(() -> new ConfigurationNotFoundException(configName)), getAllTags(tags)));
+        RateLimiterConfig rateLimiterConfig = getConfiguration(configName)
+            .orElseThrow(() -> new ConfigurationNotFoundException(configName));
+        RateLimiter rateLimiter = computeIfAbsent(name, () -> RateLimiter.of(name, rateLimiterConfig, getAllTags(tags)));
+        if(rateLimiter.getRateLimiterConfig() != rateLimiterConfig) {
+           throw new IllegalStateException("Rate limiter '" + name + "' was already initialized with configurationName" +
+               " different that '" + configName +"'");
+        }
+        return rateLimiter;
     }
 }
--- a/resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/ratelimiter/RateLimiterMethodInterceptor.java
+++ b/resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/ratelimiter/RateLimiterMethodInterceptor.java
@@ -17,6 +17,7 @@
 package io.github.resilience4j.ratpack.ratelimiter;
 
 import com.google.inject.Inject;
+import io.github.resilience4j.core.StringUtils;
 import io.github.resilience4j.core.lang.Nullable;
 import io.github.resilience4j.ratelimiter.RateLimiterRegistry;
 import io.github.resilience4j.ratelimiter.annotation.RateLimiter;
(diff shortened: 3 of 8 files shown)


RobWin:
You can't create two RateLimiter instances with the same name, but with different configs.
If you need that you have to create two RateLimiters.

brempusz:
> You can't create two RateLimiter instances with the same name, but with different configs.
> If you need that you have to create two RateLimiters.
Yes, I know. But there is noting in the API that will prevent such use or warn about it. 
It is not obvious form Javadoc (at least for me).
 https://github.com/resilience4j/resilience4j/blob/4cce868397ff6bc046a1bb70892d5b652671b6c7/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterRegistry.java#L234-L235

If you think I should continue with this implementation then I would like to as for some guidance about how to test this. I looked at test in spring, spring boot and spring boot 2 but there is nothing like "check if annotation parameters are passed to registry".

brempusz:
I implemented simple test and it shows that: if there is a configuration for a rate limiter and you use new parameter (`configurationName`) with such rate limiter it will not work as expected (I added check for mismatched name/configuration name that throws `IllegalStateException`.
So implementation of `configurationName` is not as simple as I initially expected...

dlsrb6342:
I think this implementation is good. But Could you move configuration checking logic to `AbstractRegistry`? That logic is common in all Registries .

brempusz:
It looks like the best place to implement this is `AbstractRegistry#computeIfAbsent` but right now `AbstractRegistry` doesn't tie name of entry with name of configuration used to create the entry (you can pass any supplier and don't use configuration at all to create the entry).

From top of my head:
```Java
    protected E computeIfAbsent(String name,  String configurationName, Supplier<E> supplier) {
        C configuration = getConfiguration(configurationName) // XXX: requireNotNull(configurationName, CONFIG_NAME_MUST_NOT_BE_NULL)?
            .orElseThrow(() -> new ConfigurationNotFoundException(configurationName));

        // Use io.vavr.Tuple2<String, E> where _1() is config name and _2() is E
        // replace protected final ConcurrentMap<String, E> entryMap;
        // with protected final ConcurrentMap<String, Tuple2<String, E>> entryMap;
        Tuple2<String, E> t = entryMap2.computeIfAbsent(Objects.requireNonNull(name, NAME_MUST_NOT_BE_NULL), k -> {
            E entry = supplier.get();
            eventProcessor.processEvent(new EntryAddedEvent<>(entry));
            return new Tuple2<>(configurationName, entry);
        });

        if(!configurationName.equals(t._1())) {
            throw new IllegalStateException("Entry '" + name + "' was already initialized with configuration '" +
                t._1() + "', different that requested'" + configurationName + "'");
        }
        return t._2();
    }
```

It is also possible to use `Function<C, E> supplier` and
```
E entry = supplier.apply(configuration); // XXX: force supplier to use configuration
```

RobWin:
Just to make sure we have a common understanding.

Not every RateLimiter/CircuitBreaker/Bulkhead instance must have a corresponding configuration entry in the configurations Map. The configurations map can even only consist of a default config. and still there can be RateLimiter/CircutiBreaker/Bulkhead instances with custom configs.

The configurations Map just allows to store common configurations, which you can look up if you want to create a new instance based on a common configuration.

The configurations Map could even change over the lifetime of an application. For example you could overwrite a common configuration at runtime. Only newly created RateLimiter/Bullkhead/CircuitBreaker instances are using the updated configuration. Old RateLimiter/Bullkhead/CircuitBreake instances are still using the previous configuration until you replace the instances. That's because all instances are immutable by design.

If you use Spring Cloud Config and overwrite configurations are runtime, we are actually recreating the Registry and thus replacing all instances.

RobWin:
I really regret that I haven't called the methods `registry.circuitBreaker` or `registry.bulkhead` just `registry.getOrAdd()` or `registry.createOrGet()`.

I was inspired too much by Dropwizard Metric's MetricRegistry where you can create a Counter like `registry.counter(...)`

brempusz:
@RobWin 

> Just to make sure we have a common understanding.
> (...)

> I really regret that I haven't called the methods registry.circuitBreaker() or registry.bulkhead() just registry.getOrAdd() or registry.createOrGet().

Thanks, that explains a lot. 

So my only concern is that calling (for example) `bulkhead("myname", mySupplier, myTags)` twice with different supplier will silently ignore supplier from 2nd call and return the same instance as first call (I know you should not do it but it is just so easy to do it by mistake...).

Just to check if this would work I changed `AbstractRegistry` once again: I replaced `final ConcurrentMap<String, Tuple2<String, E>> entryMap` with `final ConcurrentMap<String, Tuple2<C, E>> entryMap` and now it is possible to verify if call to `computeIfAbsent` is consistent.  
If you are interested in such implementation please let me know - I will work on it a bit more (update tests). If not - just let me know and I will drop this PR.

RobWin:
Yes, I agree that it would be more clear to the user if the method would be called `getOrCreate` or `computeIfAbsent`.  Then it would be more clear that the  `supplier` function is only invoked when no entry exists. Maybe we should improve the JavaDoc documentation to make it more clear.

The problem is that the config (e.g. BulkheadConfig) is only one part which must be compared to check consistency. The `tags` map might be different as well.

dlsrb6342:
Hi @brempusz ,
I'm sorry to tell you that our core team discussed about your PR and decided to use annotation as just a marker point. So annotation have only instance name.
Assigning a config to an instance must be in a config file.

And we decide to add `**Customizer` for all core module. Please check #790 and #793.

brempusz:
@dlsrb6342, thanks for the update.  I am closing the PR.

PR #793, opened 2020-01-07
Issue 773: add support to rate limiter spring instance configuration …

Romeh:
…customizer


resilience4j-framework-common/src/main/java/io/github/resilience4j/common/ratelimiter/configuration/CompositeRateLimiterCustomizer.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * the composite  of any circuit breaker {@link RateLimiterConfigCustomizer} implementations.
```
dlsrb6342: Please fix typo
```suggestion
 * the composite  of any rate limiter {@link RateLimiterConfigCustomizer} implementations.
```

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/ratelimiter/configuration/CompositeRateLimiterCustomizer.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class CompositeRateLimiterCustomizer {
+
+    private final Map<String, RateLimiterConfigCustomizer> customizerMap = new HashMap<>();
+
+    public CompositeRateLimiterCustomizer(List<RateLimiterConfigCustomizer> customizers) {
+        if (customizers != null && !customizers.isEmpty()) {
+            customizerMap.putAll(customizers.stream()
+                .collect(
+                    Collectors.toMap(RateLimiterConfigCustomizer::name, Function.identity())));
```
dlsrb6342: Don't we need to check that there are multiple customizer for same name?

Romeh: we can throw an exception to show that is not allowed to have more than one customizer per instance name , is this what u r asking for :) ? or i misunderstood ?

Romeh:
> I think that if there is a method that users can add or replace Customizer after `Composite**Customizer` initialized, it would be good. What do you think?

can u please describe more when the user need to do that?  or what will be the use case ?

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/ratelimiter/configuration/CompositeRateLimiterCustomizer.java
dlsrb6342: Yes right. We can throw an exception or write on docs about supporting just one customizer per instance name.

dlsrb6342:
> > I think that if there is a method that users can add or replace Customizer after `Composite**Customizer` initialized, it would be good. What do you think?
> 
> can u please describe more when the user need to do that? or what will be the use case ?

Sorry. I misunderstood that when customizer is applied. Please never mind this comment.

PR #821, opened 2020-01-28
Remove previous meters before registering new one

dlsrb6342:
If there is old meter with same id in MeterRegistry, MeterRegistry doesn’t do anything. new one cannot be registered.
Because of this, we cannot get metrics of new instance with same id after refreshing the spring context.


RobWin:
Could you please also add a test?

dlsrb6342:
Added some tests

Issue #822, opened 2020-01-28
Resilience4j RateLimiter seems to ignore configuration


RobWin:
Thanks for raising a Resilience4j issue.
Please provide a brief description of your problem along with the versions you are using.
If possible, please also consider putting together a complete JUnit test that reproduces the issue.

Resilience4j version:
1.2.0

Java version:

Problem description:

https://stackoverflow.com/questions/59947739/resilience4j-ratelimiter-seems-to-ignore-configuration

I have a problem with Resilience4j RateLimiter

```
public static void main(final String[] args) throws InterruptedException {
    final ExternalService service = new ExternalService();
    final ExecutorService executorService = Executors.newFixedThreadPool(30);

    final RateLimiterConfig config = RateLimiterConfig.custom()
        .limitRefreshPeriod(Duration.ofSeconds(10))
        .limitForPeriod(3)
        .timeoutDuration(Duration.ofSeconds(12))
        .build();

    final RateLimiter rateLimiter = RateLimiter.of("RateLimiter", config);

    final Callable<Response<String>> callable = RateLimiter.decorateCallable(
        rateLimiter, () -> service.get(200, "OK")
    );

    executorService.submit(callable); //fine in first period
    executorService.submit(callable); //fine in first period
    executorService.submit(callable); //fine in first period
    executorService.submit(callable); //should wait 10 sec and fine in second period
    executorService.submit(callable); //should wait 10 sec and fine in second period
    executorService.submit(callable); //should wait 10 sec and fine in second period
    executorService.submit(callable); //should exit with timeout after 12 seconds
    executorService.submit(callable); //should exit with timeout after 12 seconds
    executorService.submit(callable); //should exit with timeout after 12 seconds


    Thread.sleep(Duration.ofSeconds(40).toMillis());
    executorService.shutdown();
}
```

In ExternalService I have some basic logging with localTime of responses. I think that it should work as I explained in comments, but my response is:

```
> Task :Main.main()
[12:24:53.5] Return standard response
[12:24:53.5] Return standard response
[12:24:53.5] Return standard response
[12:25:03.5] Return standard response
[12:25:03.5] Return standard response
[12:25:03.5] Return standard response
[12:25:03.5] Return standard response
[12:25:03.5] Return standard response

BUILD SUCCESSFUL in 40s
```

So it seems that the first cycle is good, but after that, FIVE next threads are allowed by RateLimiter, and the last thread is never called.

RobWin:
[...]

PR #824, opened 2020-01-28
Fix wait time calculation in AtomicRateLimiter

storozhukBM:
Bug fix of issue #822. Issue was introduced in PR #672.
Affected versions: 1.2.0
Summary: during concurrent permission reservations we were rounding down count of full cycles to wait for next permissions. So it was possible to get `permissionsPerCycle - 1` more permissions in the same cycle

### DIFF
--- a/resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
+++ b/resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
@@ -115,14 +115,13 @@ public abstract class RateLimitersImplementationTest {
             long current = time.toMillis();
             long delta = Math.abs(previousDuration - current);
             runningDeltas.add(delta);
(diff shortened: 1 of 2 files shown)


resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
```diff
@@ -66,8 +74,62 @@ public void tryAquiringBigNumberOfPermitsAtEndOfCycleTest() {
...
+        then(terminated).isTrue();
+
+
+        ArrayList<Long> runningDeltas = new ArrayList<>();
+        long previousDuration = times.get(0).toMillis();
+        for (Duration time : times) {
+            long current = time.toMillis();
+            long delta = Math.abs(previousDuration - current);
+            runningDeltas.add(delta);
+            System.out.println("" + current + " [" + delta + "]");
```
RobWin: Could you please remove the System.out.println from the test?

PR #831, opened 2020-01-31
Added withTimeLimiter and withThreadPoolBulkhead methods to Decorator…

RobWin:
Added withTimeLimiter, withThreadPoolBulkhead and withFallback methods to Decorators class.

### DIFF
--- a/resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java
+++ b/resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java
@@ -13,6 +13,7 @@ import io.github.resilience4j.retry.Retry;
 import io.github.resilience4j.timelimiter.TimeLimiter;
 import io.vavr.CheckedFunction0;
 import io.vavr.CheckedFunction1;
+import io.vavr.CheckedFunction2;
 import io.vavr.CheckedRunnable;
 
 import java.util.List;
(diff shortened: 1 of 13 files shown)


RobWin:
I had to do a change in the ThreadPoolBulkhead which is not backwards-compatible. But it should be ok, because it's more of a bugfix.

When a Runnable was submitted, the ThreadPoolBulkhead returned a Runnable instead of CompletionStage<Void>

RobWin:
This PR allows to use the Decorators API as follows:

```
CompletableFuture<String> future = Decorators
    .ofSupplier(() -> helloWorldService.returnHelloWorld())
    .withThreadPoolBulkhead(ThreadPoolBulkhead.ofDefaults("helloBackend"))
    .withTimeLimiter(TimeLimiter.ofDefaults(), Executors.newSingleThreadScheduledExecutor())
    .withCircuitBreaker(circuitBreaker)
    .withFallback(TimeoutException.class, (e) -> "Recovery")
    .get().toCompletableFuture();
```

resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java
```diff
@@ -92,13 +102,29 @@ private DecorateSupplier(Supplier<T> supplier) {
...
+
         public Supplier<T> decorate() {
             return supplier;
         }
 
         public T get() {
             return supplier.get();
         }
+
+
```
dlsrb6342: nit: could you remove empty lines?

xlucasdemelo:
Why decorateRunnable does not return a Runnable anymore?

Romeh:
> Why decorateRunnable does not return a Runnable anymore?
> 
> How can i execute a Supplier<CompletionStage> inside a ExecutorService ???

@xlucasdemelo  are you taking about Threadpool bulkhead for that question? if yes thread pool bulkhead has its own thread-pool executor service which be used to invoke ur runnable and return to completion stage for listentinh to the results of that invokation

PR #838, opened 2020-02-04
Support to extend Micrometer tags in TimeLimiter

dlsrb6342:
Related with https://github.com/resilience4j/resilience4j/pull/749. 
Added support to extend tags in TimeLimiter.

### DIFF
--- a/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
+++ b/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
@@ -61,6 +61,23 @@ public interface TimeLimiter {
     }
 
     /**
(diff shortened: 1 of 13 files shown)


resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/TimeLimiter.java
```diff
@@ -60,6 +60,23 @@ static TimeLimiter of(String name, TimeLimiterConfig timeLimiterConfig) {
         return new TimeLimiterImpl(name, timeLimiterConfig);
     }
 
+    /**
+     * Creates a CircuitBreaker with a custom CircuitBreaker configuration.
```
RobWin: TimeLimiter  instead of CircuitBreaker

PR #853, opened 2020-02-07
Text files updated:

hexmind:
README.adoc:
spaces removed
modules list simplified (duplication of User Guide)
wrong link to **Hystrix** vs Istio by Nicolas Fränkel replaced
Copyright updated

CONTRIBUTING.adoc
code formatting comment added

ISSUE_TEMPLATE
reordered and simplified

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -53,13 +53,13 @@ Bulkhead bulkhead = Bulkhead.ofDefaults("backendService");
 Supplier<String> supplier = () -> backendService
   .doSomething(param1, param2);
 
-// Decorate your call to backendService.doSomething() 
+// Decorate your call to backendService.doSomething()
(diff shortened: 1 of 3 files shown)


README.adoc
```diff
@@ -99,13 +99,7 @@ Setup and usage is described in our *https://resilience4j.readme.io/docs[User Gu
 
 == Overview
 
-Resilience4j provides several core modules and add-on modules:
-
-=== All core modules and the Decorators class
-
-* resilience4j-all
```
RobWin: I think we should keep resilience4j-all on the readme

hexmind: It is mentioned below as "TIP: For core modules package or `Decorators`(...)"

PR #865, opened 2020-02-12
Adding timelimiter support for ratpack apps

drmaas:


resilience4j-ratpack/src/main/java/io/github/resilience4j/ratpack/timelimiter/TimeLimiterMethodInterceptor.java
```diff
@@ -0,0 +1,154 @@
...
+                } catch (Throwable t) {
+                    final CompletableFuture<?> promise = new CompletableFuture<>();
+                    promise.completeExceptionally(t);
+                    return (CompletionStage)promise;
+                }
+            }).toCompletableFuture();
+            completeFailedFuture(new TimeoutException(), fallbackMethod, future);
+            return future;
+        } else {
+            return handleProceedWithException(invocation, timeLimiter, fallbackMethod);
```
RobWin: In our Spring Boot TimeLimiterAspect we throw an exception if the return type is not supported.
We currently do not support a usage of the `@TimeLimiter` annotation and a basic return types.
It could be somehow confusing to a developer that the method is "implicitly" executed in a thread pool, but the thread is blocking. I think it would be better if the developer decides it explicitly and uses a ` @Bulkhead(type = Type.THREADPOOL)` or Promise.

drmaas: @RobWin that make sense - I can fix that up

Issue #703, opened 2019-10-19
how Resilience4J cleans up circuitBreaker objects in memory


KrnSaurabh:
I am working on an aggregator project that calls more than hundreds of  external Apis to build the response. I am creating circuitBreaker instance per url and ended up creating lots of circuitBreaker instances. Though few circuitBreaker instances are rarely used, it is still occupying memory. Is there a way we can improve memory usage and do some cleanup by garbage collecting those rarely used objects and if a request comes later then Resilence4j anyways gets or creates an instance.

hexmind:
You can use `CircuitBreakerRegistry` for on demand creation (eg. using lazy Spring beans).  
Maybe you can have one `CircuitBreaker` per host, not each URL. See also https://github.com/resilience4j/resilience4j/issues/650
But if only "few instances are rarely used" I am not sure is it big optimization.

storozhukBM:
The other option would be to implement your own CircuitBreakerRegestry on top of any LRU cache library that you want.

KrnSaurabh:
Thanks @hexmind / @storozhukBM . #650  make sense.

adolfoweloy:
I think that #650 doesn't make sense for me yet. Although that issue is already closed, it could be solved if the way the circuit breakers are stored was easier to extend. I think that Spring-retry has something that allows you to replace the in-memory mechanism by anything you want (i.e something that uses an external cache such as memcached) and IMHO this is something that more developers might expect from resilience4j.
@storozhukBM , I also believe that implementing an entire `CircuitBreakerRegistry` doesn't seem to be a good option considering the amount of things to be done (maybe increasing the chance of introducing bugs).

Isn't really worthwhile allowing for a seamless way to plug the circuit breaker "manager" (also seems to be a good opportunity to contribute to this project)?

hexmind:
> Another thing that I believe that would be positive by allowing for an easier way to add a circuit breaker "manager" is that by doing that we could avoid different projects creating different solutions to the same thing. I believe that the surface area for customisation could be reduced this way.

@adolfoweloy  if your class implements `Registry<CircuitBreaker, CircuitBreakerConfig>` directly (omitting `CircuitBreakerRegistry`) there will be only 7 "CRUD" methods to implement (almost twice less). Does it sound like your CircuitBreakerManager idea?

adolfoweloy:
Thanks for your answer @hexmind. That really sounds reasonable to me.
[...]

PR #882, opened 2020-02-21
WIP: #703 -Added custom RegistryStore and builder

KrnSaurabh:
This PR is WIP for #703  and created to get initial review. I have added `RegistryStore` Interface and its default implementation that replaces the Concurrent HashMap in `AbstractRegistry` Class. Also I have added builder methods in `CircuitBreakerRegistry` so that any custom implementation of `RegistryStore` can be added to the `CircuitBreakerRegistry`.

I am still working on unit tests/ comments/ javadocs.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerRegistry.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerRegistry.java
@@ -21,10 +21,12 @@ package io.github.resilience4j.circuitbreaker;
 
 import io.github.resilience4j.circuitbreaker.internal.InMemoryCircuitBreakerRegistry;
 import io.github.resilience4j.core.Registry;
+import io.github.resilience4j.core.RegistryStore;
 import io.github.resilience4j.core.registry.RegistryEventConsumer;
 import io.vavr.collection.HashMap;
 import io.vavr.collection.Seq;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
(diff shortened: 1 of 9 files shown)


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerRegistry.java
```diff
@@ -261,4 +263,60 @@ CircuitBreaker circuitBreaker(String name,
...
+            return this;
+        }
+
+        public Builder withCircuitBreakerConfig(CircuitBreakerConfig circuitBreakerConfig) {
+            circuitBreakerConfigsMap.put(DEFAULT_CONFIG, circuitBreakerConfig);
+            return this;
+        }
+
+        public Builder withCircuitBreakerConfigDefaults() {
+            circuitBreakerConfigsMap.put(DEFAULT_CONFIG, CircuitBreakerConfig.ofDefaults());
```
RobWin: `withCircuitBreakerConfigDefaults` is not necessary.  If the config map is empty, the InMemoryCircuitBreakerRegistry creates a default config.

KrnSaurabh: @RobWin  I agree that if the config map is empty, InMemoryCircuitBreakerRegistry creates a default config but I implemented this method so that user can take conscious decision to create CircuitBreakerRegistry configs by default and does not leave the hashmap empty unknowingly. Let me know if that makes sense, I will remove the method otherwise.

RobWin: A CircuitBreakerRegistry always has a default CircuitBreakerConfig. It doesn't really matter if the user invokes `withCircuitBreakerConfigDefaults` or not. 
If this method exists, it might be confusing for a users way a CircuitBreakerRegistry has defaults, even if the user didn't invoke the method. I think we should remove it.

KrnSaurabh:
Local build is successful and all test cases are passing. Need to check why all checks have failed on travis here.

KrnSaurabh:
@RobWin with these changes implemented, do we need to update the document as well ?

RobWin:
I can update the documentation. 
But we should also apply this builder pattern to RetryRegistry, RateLimiterRegistry and BulkheadRegistry as well.

KrnSaurabh:
Yeah sure @RobWin .Is it okay if I create separate issue for each component (RateLimiter, BulkHead etc) and apply the builder pattern.

RobWin:
Yes that's fine for me

PR #884, opened 2020-02-24
Add support for RxJava 3

nickrobison-usds:
This PR adds basic support for RxJava 3.

It's essentially a copy/paste from the RxJava2 module with the appropriate import paths renamed.
I'm not very familiar with the codebase, so I'm sure I'm missing somethings, but I thought this might be a useful starting point.

It also includes a couple of changes to the test code to address some removed methods:

- Migrated `cancel()` calls on Observable to `dispose()` as that method is no longer present.
- Removed `isSubscribed()` assertion as it's not longer a public method.

Thanks for a great library!


RobWin:
Thank you. I have a look at it.

RobWin:
I've know idea yet why code coverage fails.

```
> Task :resilience4j-ratelimiter:jacocoTestReport
[ant:jacocoReport] Classes in bundle 'resilience4j-ratelimiter' do no match with execution data. For report generation the same class files must be used as at runtime.
[ant:jacocoReport] Execution data for class io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter$AtomicRateLimiterMetrics does not match.
```

RobWin:
I could imagine that because we have two modules now which are using the same package and class names, jacoco is confused. We had the comparable issue with our Spring Boot 1 and Spring Boot 2 modules.

RobWin:
Nice. Works now. Thanks.

PR #906, opened 2020-03-06
Fair permission acquisition for SemaphoreBulkhead

jihoonson:
The Javadoc of `Semaphore.tryAcquire()` says:

```java
    /**
     * Acquires a permit from this semaphore, only if one is available at the
     * time of invocation.
     *
     * <p>Acquires a permit, if one is available and returns immediately,
     * with the value {@code true},
     * reducing the number of available permits by one.
     *
     * <p>If no permit is available then this method will return
     * immediately with the value {@code false}.
     *
     * <p>Even when this semaphore has been set to use a
     * fair ordering policy, a call to {@code tryAcquire()} <em>will</em>
     * immediately acquire a permit if one is available, whether or not
     * other threads are currently waiting.
     * This &quot;barging&quot; behavior can be useful in certain
     * circumstances, even though it breaks fairness. If you want to honor
     * the fairness setting, then use
     * {@link #tryAcquire(long, TimeUnit) tryAcquire(0, TimeUnit.SECONDS) }
     * which is almost equivalent (it also detects interruption).
     *
     * @return {@code true} if a permit was acquired and {@code false}
     *         otherwise
     */
    public boolean tryAcquire() {
        return sync.nonfairTryAcquireShared(1) >= 0;
    }
```

Based on this, `SemaphoreBulkhead.tryEnterBulkhead()` behaves differently depending on the configured timeout. This PR makes the behavior consistent.

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/SemaphoreBulkhead.java
@@ -230,21 +230,14 @@ public class SemaphoreBulkhead implements Bulkhead {
      * @return true if caller was able to wait for permission without {@link Thread#interrupt}
      */
     boolean tryEnterBulkhead() {
-
-        boolean callPermitted;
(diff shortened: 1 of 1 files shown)


RobWin:
Thx

PR #914, opened 2020-03-10
adding non-coroutine kotlin wrapper functions as a convenience

drmaas:
@RobWin @bradnewman I didn't add any tests - let me know what kind of tests would add value, or if any other wrapper functions would be useful.


RobWin:
Yes, I think the extension functions make sense.
But could you please still add some tests.

drmaas:
Looking at why build failed

drmaas:
@RobWin ptal when you get a change. Build is fixed.

RobWin:
Thx

PR #926, opened 2020-03-16
CircuitBreaker record a failure when a certain result is returned for reactor operator

sulabhk95:
#384 
Add another method with predicate to compose publisher.


RobWin:
Hi,
I think instead of introducing this only to the Reactor operator, it should be implemented comparable to `RetryConfg.retryOnResult`. 
It should be part of the core of Retry so that we can also add it to RxJava2 and Spring Boot Aspects.

sulabhk95:
Oh, i see. Will check and raise another PR. Thanks for review.

Issue #939, opened 2020-03-24
actuator shows Whitelabel error page with "Invalid Event"


rusyasoft:
Resilience4j version: 1.3.1

Java version: 1.8

I've been trying resilience4j-spring-boot2-demo example, with implementation "io.github.resilience4j:resilience4j-retrofit:1.3.1"

```java
private static final String BACKEND_MINE = "backendMine";

@Retry(name = BACKEND_MINE, fallbackMethod = "getMineFallback2")
@CircuitBreaker(name = "getMineSuccessName", fallbackMethod = "getMineFallback")
public MineResponse getMine(final String userId, final Integer contentId) throws IOException {
    OutCome<MineResponse> resultResponse = RetrofitUtil.resilentProcess(mineCalls.getMineSuccess());
    if (resultResponse != null) {
        return resultResponse.getResult();
    }
    return null;
}

private MineResponse getMineFallback(final String userId, final Integer contentId, CallNotPermittedException ex) {
    MineResponse mineResponse = new MineResponse();
    mineResponse.setField1(1);
    mineResponse.setField2(1);
    System.out.println("fallback exception ->  " + ex.getMessage() + " ex: " + ex);
    return mineResponse;
}

private MineResponse getMineFallback2(final String userId, final Integer contentId, Exception ex) {
    MineResponse mineResponse = new MineResponse();
    mineResponse.setField1(1);
    mineResponse.setField2(1);
    System.out.println("fallback2 exception ->  " + ex.getMessage() + " ex: " + ex);
    return mineResponse;
}
```
names at Annotation:
getMineSuccessName - has no configuration at all
backendMine - has following configuration in application.yml

application.yml
```YAML
resilience4j.retry:
    configs:
        default:
            maxRetryAttempts: 3
            waitDuration: 100
            retryExceptions:
                - org.springframework.web.client.HttpServerErrorException

            ignoreExceptions:
                - io.github.robwin.exception.BusinessException
    instances:
        backendA:
            baseConfig: default
        backendB:
            baseConfig: default
        backendMine:
            maxRetryAttempts: 3
            waitDuration: 100
            ignoreExceptions:
                - io.github.robwin.exception.BusinessException
```

When I run this application, everything seem works fine as expected. When other side is not responding and I call getMine() twise the /actuator/circuitbreakerevents shows "Whitelabel Error Page" with Invalid event. 

[...]

PR #940, opened 2020-03-24
Related to issue #939

rusyasoft:
- FAILURE_RATE_EXCEED exception has been triggered but it has no case in the CircuitBreakerEventDTOFactory
- This issue caused by relation between minimum call numbers and failureRateThreshold

### DIFF
--- a/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/monitoring/endpoint/CircuitBreakerEventDTOFactory.java
+++ b/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/monitoring/endpoint/CircuitBreakerEventDTOFactory.java
@@ -58,6 +58,10 @@ public class CircuitBreakerEventDTOFactory {
                 CircuitBreakerOnFailureRateExceededEvent onFailureRateExceededEvent = (CircuitBreakerOnFailureRateExceededEvent) event;
                 return newCircuitBreakerEventDTOBuilder(onFailureRateExceededEvent)
                     .build();
+            case SLOW_CALL_RATE_EXCEEDED:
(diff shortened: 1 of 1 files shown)


resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/monitoring/endpoint/CircuitBreakerEventDTOFactory.java
```diff
@@ -54,6 +54,10 @@ public static CircuitBreakerEventDTO createCircuitBreakerEventDTO(CircuitBreaker
                 CircuitBreakerOnCallNotPermittedEvent onCallNotPermittedEvent = (CircuitBreakerOnCallNotPermittedEvent) event;
                 return newCircuitBreakerEventDTOBuilder(onCallNotPermittedEvent)
                     .build();
+            case FAILURE_RATE_EXCEEDED:
```
dlsrb6342: I think we need to add `SLOW_CALL_RATE_EXCEEDED` type too. 
What do you think @RobWin ?

RobWin: Yes

dlsrb6342: @rusyasoft Please add `SLOW_CALL_RATE_EXCEEDED ` too.

rusyasoft: @dlsrb6342  added SLOW_CALL_RATE_EXCEEDED case

Issue #913, opened 2020-03-10
Actuator shutdown not working as expected


zhivkostanishev1:
Resilience4j version: 1.3.1

Java version: 8

Hello,

Hello,
As we are going forward with the integration of `resilience4j` in our project, we're facing a problem. We are not able to successfully shutdown our java process (JVM) using Spring's actuator shutdown endpoint whenever our system sends a request and utilizes some of the resilience4j's functionality (`Retry`, `CircuitBraker`, `TimeLimiter`, `BulkHead`).
This is also reproducible with the [demo](https://github.com/resilience4j/resilience4j-spring-boot2-demo) application. After calling http://localhost:9080/backendA/futureSuccess we're trying to close the application with http://localhost:9080/actuator/shutdown but only the spring application context is closed and the application doesn't exit at all.

dlsrb6342:
Thank you for reporting. 

I tested with our demo application. If I delete `TimeLimiter` and `ThreadPoolBulkhead` from `futureSuccess` method, application is closed. 

Need to find cause.

RobWin:
I guess it's the `ScheduledExecutorService` in the `TimeLimiterAspect`. Could it be that the shutdown hook does not work?

dlsrb6342:
Shutdown hook that we register for `ExecutorService` wasn't execute if calling  http://localhost:9080/backendA/futureSuccess. But if didn't call any api, shutdown hook executed.

zhivkostanishev1:
We discovered that when we shutdown `FixedThreadPoolBulkhead`'s `executorService` and
`TimeLimiterAspect`'s `timeLimiterExecutorService` the actuator's shutdown is performed as expected.
If shutdown is not performed on either of these some threads are left waiting
and the java process hangs.
The down side of our finding is that we use reflection, in `PreDestroy` method, to get to these executor services, because they are with private modifiers.

RobWin:
Yes indeed, we have to make the cleanup (close) method public. The jvm shutdown hook does not work in Spring as expected.

zhivkostanishev1:
It would be very nice if the cleanup methods in the core functionality are public, but also in the Spring/Spring Boot use-case if we have the following:

- property in the config (something like a boolean `shutdownOnSpringContextClose`)
- if the property is set to `true`, a bean with a `PreDestroy` or `ApplicationContextListener` is auto configured to close/cleanup/shutown all relevant `TimeLimiters`, `Bulkheads` etc.

[...]

PR #945, opened 2020-03-28
Issue #913: Fix bug not closed executorService

dlsrb6342:
I changed classes with ExecutorService to implement AutoClosable.
I tested with our spirng-boot2-demo and it was shutdowned properly.


RobWin:
Thx.
Does Spring close AutoClosable beans automatically?

dlsrb6342:
@RobWin 
Yes. When shutdown endpoint called, Bean's `close` method is automatically called.

yyankov:
Glad to see that this is getting fixed as my team bumped into an issue for Spring Boot microservice graceful shutdown which is caused by that executorService not being closed properly. Any plans on when this change can be merged?

PR #962, opened 2020-04-21
Issue:960: Fix Automatic-Module-Name: A Java module name follows the …

ekanthi:
Module names should follow same naming rules as Java packages. However, you should not use underscores (_) in module names (or package names, class names, method names, variable names etc.) from Java 9 and forward, because Java wants to use underscore as a reserved identifier in the future.

### DIFF
--- a/resilience4j-framework-common/build.gradle
+++ b/resilience4j-framework-common/build.gradle
@@ -7,4 +7,4 @@ dependencies {
     compile project(':resilience4j-timelimiter')
 }
 
-ext.moduleName = 'io.github.resilience4j.framework-common'
+ext.moduleName = 'io.github.resilience4j.framework.common'


resilience4j-framework-common/build.gradle
```diff
@@ -7,4 +7,4 @@ dependencies {
     compile project(':resilience4j-timelimiter')
 }
 
-ext.moduleName = 'io.github.resilience4j.framework-common'
\ No newline at end of file
+ext.moduleName = 'io.github.resilience4j.framework.common'
```
RobWin, line 10: Could you please change it into `io.github.resilience4j.frameworkcommon`

In the modules `spring-boot2` and `spring-boot` we also just removed `-`.

RobWin:
Never mind. I quickly fix it

PR #970, opened 2020-04-25
implemented SpelResolver for spring (#711)

### LINKED ISSUE #711: Support SpEL in annotations
How about support SpEL to name field in annotatations like [key field in spring Cacheable annotation](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/annotation/Cacheable.html#key--)?
Then we can use a method parameter as a circuitbreaker or others.

lespinsideg:
resolves #711 
- implemented `SpelResolver` which can resolve method SPeL expression as well as bean SPeL expression.
- added `SpelRootObject` for `#root.args[]` and `#root.methodName` expressions.


RobWin:
Thanks.
Can you tell us when the PR is ready to be revied?

lespinsideg:
@RobWin It's done already. Please review it :)

PR #981, opened 2020-05-04
Adding fairCallHandlingEnabled flag to BulkheadConfig. It indicates w…

### LINKED ISSUE #979: Semaphore bulkhead fair and unfair strategies
Resilience4j version: 1.4.0

Java version: 11.0.2

Hello,

A #906 pull request changed the default behaviour of Semaphore Bulkhead. In version 1.2.0 by default new calls were given the permission as long as the capacity threshold has not been exceeded. It means that even if one thread has been stucked, request processing has been continued. In version 1.4.0 that possibility was forbidden by changing the implementation of acquirePermission() method in order to use the so called fair strategy that ensures to keep order of incomming request. It means that even if only one call last very long, others cannot be performed and end up with BulkheadFullException. My usecase consists of set of available resources that is protected by Bulkhead. For every request a new thread is created as long as there is any resource availble in set mentioned above. 

My question is shouldn't be there any possibility to change the default fair strategy into nofair? Maybe I should use other component instead?

First of all, the approach with possible stategy change leaves the functional backward compatibility to version 1.2.0. Secondly, it is quite useful to let user to decide how he would like to handle his requests. For many situations there is no need to maintain the order of calls and for that cases the unfair strategy would be more appropriate.

ACzerw:
…hether FairSync or NonfairSync should be used in Semaphore.

 Closes #979

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/BulkheadConfig.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/BulkheadConfig.java
@@ -30,16 +30,19 @@ public class BulkheadConfig {
     public static final int DEFAULT_MAX_CONCURRENT_CALLS = 25;
     public static final Duration DEFAULT_MAX_WAIT_DURATION = Duration.ofSeconds(0);
     public static final boolean DEFAULT_WRITABLE_STACK_TRACE_ENABLED = true;
+    public static final boolean DEFAULT_FAIR_CALL_HANDLING_STRATEGY_ENABLED = true;
 
     private final int maxConcurrentCalls;
     private final Duration maxWaitDuration;
     private final boolean writableStackTraceEnabled;
+    private final boolean fairCallHandlingEnabled;
 
     private BulkheadConfig(int maxConcurrentCalls, Duration maxWaitDuration,
-        boolean writableStackTraceEnabled) {
+        boolean writableStackTraceEnabled, boolean fairCallHandlingEnabled) {
         this.maxConcurrentCalls = maxConcurrentCalls;
         this.maxWaitDuration = maxWaitDuration;
         this.writableStackTraceEnabled = writableStackTraceEnabled;
+        this.fairCallHandlingEnabled = fairCallHandlingEnabled;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


RobWin:
Yes, LGTM. Thank you. 

But I don't think that I understand this behavior:
> However, in case at least one call gets stuck or lasts very long no other newly submitted calls
 will be accepted regardless of whether {@link #maxConcurrentCalls} limit has been reached.

Is that really the case?

ACzerw:
Yesterday, I was trying to write such unit test but couldn't managed.
As for behavior, I had such case in project for my client. Implementation consists of Circuit Breaker change state listener which changes the maxConcurentCalls parameter value. Unfortunately, I used the changeConfig method of Bulkhead interface and such call stuck. Introducing new requests end up with BulkheadFullException due to implementation of FairSync in Semaphore:
`    static final class FairSync extends Sync {
        private static final long serialVersionUID = 2014338818796000944L;

        FairSync(int permits) {
            super(permits);
        }

        protected int tryAcquireShared(int acquires) {
            for (;;) {
                if (hasQueuedPredecessors())
                    return -1;
                int available = getState();
                int remaining = available - acquires;
                if (remaining < 0 ||
                    compareAndSetState(available, remaining))
                    return remaining;
            }
        }
    }`

storozhukBM:
@RobWin 

```
However, in case at least one call gets stuck or lasts very long no other newly submitted calls
will be accepted regardless of whether {@link #maxConcurrentCalls} limit has been reached.
```

Yes, you're right even with `non-fair` semaphore, you can't get more concurrent calls than  {@link #maxConcurrentCalls}. What `non-fair` semaphore does can be overly simplified and described as a replacement of a queue by a stack. So the threads waiting at the bottom of the stack can never get permission if there are always new threads joining to the top of a stack. 

So this part of a documentation should be fixed.

ACzerw:
Sorry, but I'm afraid I don't clearly understand what should be fixed. Maybe I didn't choose the best words to describe the situation but that's what I meant: if at least one call get stuck no new request will be accepted event if the maxConcurrentCalls has not been reached.
Example:
The maxConcurrentCalls parameter value is 10. One call has stuck so the bulkhead capacity is 9 now. However, if we try to invoke new request it would not be accepted and throw BulkheadFullException regardless of fact that there are 9 slots left.

If you could describe in more detail what should be fixed I would be very grateful.

RobWin:
I can't believe that a stuck call can block the whole Semaphore. A call which is running should not be  in the queue to get a permit anymore, because he already got a permit from the semaphore. A stuck call just prevents that the permit is released again. 
Or am I totally wrong? I don't think so.

storozhukBM:
@RobWin Yes, you are right. A stuck call can not block the whole Semaphore.

ACzerw:
I think that my case could be a bit more complex. If I'm not wrong I used Bulkhead without queue by not setting the maxWaitDuration parameter by my own. As a result, the default value should be used which is 0L and there shouldn't be any queue. A request changing state of my Circuit Breaker had to change the Bulkhead configuration as well using changeConfig() method and get stuck. Don't know why and how, that call became the predecessor for next requests according to FairSync.hasQueuedPredecessors() implementataion and as my queue capacity was 0 next request end up with BulkheadFullException thrown.

I fixed the description in documentation. Could you please take a look at it?

RobWin:
Yes I have a look at it. Thanks

PR #986, opened 2020-05-06
Support enabling both exponential backoff and randomized wait

### LINKED ISSUE #982: Support configuring exponential random backoff via Spring config
Resilience4j version: 1.4.0

Java version: All :)

Thanks for a great project - we've recently adopted this in our Spring Boot apps.

We do find it cumbersome to need to create RetryConfigCustomizers only to be able to configure exponential backoff with random jitter. We would otherwise (so far) be able to just provision everything in our Spring config.

Would it be possible to support enabling both via RetryConfigurationProperties?  Seems like maybe enabling both exponential backoff and randomized wait could create an IntervalFunction.ofExponentialRandomBackoff instead of throwing IllegalStateException ?

Thanks!

internetstaff:
This adds support for simultaneously enabling exponential backoff and randomized wait in RetryConfigurationProperties by reusing existing properties.

Fixes #982 

I'm not entirely satisfied with the unit test, but didn't want to test for e.g. not equal to the unrandomized value and then risk a random test failure.

Let me know if the slight cleanup I did is unwelcome - it helped me read it and Sonar was taunting me. :)

Thanks!

### DIFF
--- a/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/retry/configuration/RetryConfigurationProperties.java
+++ b/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/retry/configuration/RetryConfigurationProperties.java
@@ -121,12 +121,6 @@ public class RetryConfigurationProperties extends CommonProperties {
             return builder.build();
         }
 
-        if (properties.enableExponentialBackoff != null && properties.enableExponentialBackoff
-            && properties.enableRandomizedWait != null && properties.enableRandomizedWait) {
-            throw new IllegalStateException(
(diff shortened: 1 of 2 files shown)


Romeh:
@internetstaff  can u please check the test setup as the build is hanging for more than 60 minutes?

internetstaff:
@Romeh I'm not sure where to start - tests work locally except for one pre-existing failure. It doesn't seem obviously related to my change?

Romeh:
@internetstaff  i will trigger the build again and see how it will go

RobWin:
There seems to be only an issue with Azure

Issue #942, opened 2020-03-25
CB metric "resilience4j/circuitbreaker/calls" is broken for Stackdriver


YashchenkoN:
Resilience4j version: 1.3.1

Java version: 11

I use Micrometer Stackdriver Registry for publishing metrics. It returns an error while sending metrics to Stackdriver: 
```
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: One or more TimeSeries could not be written: Value type for metric custom.googleapis.com/resilience4j/circuitbreaker/calls must be DISTRIBUTION, but is DOUBLE.: timeSeries[37]
```

I investigated and found that `AbstractCircuitBreakerMetrics` creates 2 semantically different metrics using a single name which is "callsMetricName".

The first metric is timer:
```
        Timer successfulCalls = Timer.builder(names.getCallsMetricName())
            .description("Total number of successful calls")
            .tag(TagNames.NAME, circuitBreaker.getName())
            .tag(TagNames.KIND, KIND_SUCCESSFUL)
            .tags(customTags)
            .register(meterRegistry);
```
and the second metric is counter:
```
        Counter notPermittedCalls = Counter.builder(names.getCallsMetricName())
            .description("Total number of not permitted calls")
            .tag(TagNames.NAME, circuitBreaker.getName())
            .tag(TagNames.KIND, KIND_NOT_PERMITTED)
            .tags(customTags)
            .register(meterRegistry);
```

`Timer` and `Counter` are different types in Stackdriver and therefore Stackdriver returns an error, because the name of the metric is attached to a specific type.

Is it possible to change the name of the metric in order to avoid this problem?

RobWin:
Hi,
it would be a non-backward compatible change.
We should be able to manage it. We should introduce a new name in `MetricNames`: `getNotPermittedCallsMetricName` and set it by default to the same name like `getCallsMetricName`.
You could set `getNotPermittedCallsMetricName` to a new name in your application.

Would you like to raise a PR?

jkschneider:
@RobWin In one sense it would be a non-backwards compatible change. But I think in most cases, even when the monitoring system accepts these two types right now, it is sort of broken already.

Consider the case where you compute a dimensionally aggregable average over the timer series, but only selecting the metric name. It rolls up this counter into the numerator, but only the two timers in the denominator (in the best case). Yikes.

YashchenkoN:
@RobWin please review #990

PR #990, opened 2020-05-10
Introduce new CB metric "notPermittedCalls" 

YashchenkoN:
and set it to "calls" metric name by default in order to maintain backward compatibility

### DIFF
--- a/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/CircuitBreakerMetricNames.java
+++ b/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/CircuitBreakerMetricNames.java
@@ -7,6 +7,7 @@ public class CircuitBreakerMetricNames {
     private static final String DEFAULT_PREFIX = "resilience4j.circuitbreaker";
 
     public static final String DEFAULT_CIRCUIT_BREAKER_CALLS = DEFAULT_PREFIX + ".calls";
+    public static final String DEFAULT_CIRCUIT_BREAKER_NOT_PERMITTED_CALLS = DEFAULT_PREFIX + "not.permitted.calls";
     public static final String DEFAULT_CIRCUIT_BREAKER_STATE = DEFAULT_PREFIX + ".state";
     public static final String DEFAULT_CIRCUIT_BREAKER_BUFFERED_CALLS =
         DEFAULT_PREFIX + ".buffered.calls";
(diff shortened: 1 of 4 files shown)


YashchenkoN:
Related to #942

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/CircuitBreakerMetricNames.java
```diff
@@ -17,6 +17,7 @@
     public static final String DEFAULT_CIRCUIT_BREAKER_SLOW_CALL_RATE =
         DEFAULT_PREFIX + ".slow.call.rate";
     private String callsMetricName = DEFAULT_CIRCUIT_BREAKER_CALLS;
+    private String notPermittedCallsMetricName = callsMetricName;
```
jkschneider, line 20: Per my comment above, I'd strongly suggest we rename this default.

RobWin: Yes, I agree. People can still change it to the old value to be backwards compatible

YashchenkoN: Renamed

RobWin:
Seems some tests failed now

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/CircuitBreakerMetricNames.java
```diff
@@ -55,6 +57,16 @@ public String getCallsMetricName() {
         return callsMetricName;
     }
 
+    /**
+     * Returns the metric name for circuit breaker not permitted calls, defaults to {@value
+     * DEFAULT_CIRCUIT_BREAKER_CALLS}.
```
RobWin, line 62: Please change JavaDoc to `DEFAULT_CIRCUIT_BREAKER_NOT_PERMITTED_CALLS`

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/CircuitBreakerMetricNames.java
```diff
@@ -123,6 +135,18 @@ public Builder callsMetricName(String callsMetricName) {
             return this;
         }
 
+        /**
+         * Overrides the default metric name {@value CircuitBreakerMetricNames#DEFAULT_CIRCUIT_BREAKER_CALLS}
```
RobWin, line 139: Please adapt JavaDoc

PR #996, opened 2020-05-17
Configuration module

### LINKED ISSUE #957: Feature Request: File-based configuration for CircuitBreakerRegistry
Resilience4j version: resilience4j-circuitbreaker-1.3.1

Java version: 8

Is there a standard way to configure resilience4j's circuit breaker, similar to what is in [resilience4j-spring](https://github.com/resilience4j/resilience4j/tree/master/resilience4j-spring/) or [resilience4j-ratpack](https://github.com/resilience4j/resilience4j/tree/master/resilience4j-ratpack), but generically using a file rather than tying to a web container's configuration mechanism.

brandonheck:
Adds commons-configuration2 based configuration module for Bulkhead and CircuitBreaker. Fixes #957


brandonheck:
I'd be glad to build on this to add reading in configuration via commons-configuration2 for other modules, but I'd like to get any feedback that there may be on what I've been able to do so far. Please let me know if I should change this to a draft PR.

RobWin:
I think it would be nice if we could support a similar properties file format like in Spring Boot.
In Spring Boot you can configure instances, configure configs and even inherit from configs.

```
resilience4j.circuitbreaker. configs.default.slidingWindowSize=100
resilience4j.circuitbreaker. configs.default.permittedNumberOfCallsInHalfOpenState=10
resilience4j.circuitbreaker. configs.default.waitDurationInOpenState=10000

resilience4j.circuitbreaker.configs.someSharedConfig.slidingWindowSize=50
resilience4j.circuitbreaker.configs.someSharedConfig.permittedNumberOfCallsInHalfOpenState= 10
   
resilience4j.circuitbreaker.instances.backendA.baseConfig=default
resilience4j.circuitbreaker.instances.backendA.waitDurationInOpenState: 5000

resilience4j.circuitbreaker.instances.backendB.baseConfig=someShared
```

resilience4j-commons-configuration/src/main/java/io/github/resilience4j/configuration/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+        this.config = config.subset(circuitBreakerPrefix);
+    }
+
+    public CircuitBreakerConfig get(String name, final CircuitBreakerConfig defaultConfig) {
+        final Configuration namedConfig = config.subset(name);
+
+        CircuitBreakerConfig baseConfig =
+            Optional.ofNullable(namedConfig.getString("baseConfig")).map(b -> {
+                return this.get(b, defaultConfig);
+            }).orElse(defaultConfig);
```
brandonheck, line 35: There is minor concern to be aware of in the way that this is implemented since it is recursive, but I didn't want to over-engineer it if the maintainers are not concerned. Feedback on this welcome before I add similar logic to BulkheadConfiguration.

RobWin:
Please tell me when I can review the code again.

brandonheck:
Thanks for your time @RobWin. I believe that this is ready to be reviewed.

I am working on changes for RateLimiter, TimeLimiter, and Retry, but would like any feedback on BulkheadConfiguration and CircuitBreakerConfiguration that may need to be applied to any other changes (and I'm happy to submit those as a separate PR if preferred).

RobWin:
I think we should clarify the goal of the story/PR:
My understand is:

"As a User I want to configure and initialize a `CircuitBreakerRegistry` by using an external configuration like a properties or xml file. Reloading of file-based config files should be supported."

This story implies that the user also can configure and initialize CircuitBreaker instances. In Spring Boot we already have a format for property files and yaml files -> https://resilience4j.readme.io/docs/getting-started-3#configuration
We should really stick to that format so that there are not different formats per framework, e.g. Spring Boot, Micronaut, Ratpack and now Commons Configuration. We can even share code between frameworks. 
Please see [CircuitBreakerConfigurationProperties](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/configuration/CircuitBreakerConfigurationProperties.java).  The class already allows to c  configs and config inheritance.

In order to create and in initialize a `CircuitBreakerRegistry` correctly, you can have a look at [CircuitBreakerConfiguration](https://github.com/resilience4j/resilience4j/blob/378dd2b65889b83cda2a2b3d4203e86604b5044e/resilience4j-spring/src/main/java/io/github/resilience4j/circuitbreaker/configure/CircuitBreakerConfiguration.java#L78)

I think the task should be to be able to create a `CircuitBreakerConfigurationProperties` class via Commons Configuration 2. Then we can almost reuse the existing code.

RobWin:
Thank you for your effort 👍

RobWin:
Do you have time to finish this PR? :)

brandonheck:
I do plan to follow up but have not been able to prioritize the effort. Thank you for the reference for the properties. I am, unfortunately, having difficulty grepping how to use [CircuitBreakerConfigurationProperties](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/circuitbreaker/configuration/CircuitBreakerConfigurationProperties.java) based on how it is implemented in the resilience4j-spring module. Are there any resources that I have missed that may help me form a better understand of how this is intended to be used/reused?

Issue #991, opened 2020-05-11
CircuitBreaker transitionToForcedOpenState and transitionToDisabled still going to HALF_OPEN if previous state was OPEN


gdiddy21:
Resilience4j version: 1.3.1 and 1.4.0

Java version: 8

CircuitBreaker transitionToForcedOpenState() and transitionToDisabledState() still going to HALF_OPEN if previous state was OPEN

How to reproduce.

Create a Circuit Breaker with configuration like this:              
```
     CircuitBreakerConfig config = CircuitBreakerConfig.custom()
                .waitIntervalFunctionInOpenState(IntervalFunction.ofExponentialBackoff(30000, 1.5))
                .enableAutomaticTransitionFromOpenToHalfOpen()
                .slidingWindowType(CircuitBreakerConfig.SlidingWindowType.TIME_BASED)
                .build();

        CircuitBreakerRegistry circuitBreakerRegistry = CircuitBreakerRegistry.of(config);
        CircuitBreaker circuitBreaker = circuitBreakerRegistry.circuitBreaker("TestCB");
        
        circuitBreaker.transitionToOpenState();
        circuitBreaker.transitionToForcedOpenState();
        // or circuitBreaker.transitionToDisabledState();
```
after 30 seconds or so the circuitBreaker will transition to HALF_OPEN state.  I believe there is a ScheduledExecutorService that has this scheduled when set to OPEN State but it never gets cleared when transitioning to DISABLED or FORCED_OPEN.

RobWin:
Good findings. Thanks

razum90:
Hi, is this released yet? I'm still getting this issue using resilience4j-circuitbreaker 1.5.0.

PR #997, opened 2020-05-19
Issue #991 Cancel auto transition when trasition happend in Open state

dlsrb6342:


RobWin:
Thanks,
could you please add a test to verify the expected behavior?

dlsrb6342:
@RobWin 
Added tests. Please have a look :)

RobWin:
Is it possible to use the MockClock instead of Thread.sleep? Not sure if the scheduler supports it.

dlsrb6342:
@RobWin 

```java
@Test
public void shouldCancelAutoTransition() {
    ScheduledFuture<?> mockFuture = mock(ScheduledFuture.class);
    doReturn(mockFuture)
        .when(schedulerMock).schedule(any(Runnable.class), any(Long.class), any(TimeUnit.class));

    // Auto transition scheduled
    circuitBreaker.transitionToOpenState();

    // Auto transition should be canceled
    circuitBreaker.transitionToForcedOpenState();

    // Verify scheduled future is canceled
    then(mockFuture).should(times(1)).cancel(true);
}

```

mocked future and verified `future.cancel()` is called only one time. 
I think this is enough to test behavior. What do you think?

RobWin:
But with a MockClock wr can also mock the time and dont need thread sleep. Please have a look at the other CB state transition tests.

dlsrb6342:
I think `ScheduledExecutorService` doesn't support `MockClock`. 

Other auto transition tests are also just mocking `ScheduledExecutorService`.
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerAutoTransitionStateMachineTest.java

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
```diff
@@ -293,6 +291,9 @@ private void stateTransition(State newState,
         UnaryOperator<CircuitBreakerState> newStateGenerator) {
         CircuitBreakerState previousState = stateReference.getAndUpdate(currentState -> {
             StateTransition.transitionBetween(getName(), currentState.getState(), newState);
+            if (currentState instanceof OpenState) {
```
RobWin, line 294: I'm not happy that this state-specific code is in generic method of the CircuitBreakerStateMachine.
Maybe we can add a `postTransitionHook` or `preTransitionHook` to states which are invoked here in a generic form?

dlsrb6342: Yes right. that would be better. I'll add `preTransitionHook`.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
```diff
@@ -695,12 +706,23 @@ public CircuitBreakerMetrics getMetrics() {
...
+
         private void toHalfOpenState() {
             if (isOpen.compareAndSet(true, false)) {
                 transitionToHalfOpenState();
             }
         }
 
+        private void cancelAutomaticTransitionToHalfOpen() {
+            if (transitionToHalfOpenFuture != null) {
+                transitionToHalfOpenFuture.cancel(true);
```
RobWin, line 722: what happens if the future is already completed?

RobWin: does it really work? 
https://stackoverflow.com/questions/29181905/future-cancel-is-not-cancelling-the-scheduled-execution-of-scheduledexecutorse

dlsrb6342: Yes. Needed to check whether it is done or is cancelled already.
Thank you.

RobWin:
Thanks

Issue #935, opened 2020-03-23
Circuit breaker stuck in HALF_OPEN state


bankuruprasannakumar:
Resilience4j version:0.15.0

Java version:11

Circuit breaker was stuck in HALF_OPEN state on one of production node. Not able to replicate this. Is there any known scenario in which this can happen and can we replicate this? https://github.com/resilience4j/resilience4j/issues/903 seems to be speaking on the similar lines.

RobWin:
Hi, you are using an old version of Resiliencej4. There have been many bug fixes and improvements in the meantime.

Can you please provide more context. How do you use Resilience4j? Do you use a reactive operator?

KrnSaurabh:
@RobWin Do we need to investigate something here or any help needed ?

RobWin:
You could provide me some more context.  How do you use Resilience4j? Do you have some logs?
Do you catch exceptions in your code instead of propagating them?

KrnSaurabh:
Hi @RobWin , sorry I didn't face this issue or reported it so I have no more context. I just volunteered to help since it was tagged as `help wanted`.

RobWin:
Ah ok,
We can think of a better idea how to solve the challange that only a certain amount of calls should be possible in HALF_OPEN state.
The concept of AtomInteger does not seem to work perfectly, if some calls are permitted but never seem to finish either successfully or with a failure.

KrnSaurabh:
Hi @RobWin if a call never responds back with a success or failure how CircuitBreaker should treat those calls ? state transition takes place only based on aggregated outcome of the calls if completed successfully or failed or failed with ignored exception. Should CircuitBreaker ignore those calls who never responds with success or failure and release permission ?
Currently CB infinitely waits for those stuck call to complete and hence possible to get stuck in Half Open state, can we put any timeout or wait duration on Half Open state as well, similar to waitDurationInOpenState so that once the timeout is over we aggregate the calls outcome marking incomplete calls as success, failure or ignored and make a state transition to Open or Close state.

RobWin:
Thats a good question. Either we introduce a new configuration parameter that the test calls in HALF_OPEN state must happen in a certain period of time, or we have to introduce a timeout interval per test call and if the test calls times out, we have to increase the AtomicInteger again. But then we have to track all test calls. That sucks :(

KrnSaurabh:
[...]

PR #1007, opened 2020-05-25
Issue #935 WIP: Implementing token based permits in HalfOpen State

KrnSaurabh:
This PR is still under WIP as part of Issue #935 .

We have replaced existing acquire permission approach based on atomic Integer count to a token based permit mechanism. A `token` contains the timestamp when the token was issued to a permitted request. 

New config parameter has beed added to Circuit Breaker as `timeoutIntervalPerCallInHalfOpenState`, post the timeout duration requests will be considered unresponsive and will be marked as completed with Error.

When a request comes in HalfOpen state, `tryAcquirePermission`/`acquirePermission` returns a valid token to permitted request if `permittedNumberOfCallsInHalfOpenState` has not exceeded and stores the issued token into a HashSet. After completing the request, either successful completion or completion with error or ignoredException, request releases the issued token back and token is deleted from HashSet containing the used tokens marking the request appropriately `onError()` or `onSuccess()`.

When a new request comes in and `permittedNumberOfCallsInHalfOpenState` has exceeded, a `CallNotPermittedException` is thrown and no token is being issued to the request. On a request when `permittedNumberOfCallsInHalfOpenState` exceeds, we iterate through HashSet containing used token and remove all the token having `timeoutIntervalPerCallInHalfOpenState` exceeded, considering them unresponsive requests marking them failure by calling `onError()`.

This PR is WIP. Due to change in contract of `tryAcquirePermission()` and `acquirePermission`, many changes has to propagate down to the calling code keeping the new contract. Before I make those changes, I would like to get the implementation reviewed and agreed.
@RobWin Kindly review and provide your suggestions.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
@@ -942,6 +942,13 @@ public final class CircuitBreakerStateMachine implements CircuitBreaker {
             this.permittedNumberOfCalls = new AtomicInteger(permittedNumberOfCallsInHalfOpenState);
             this.isHalfOpen = new AtomicBoolean(true);
             this.attempts = attempts;
+
+            final long waitDurationInHalfOpenState = circuitBreakerConfig.getWaitDurationInHalfOpenState().toMillis();
+            if (waitDurationInHalfOpenState >= 1000) {
+                ScheduledExecutorService scheduledExecutorService = schedulerFactory.getScheduler();
+                scheduledExecutorService
+                    .schedule(this::toOpenState, waitDurationInHalfOpenState, TimeUnit.MILLISECONDS);
+            }
         }
 
         /**
@@ -967,6 +974,12 @@ public final class CircuitBreakerStateMachine implements CircuitBreaker {
             if (!tryAcquirePermission()) {
                 throw CallNotPermittedException
                     .createCallNotPermittedException(CircuitBreakerStateMachine.this);
+            }
+        }
+
+        private void toOpenState() {
+            if (isHalfOpen.compareAndSet(true, false)) {
+                transitionToOpenState();
             }
         }
 
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -38,6 +38,7 @@ public class CircuitBreakerConfig {
     public static final int DEFAULT_MINIMUM_NUMBER_OF_CALLS = 100;
     public static final int DEFAULT_SLIDING_WINDOW_SIZE = 100;
     public static final int DEFAULT_SLOW_CALL_DURATION_THRESHOLD = 60; // Seconds
+    public static final int DEFAULT_WAIT_DURATION_IN_HALF_OPEN_STATE = 0; // Seconds. It is an optional parameter
     public static final SlidingWindowType DEFAULT_SLIDING_WINDOW_TYPE = SlidingWindowType.COUNT_BASED;
     public static final boolean DEFAULT_WRITABLE_STACK_TRACE_ENABLED = true;
     private static final Predicate<Throwable> DEFAULT_RECORD_EXCEPTION_PREDICATE = throwable -> true;
@@ -64,6 +65,8 @@ public class CircuitBreakerConfig {
     private float slowCallRateThreshold = DEFAULT_SLOW_CALL_RATE_THRESHOLD;
     private Duration slowCallDurationThreshold = Duration
         .ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD);
+    private Duration waitDurationInHalfOpenState = Duration
+        .ofSeconds(DEFAULT_WAIT_DURATION_IN_HALF_OPEN_STATE);
 
 
     private CircuitBreakerConfig() {
@@ -159,6 +162,10 @@ public class CircuitBreakerConfig {
         return slowCallDurationThreshold;
     }
 
+    public Duration getWaitDurationInHalfOpenState() {
+        return waitDurationInHalfOpenState;
+    }
+
     public enum SlidingWindowType {
         TIME_BASED, COUNT_BASED
     }
@@ -189,6 +196,8 @@ public class CircuitBreakerConfig {
         private float slowCallRateThreshold = DEFAULT_SLOW_CALL_RATE_THRESHOLD;
         private Duration slowCallDurationThreshold = Duration
             .ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD);
+        private Duration waitDurationInHalfOpenState = Duration
+            .ofSeconds(DEFAULT_WAIT_DURATION_IN_HALF_OPEN_STATE);
 
 
         public Builder(CircuitBreakerConfig baseConfig) {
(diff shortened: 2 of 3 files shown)


Romeh:
@KrnSaurabh @RobWin 

Quickly checked the draft implementation , please find out my remarks for our own discussion and validation : 

I do not like we enforce timeout exception for unresponsive calls in half open state for many reasons : 

-  User may not like u are forcing specific timeout for their calls which may not fit their requirements and will like to take control of it
- If they will take control of it ,they may set it to very big value which will take us again to same situation we have now , tokens with very long timeout -> stuck in same state for long ime till the external service http server timeout the TCP connection if it beyond the configured server connection timeout of any
- creating on error time our exceptions will create error events recording in the circuit breaker which can lead to to false statistics and it can impact the state transition if they do not ignore that exception 

what we can do is to free the bucket hashset based into the waiting time u introduced without failing the request which will allow new requests to go through and it can lead to unblock the wait in half open state if required successful test calls are recorded and let the users handle the long running calls by themselves by using proper timeout management through time limiter for example or other ways of doing it .

What do u think ?

KrnSaurabh:
@Romeh we had our initial idea of this implementation very much similar to above.  For ref:
https://github.com/resilience4j/resilience4j/issues/935#issuecomment-611474815

Considering we avoid breaking the API and hide the token validation inside the state management of circuit breaker itself, my only concern here is, If we free up the used token buckets and increase atomic Integer count for the token that has expired `timeoutIntervalPerCallInHalfOpenState` , we end up allowing calls more than `permittedNumberOfCallsInHalfOpenState` . 

Actually with this approach, we will keep on allowing calls in HalfOpen state forever until atleast `minimumNumberOfCalls` in HalfOpen state responds with either success or failure and state change occurs to either closed or open. This is because when we free up token that has has expired `timeoutIntervalPerCallInHalfOpenState`, we have no way to know if the call corresponding to that token was success/failure or unresponsive considering we are not changing the contract. Hence we keep deleting the token from hash and incrementing atomic Integer count infinitely until state change occurs to closed or open.

Now in scenario where most of the calls to the end server results into long running/unresponsive call, we will still stuck in HalfOpen state, but off course we will keep allowing calls until it triggers a state change to open or closed.

 This is offcourse is an improvement since now we are giving a chance to application to recover itself by not blocking calls in case of unresponsive calls in HalfOpen state, but I am not sure if we solve the problem completely. 
Also It is possible that some user might want to consider long running/unresponsive calls as failure.

KrnSaurabh:
@Romeh @RobWin  Any conclusion on this or shall I go ahead and make changes as suggested by @Romeh  in the last comment ?

RobWin:
I have a look at it on Tuesday.  We have public holiday here

RobWin:
We always tried to reduce the amount of objects which are created by a CircuitBeraker per call in order to reduce GC pressure.
That's why we never used `Instant` or `Duration`, but `long` and `TimeUnit`.  We should try to avoid to create an `Optional`,  `Token` and an `Instant` per call now.

I also would like to void breaking the API. But how can we achieve this? We need to know which token was used properly in order to delete it from the HashMap.
@Romeh Do you have an idea?

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/Token.java
```diff
@@ -0,0 +1,15 @@
+package io.github.resilience4j.circuitbreaker.internal;
+
+import java.time.LocalDateTime;
+
+public class Token {
+    private LocalDateTime timestamp;
```
RobWin, line 6: I think we should use `System.nanoTime()` and  `long` as the creation timestamp in order to avoid too many objects.
It could be created inside of the constructor, or?

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -502,27 +508,27 @@ static CircuitBreaker of(String name,
...
+     * Returns Optional.empty() when the state is HALF_OPEN and the number of test calls has been reached. If
      * the state is HALF_OPEN, the number of allowed test calls is decreased. Important: Make sure
      * to call onSuccess or onError after the call is finished. If the call is cancelled before it
      * is invoked, you have to release the permission again.
      *
-     * @return {@code true} if a permission was acquired and {@code false} otherwise
+     * @return {@code Token} if a permission was acquired and {@code Optional.empty()} otherwise
      */
-    boolean tryAcquirePermission();
+    Optional<Token> tryAcquirePermission();
```
RobWin, line 520: Even if I like `Optional` we should try to avoid it in order to reduce GC pressure.

RobWin:
@dlsrb6342 @storozhukBM @hexmind What is your opinion?

dlsrb6342:
I agree with @Romeh (https://github.com/resilience4j/resilience4j/pull/1007#issuecomment-633718997) and @hexmind (https://github.com/resilience4j/resilience4j/issues/935#issuecomment-637702579). 
I think we just check `waitDurationInHalfOpenState` and timeout should be handled by users.

RobWin:
https://github.com/resilience4j/resilience4j/issues/935#issuecomment-638039472

KrnSaurabh:
@RobWin  @hexmind 
Refactored the code as per the comment https://github.com/resilience4j/resilience4j/issues/935#issuecomment-638904524.

Added an optional configuration parameter `waitDurationInHalfOpenState`  for circuit breaker. If the parameter is not set, circuit breaker would behave same as before, i.e circuit breaker would wait in Half Open state infinitely until long running/unresponsive calls completes with success or failure.

`waitDurationInHalfOpenState` if configured, controls how long the CircuitBreaker should stay in Half Open state before it switches to open.

`waitDurationInHalfOpenState` shall be configured to at least 1 second or else it throws error if it is set to less than 1 sec. This constrained shall be discussed and agreed upon.

RobWin:
Thanks, I have a look at it

RobWin:
The downside I still see is that, if you set waitDurationInHalfOpenState to a larger value like 5 Minutes and a request is stuck, the CB might be stuck multiple minutes until the timeout duration is elapsed.

But at least it's the most simple and non-intrusive solution.

RobWin:
I would be fine with this change. Could you please add some tests?

KrnSaurabh:
Sure. I will add unit test cases for same.

KrnSaurabh:
> The downside I still see is that, if you set waitDurationInHalfOpenState to a larger value like 5 Minutes and a request is stuck, the CB might be stuck a few minutes until the timeout duration is elapsed.

I totally agree with you @RobWin . Token based approach made much more sense to me but had challenges breaking the api. May be we can plan some long term solution later similar to those lines and can even accommodate api changes.

RobWin:
Let's do this change for 1.X.
Maybe we can include some new ideas for a 2.x CircuitBreaker.

RobWin:
I will merge it as soon as you have added some tests. 👍

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
```diff
@@ -942,6 +942,13 @@ public CircuitBreakerMetrics getMetrics() {
             this.permittedNumberOfCalls = new AtomicInteger(permittedNumberOfCallsInHalfOpenState);
             this.isHalfOpen = new AtomicBoolean(true);
             this.attempts = attempts;
+
+            final long waitDurationInHalfOpenState = circuitBreakerConfig.getWaitDurationInHalfOpenState().toMillis();
+            if (waitDurationInHalfOpenState >= 1000) {
+                ScheduledExecutorService scheduledExecutorService = schedulerFactory.getScheduler();
+                scheduledExecutorService
```
dlsrb6342, line 949: I think this `ScheduledFuture` should be canceled if Circuit transit to other states. 
https://github.com/resilience4j/resilience4j/issues/991.

RobWin: Yes

RobWin: Yes, we introduces a preTransitionHook -> https://github.com/resilience4j/resilience4j/commit/378dd2b65889b83cda2a2b3d4203e86604b5044e

KrnSaurabh:
Kindly review #1042 . I accidentally deleted the remote branch and I have created a new PR #1042  incorporating all the review comments on this PR. Kindly review the same.
Sorry for the inconvenience caused.

PR #1024, opened 2020-06-08
Add predicate to record a failure depending on the result

albertllousas:
This PR adds a CircuitBreaker record a failure predicate for a generic results not just exceptions.

The drawback with the current implementation of the circuit breaker is that it works with the assumption that a function fails when you have an exception, but the reality is that you can express a failure in several ways, for example using monads like Either or custom Result classes, where you carry the exception or whatever you consider an error in the result of the function itself.

### DIFF
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,6 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Project copy" />
   </state>
 </component>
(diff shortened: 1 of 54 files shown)


RobWin:
We always tried to reduce the amount of objects which are created by a CircuitBeraker per call in order to reduce GC pressure.
That's why we never used Instant or Duration, but long and TimeUnit. I think we should try to avoid to create an Optional per call.

RobWin:
We can us method overloading to provide a second `onSuccess` method which takes a result as a parameter only when a result is available.

RobWin:
Do you still have time to enhance this PR?

RobWin:
Any progress planned? :)

RobWin:
@mkopylec Would you like to take over?

mkopylec:
No time for that, but if I find some, I'll try to do it.

RobWin:
ok, I do it

RobWin:
See https://github.com/resilience4j/resilience4j/pull/1232

Issue #1018, opened 2020-06-04
[Question] Spring Boot Servlet Implementation with @Timelimiter 


SimonScholz:
Resilience4j version: 1.5.0

Java version: 1.8

Hi,

we still use the Servlet part and not yet the Webflux part of Spring Boot, but still want to benefit from the `@TimeLimiter` annotation.
The code of the `TimeLimiterAspect` enforces to use a `CompletionStage` as return type in case there is no suitable `TimeLimiterAspectExt` available.
Since we're using the servlet implementation methods, which are annotated with the `@TimeLimiter` annotation used to be blocking and therefore simply returned the desired type directly.

Now to archive the same goal we simply return a `CompletableFuture<MyType>` and the caller of this method calls `get()` on the returned `CompletableFuture<MyType>`.
Would this be the right approach? Or to you see any undesired side effects when using the `@TimeLimiter` annotation like this?
I am asking this because sometimes we face some delays when calling other rest APIs by using Timelimiter, and I am afraid that the timelimiter threading overhead causes these delays?

Unfortunately it is not really easy to make blocking code non blocking.

Thanks in advance for an answer to all these questions ;-)

RobWin:
Hi, do you really have to use annotations? I would suggest to use `Future` and decorate them with a TimeLimiter. It's still blocking code, but I understood that this is not an issue?

RobWin:
Did it help you?

SimonScholz:
You are talking about the `executeFutureSupplier` method on the Timelimiter, right?

```
// The blocking variant which is basically future.get(timeoutDuration, MILLISECONDS)
String result = timeLimiter.executeFutureSupplier(
  () -> CompletableFuture.supplyAsync(() -> helloWorldService::sayHelloWorld));
```

We use this annotation in conjuction with `@CircuitBreaker` and `@BulkHead` and using a different approach for the `Timelimiter` feels strange to me. 
Besides that I am not sure whether the `TimeLimiter` will also propagate Metrics or if I have to wire up the `M̀eterRegistry`? Does it?
Maybe we can adjust the `@Timelimiter` annotation with a property to specify that the `executeFutureSupplier` should be used? What do you think about this?
I can also try to provide a PR for this in case you say that this might make sense.

RobWin:
The Aspect extension mechanism allows you to add support for any return type to your application.
You can implement a `FutureTimeLimiterAspectExt` and register it as a `@Bean`.

[...]

PR #1029, opened 2020-06-17
Also make use of executeFutureSupplier method with TimeLimiter annotation

SimonScholz:
Hi,

it would be nice to also make use of the `executeFutureSupplier` method by using the `Timelimiter` annotation. Also see https://github.com/resilience4j/resilience4j/issues/1018

This is just a draft and I'll add documentation for the changed in case you like them.

### DIFF
--- a/resilience4j-spring/src/main/java/io/github/resilience4j/timelimiter/configure/TimeLimiterAspect.java
+++ b/resilience4j-spring/src/main/java/io/github/resilience4j/timelimiter/configure/TimeLimiterAspect.java
@@ -86,18 +86,19 @@ public class TimeLimiterAspect implements Ordered, AutoCloseable {
         Class<?> returnType = method.getReturnType();
 
         String fallbackMethodValue = spelResolver.resolve(method, proceedingJoinPoint.getArgs(), timeLimiterAnnotation.fallbackMethod());
+        boolean blocking = timeLimiterAnnotation.blocking();
         if (StringUtils.isEmpty(fallbackMethodValue)) {
-            return proceed(proceedingJoinPoint, methodName, timeLimiter, returnType);
+            return proceed(proceedingJoinPoint, methodName, timeLimiter, returnType, blocking);
         }
         FallbackMethod fallbackMethod = FallbackMethod
             .create(fallbackMethodValue, method,
                 proceedingJoinPoint.getArgs(), proceedingJoinPoint.getTarget());
         return fallbackDecorators.decorate(fallbackMethod,
-            () -> proceed(proceedingJoinPoint, methodName, timeLimiter, returnType)).apply();
+            () -> proceed(proceedingJoinPoint, methodName, timeLimiter, returnType, blocking)).apply();
     }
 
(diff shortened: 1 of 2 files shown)


resilience4j-spring/src/main/java/io/github/resilience4j/timelimiter/configure/TimeLimiterAspect.java
```diff
@@ -139,8 +143,17 @@ private static TimeLimiter getTimeLimiterAnnotation(ProceedingJoinPoint proceedi
         }
     }
 
-    private static Object handleJoinPointCompletableFuture(
-            ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.timelimiter.TimeLimiter timeLimiter) throws Throwable {
+    private static Object handleJoinPointFuture(
+            ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.timelimiter.TimeLimiter timeLimiter, boolean blocking) throws Throwable {
+        if(blocking) {
+            return timeLimiter.executeFutureSupplier(() -> {
```
RobWin, line 149: I think this doesn't work.
The method which is intercepted returns a `Future`, but the Aspect returns a value. 
I think the Aspect should/must return the same type like the intercepted method.

SimonScholz: Oh yes. My bad. I just updated the PR accordingly, so that it is similar to https://resilience4j.readme.io/docs/timeout#decorate-and-execute-a-functional-interface

resilience4j-spring/src/main/java/io/github/resilience4j/timelimiter/configure/TimeLimiterAspect.java
```diff
@@ -139,8 +140,19 @@ private static TimeLimiter getTimeLimiterAnnotation(ProceedingJoinPoint proceedi
         }
     }
 
-    private static Object handleJoinPointCompletableFuture(
-            ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.timelimiter.TimeLimiter timeLimiter) throws Throwable {
+    private static Object handleJoinPointFuture(
+            ProceedingJoinPoint proceedingJoinPoint, io.github.resilience4j.timelimiter.TimeLimiter timeLimiter, boolean blocking) throws Throwable {
+        if(blocking) {
+            return timeLimiter.executeFutureSupplier(() ->
+                    CompletableFuture.supplyAsync(() -> {
```
RobWin, line 147: The intercepted method must return a Future. That means the task is already executed in an ExecutorService.
The developer should be able use the ExecutorService of his choice.

The TimeLimiterAspect must not execute a method (task) implictly in it's own ExecutorService which is not configurable. The ThreadPoolBulkhead should be used to run tasks asynchronously.

SimonScholz:
Closing this PR because of making use of a different solution.
See https://github.com/resilience4j/resilience4j/issues/1018

PR #1033, opened 2020-06-19
TimeLimiters decorateFutureSupplier method should also propagate TimeLimiters name on timeout

SimonScholz:
After a nice discussion with @RobWin in this issue: https://github.com/resilience4j/resilience4j/issues/1018, we decided to go with the decorateFutureSupplier method of a TimeLimiter, but we just realized that on a timeout the name of the timelimiter is not published. So unfortunately we get Errors like this in our error log:

`TimeoutException: null`

Formerly the error looked similar to this:

`TimeoutException: TimeLimiter 'FetchData' recorded a timeout exception.`

which is much more pleasant.

This PR should fix this issue.

### DIFF
--- a/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
+++ b/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
@@ -189,7 +189,7 @@ public class TimeLimiterImpl implements TimeLimiter {
         }
     }
 
-    private static TimeoutException createdTimeoutExceptionWithName(String name) {
+    static TimeoutException createdTimeoutExceptionWithName(String name) {
         return new TimeoutException(String.format("TimeLimiter '%s' recorded a timeout exception." , name));
     }
 }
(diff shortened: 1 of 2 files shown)


resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java
```diff
@@ -48,11 +48,12 @@ public TimeLimiterImpl(String name, TimeLimiterConfig timeLimiterConfig,
                 onSuccess();
                 return result;
             } catch (TimeoutException e) {
-                onError(e);
+                TimeoutException timeoutException = createdTimeoutExceptionWithName(name);
```
RobWin, line 51: Should we copy the source stacktrace of the original TimeoutException?

SimonScholz: Good point. 
Just adjusted it.

RobWin:
Thx. Could you please modify the [TimeLimiterTest](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-timelimiter/src/test/java/io/github/resilience4j/timelimiter/TimeLimiterTest.java#L28-L45)to verify it?

SimonScholz:
> Thx. Could you please modify the [TimeLimiterTest](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-timelimiter/src/test/java/io/github/resilience4j/timelimiter/TimeLimiterTest.java#L28-L45)to verify it?

I also moved the Test into the internal package to make use of the `createdTimeoutExceptionWithName` inside the Test.
In case you do not like this approach I can also change the test to just verify the timelimiter name is mentioned in the Exception's message. It's up to you what you prefer.

Issue #866, opened 2020-02-13
About component switches


WigeHuang:
Does it support configuration of component opening and closing in yml files?

RobWin:
Do you mean at runtime? You wan to open and close a CircuitBreaker manually at runtime?

WigeHuang:
Yeah, haven’t watched it for a long time, thank you

PR #1038, opened 2020-06-22
Issue #866 add spring actuator rest endpoints for force open , disable, close specifc circuitbreaker

Romeh:
This PR is the one step of the requirements of enabling changing the state at runtime for specific circuit breaker , first part in that PR is provide actuator endpoint to disable , force open , and close specific circuit breaker instance 

Second PR is to provide state config property in circuit breaker configuration so it can be done via spring cloud config as well at runtime


RobWin:
LGTM

PR #1047, opened 2020-06-26
InstanceProperties setter should return themselves for chaining

dlsrb6342:
* Some `InstanceProperties`'s setters doesn't return themselves.


RobWin:
Can Spring handle this fluent builder pattern for it's config files?

dlsrb6342:
Yes. Other setters return themselves. 
I only changed setters that doesn't return themselves unlike other setters.

Issue #1062, opened 2020-07-02
CallNotPermittedException should contain circuit breaker name (and possibly other information)


rLitto:
Resilience4j version: <= 1.5.0 

Java version: 1.11

CallNotPermittedException should contain (at least) the name of the CircuitBreaker that triggered it.

Scenarios: nested CircuitBreakers
Eg:
CircuitBreaker around (Project Reactor custom implementation) of Redis cache that in case of RedisConnectionException or CallNotPermittedException will call a backend service with another CircuitBreaker. The CallNotPermittedException will trigger a backend call only if it is generated by the `redis` circuitbreaker

RobWin:
Would you like to create a PR?

mehwhatever:
I can see this is already being done, https://github.com/resilience4j/resilience4j/blob/e0221c7f8897e6e2e4de6726dcc4d39bea415997/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java#L41-L43

Am i missing something?

mehwhatever:
Ah, do you mean that it should have a member variable that indicates the CircuitBreaker name?

RobWin:
@rLitto I guess that was you request, or? A member variable of the name?

mehwhatever:
Raised a PR to pass the causing CircuitBreaker instead of just the name, in case the user wants to check other things too.

RobWin:
Thx

PR #1069, opened 2020-07-08
Issue #1062: Adding CircuitBreaker to CallNotPermittedException for inspection

mehwhatever:
Issue #1062: Adding CircuitBreaker to CallNotPermittedException for inspection

Raised a PR for adding the circuit breaker to the exception, otherwise one would have to parse the exception message which seems hacky. Instead of adding only the name passed the entire CircuitBreaker for any other inspection people might want to do.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
@@ -24,8 +24,11 @@ package io.github.resilience4j.circuitbreaker;
  */
 public class CallNotPermittedException extends RuntimeException {
 
-    private CallNotPermittedException(String message, boolean writableStackTrace) {
+    private final transient String causingCircuitBreakerName;
+
+    private CallNotPermittedException(CircuitBreaker circuitBreaker, String message, boolean writableStackTrace) {
         super(message, null, false, writableStackTrace);
+        this.causingCircuitBreakerName = circuitBreaker.getName();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CallNotPermittedException.java
```diff
@@ -24,8 +24,11 @@
  */
 public class CallNotPermittedException extends RuntimeException {
 
-    private CallNotPermittedException(String message, boolean writableStackTrace) {
+    private final transient CircuitBreaker circuitBreaker;
+
+    private CallNotPermittedException(CircuitBreaker circuitBreaker, String message, boolean writableStackTrace) {
         super(message, null, false, writableStackTrace);
+        this.circuitBreaker = circuitBreaker;
```
RobWin, line 31: I'm not sure if we should have a reference in every exception to the CircuitBreaker instance.
I would prefer if we just add the name to the CallNotPermittedException as a field.

mehwhatever: I also felt a bit uncomfortable doing this, but also didn't want future requests asking to add other fields in the exception. What's the major concern with adding a reference?

mehwhatever: Any update? Can change to just having string if there are major concerns.

RobWin: Yes, please change to to a string.

mehwhatever: Done.

RobWin:
thx

RobWin:
Do we need this for Retry, Bulkhead and RateLimiter as well?

PR #1104, opened 2020-07-29
remove annotation mapper

pollend:
Micronaut can setup these annotation at compile time there isn't really a need for these annotation mappers.


RobWin:
Hi, sry I was on vacation

Issue #1118, opened 2020-08-11
update resilience-feign documents


vsanna:
Resilience4j version: 1.5.0

Java version: any version

CircuitBreakerOpenException was deprecated and removed. 
So I suggest that we replace CircuitBreakerOpenException with CallNotPermittedException in its README.adoc so that new developers won't be confused.

I've created an example PR. I'd like you to check when you have a chance 🙇

vsanna:
merged

PR #1119, opened 2020-08-11
replace CircuitBreakerOpenException with CallNotPermittedException

vsanna:
related issue: #1118 

## why
according to this change https://github.com/resilience4j/resilience4j/pull/512/files, 
CircuitBreakerOpenException was deprecated and removed.
Instead of that, now we are using CallNotPermittedException to show circuitbreaker is open(or half-open).

## what 
replaced CircuitBreakerOpenException with CallNotPermittedException in README.adoc


RobWin:
Thx

Issue #1115, opened 2020-08-09
ConcurrentModificationException warning log while consuming events


bhochhi:
Resilience4j version: 1.4.0/1.5.0

Java version: 1.8.0

Following is the warning log I am getting in my prod

```
java.util.ConcurrentModificationException
	at java.util.ArrayList.forEach(ArrayList.java:1260)
	at io.github.resilience4j.core.EventProcessor.processEvent(EventProcessor.java:65)
	at io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine$CircuitBreakerEventProcessor.consumeEvent(CircuitBreakerStateMachine.java:520)
	at io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine.publishEventIfPossible(CircuitBreakerStateMachine.java:343)
	at io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine.publishSuccessEvent(CircuitBreakerStateMachine.java:379)
	at io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachine.onSuccess(CircuitBreakerStateMachine.java:229)
	at io.github.resilience4j.circuitbreaker.CircuitBreaker.lambda$decorateSupplier$4(CircuitBreaker.java:200)
	at io.github.resilience4j.retry.Retry.lambda$decorateSupplier$2(Retry.java:213)
	at io.github.resilience4j.decorators.Decorators$DecorateSupplier.get(Decorators.java:166)
```

I am using retry and circuit breaker as follows
```
    public Response makeCallWithResilience(String instanceName){ 

            Retry retry = registryService.getRetryRegistry().retry(instanceName);
            CircuitBreaker circuitBreaker = registryService.getCircuitBreakerRegistry.circuitBreaker(instanceName);

            //circuit breaker is chained before retry so that it counts retries into its calculation.
            return Decorators.ofSupplier(() -> callDownstream(url))
                    .withCircuitBreaker(circuitBreaker)
                    .withRetry(retry)
                    .get();
   }
```
registryService is the singleton service that  creates retryRegistry and circuitBreakerRegistry and also initializes retry and circuitBreaker instances on application start. two more lines are subscribing the  `onRetry`, `onRetryError` and `onFailureRateExceeded` just to log the message as follows:

```
public void initOnApplicationLoad(String instanceName){
      this.retryRegistry = RetryRegistry.of(retryConfig);
[...]

PR #1134, opened 2020-08-20
Issue #1115: Fixed a bug which causes a ConcurrentModificationExcepti…

RobWin:
…on when a new event consumer is registered while a event is processed.


bhochhi:
@RobWin Is this fix planned for 1.6.0 release? I don't know much on internals but I am hoping this should not have any impact on circuit breaker metrics right?

Issue #1125, opened 2020-08-18
Flaky multi-threading behavior with BulkheadOperator inside Mono.zip 


kgyokov:
Resilience4j version: 1.5.0

Java version: 11

Hello!

When I run the code below, it only executes the first 24 iterations and then hangs. The issue seems to be triggered by a combination of factors:
1. Two parallel calls inside a _Mono.zip_,  each call on its own thread.
2. At least one of the calls returns Mono.empty
3. _.transformDeferred(BulkheadOperator.of(bulkhead))_ is placed _before .subscribeOn(scheduler)_

Maybe point 3 is not good practice, but it would be best if it doesn't cause the app to hang. 

```

import java.time.Duration;

import io.github.resilience4j.bulkhead.Bulkhead;
import io.github.resilience4j.bulkhead.BulkheadConfig;
import io.github.resilience4j.reactor.bulkhead.operator.BulkheadOperator;

import reactor.core.publisher.Mono;
import reactor.core.scheduler.Scheduler;
import reactor.core.scheduler.Schedulers;

public class TestBulkhead {

  private static Bulkhead bulkhead;
  private static Scheduler scheduler =
      Schedulers.newBoundedElastic(20, 10000, "TestScheduler");

  public static void main(String[] args) {
    BulkheadConfig config = BulkheadConfig.custom()
        .maxConcurrentCalls(15)
        .maxWaitDuration(Duration.ZERO)
        .build();

    bulkhead = Bulkhead.of("testBulkhead", config);

    for (int i = 0; i < 500; i++) {
      make2CallsInParallel().block();
      System.out.println(i);
    }

    scheduler.dispose();
  }

  private static Mono<Integer> make2CallsInParallel() {
    return Mono.zip(reactiveGetInteger(), reactiveGetInteger(), (a, b) -> a + b);
  }

  private static Mono<Integer> reactiveGetInteger() {
    return blockingGetInteger()
        .transformDeferred(BulkheadOperator.of(bulkhead)) // Putting this BEFORE the scheduler triggers the issue
        .subscribeOn(scheduler);
  }

  // imagine this is a non-reactive IO code that sometimes returns Mono.empty()  
  private static Mono<Integer> blockingGetInteger() {
    return Mono.empty();
  }
}

```

RobWin:
Do you know why the code hangs?
Can you have a look at the thread dump?

kgyokov:
[...]

PR #1137, opened 2020-08-24
Issue #1125: Fix for thread leak [TENTATIVE/PROTOTYPE]

kgyokov:
A prototype of a  solution for the issue.


kgyokov:
I just want to make sure people are OK with the overall approach

PR #1140, opened 2020-08-25
Micronaut Support for Resilience4j

RobWin:


pollend:
looks like I need to spend additional time on some additional test cases.

Romeh:
closing it as it will be included in https://github.com/resilience4j/resilience4j/pull/1179

PR #1141, opened 2020-08-25
Phase-2 Remove vavr dependecy from Bulkhead and RateLimiter

KrnSaurabh:
Removed vavr dependency from Bulkhead and RateLimiter modules and back ported contract breaking changes.
@RobWin  Kindly review.

### DIFF
--- a/resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java
+++ b/resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java
@@ -3,6 +3,7 @@ package io.github.resilience4j.decorators;
 import io.github.resilience4j.bulkhead.Bulkhead;
 import io.github.resilience4j.bulkhead.BulkheadFullException;
 import io.github.resilience4j.bulkhead.ThreadPoolBulkhead;
+import io.github.resilience4j.bulkhead.VavrBulkhead;
 import io.github.resilience4j.cache.Cache;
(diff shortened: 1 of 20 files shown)


resilience4j-all/src/main/java/io/github/resilience4j/decorators/Decorators.java
```diff
@@ -382,7 +384,7 @@ private DecorateCheckedSupplier(CheckedFunction0<T> supplier) {
         }
 
         public DecorateCheckedSupplier<T> withRateLimiter(RateLimiter rateLimiter, int permits) {
-            supplier = RateLimiter.decorateCheckedSupplier(rateLimiter, permits, supplier);
+            supplier = VavrRateLimiter.decorateCheckedSupplier(rateLimiter, permits, supplier);
```
KrnSaurabh, line 387: @RobWin I have currently added dependency `resilience4j-vavr` to `resilience4j-all` module. Do you think we should create `VavrDecorators` inside `resilience4j-vavr` module and move all vavr related decorators in that class and remove dependency `resilience4j-vavr` from `resilience4j-all` module.

RobWin: Yes, that makes sense

KrnSaurabh: sure. I will do it next PR.

Issue #936, opened 2020-03-23
Passing Threadlocal to the fallbackMethod


wundorki:
Hello!
Is there any way the Threadlocal can be passed to the fallbackMethod as well?
I have tried using a ContextPropagator, but the clear method of the propagator is called before the fallbackMethod. Is there any other way of doing this?

RobWin:
Hi,

I guess there is no option. Might be a good new enhancement / PR.

esteve8:
@RobWin i have a question for you, maybe you can help me.

I have an spring rest controller with differents end points, each endpoint returns differents types of objests. My config is :

@CircuitBreaker(name = "default", fallbackMethod = "fallback")

End points : 

@GetMapping(value = "/test1")
    public Object1 test1() {

        Optional<Object1> object1= Object1Dao.findById(1);
        return object1.orElse(null);
 }

 @GetMapping(value = "/test2")
    public Object2 test2() {

        Optional<Object2> object2= Object2Dao.findById(1);
        return object2.orElse(null);
    }

   And the fallback:

    public String fallback(Throwable ex) {
        return "Recovered with fallback: " + ex.getMessage();
    }

The problem is that i got an error that the return type of the fallback has to be the same that the object type that returns the end point that throws the error.

Do you know how can i fix this? i appreciate your help so much.

Thanks so much.

RobWin:
@esteve8 You can define two `fallback` methods with different return types.

esteve8:
ohh thats so great! congratulations for the effort and thanks @RobWin

kumarsen26:
any update on this?

PR #1143, opened 2020-08-27
Issue #936: Reusing MDC in FallbackMethod

razum90:
Pull request for issue #936.

Currently when the `invoke()` method in `FallbackMethod` is called the MDC is lost (that was present when calling the original method). Expected behaviour, at least from my personal point of view, is that the same MDC present when calling the original method, should also be present when the fallback method is invoked. One use-case for this is to be able to track the call-chain in a service through logs, including the fallback method invocation, using a request-id for example present in the MDC.

This is the behaviour I've tried to implement in this pull request. I also added the `withMdcContextMap()` method inside the `FallbackMethod` class, to be able to control what MDC should be used when invoking, in case, for any reason, one would not want the current MDC (present when calling the original method) to be used. But since that's not used, and I don't really see a use-case for it, maybe it should be removed. Let me know what you think.

### DIFF
--- a/resilience4j-spring/src/main/java/io/github/resilience4j/fallback/FallbackMethod.java
+++ b/resilience4j-spring/src/main/java/io/github/resilience4j/fallback/FallbackMethod.java
@@ -16,15 +16,14 @@
 package io.github.resilience4j.fallback;
 
 import io.github.resilience4j.core.lang.Nullable;
+import org.slf4j.MDC;
 import org.springframework.util.ConcurrentReferenceHashMap;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Reflection utility for invoking a fallback method. Fallback method should have same return type
@@ -49,9 +48,11 @@ public class FallbackMethod {
     private final Object[] args;
     private final Object target;
     private final Class<?> returnType;
+    private Map<String, String> mdcContextMap;
 
     /**
      * create a fallbackMethod method.
+     * stores the current MDC context to be used when invoking the fallback method.
      *
      * @param fallbackMethods          configured and found fallback methods for this invocation
      * @param originalMethodReturnType the return type of the original source method
@@ -66,6 +67,7 @@ public class FallbackMethod {
         this.args = args;
         this.target = target;
         this.returnType = originalMethodReturnType;
+        this.mdcContextMap = Optional.ofNullable(MDC.getCopyOfContextMap()).orElse(Collections.emptyMap());
     }
 
     /**
@@ -93,6 +95,15 @@ public class FallbackMethod {
                 methodMeta.returnType, methodMeta.targetClass, methodMeta.fallbackMethodName,
                 StringUtils.arrayToDelimitedString(methodMeta.params, ","), Throwable.class));
         }
+    }
+
+    /**
+     * @param mdcContextMap the mdc context map to be used when invoking the fallback method
+     * @return FallbackMethod itself
+     */
+    public FallbackMethod withMdcContextMap(Map<String, String> mdcContextMap) {
+        this.mdcContextMap = mdcContextMap;
+        return this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


razum90:
The build seems to be failing in master, for the same reason this build is failing, by the way.

RobWin:
I wonder if it works. The fallback method is currently invoked in the same Thread like the protected method.
Why is the MDC context lost from the Threadlocal.
Do you also use a ThreadPoolBulkhead on your method as well?

razum90:
@RobWin Yes, we are.

This is what it looks like:

```
@Bulkhead(name = "name", type = Bulkhead.Type.THREADPOOL)
@TimeLimiter(name = "name")
@CircuitBreaker(name = "name", fallbackMethod = "fallback")
public CompletableFuture<String> asyncCheck(String arg) {
    return CompletableFuture.supplyAsync(MDCUtil.withMdc(() -> client.execute(arg)));
}

public CompletableFuture<String> fallback(String arg, Exception exception) {
    log.error("...", exception);
    return CompletableFuture.completedFuture("response");
}
```

RobWin:
Sry, your code is wrong. Bulkhead already executes your method code inside a thread. You should not call `supplyAsync` and spawn another thread. Just return `CompletableFuture.completedFuture()` like in your fallback method.

razum90:
Ah, yep. Thanks, will change according to your suggestion. However should not change the fact that the MDC is empty in the fallback method right?

RobWin:
It should, because the fallback should be invoked in the same thread than the original method.

razum90:
Did you try it? I did, and the MDC was still empty for me. But maybe I missed something.

RobWin:
I will retest it, but today I'm on vacation. But if you can quickly test it, I'm happy to read your feedback.

RobWin:
What is your MDC util doing? Is the MDC also in the asyncCheck method empty? Did you configure a ContextPropagator?

razum90:
I did try it again, and the MDC is still empty in the fallback. I can see the fallback is not executed in the same thread as asyncCheck:

asyncCheck thread name: bulkhead-name-2
fallback thread name: pool-3-thread-1

razum90:
I'm not calling the MDC util method anymore after I removed the supplyAsync call according to your suggestions. The MDC is populated in the asyncCheck method. Yes, I had a look and we are using a ContextPropagator, looks like this:

`
public class DefaultContextPropagator implements ContextPropagator<Map> {

    @Override
    public Supplier<Optional<Map>> retrieve() {
        return () -> (Optional) Optional.ofNullable(MDC.getCopyOfContextMap());
    }

    @Override
    public Consumer<Optional<Map>> copy() {
        return (t) -> t.ifPresent(e -> {
            if (MDC.getCopyOfContextMap() != null) {
                MDC.clear();
            }
            MDC.setContextMap(e);
        });
    }

    @Override
    public Consumer<Optional<Map>> clear() {
        return (t) -> {
            if (MDC.getCopyOfContextMap() != null) {
                MDC.clear();
            }
        };
    }
}
`

If I remove the usage of that, the MDC is also empty in the asyncCheck method.

RobWin:
We have to debug https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring/src/main/java/io/github/resilience4j/fallback/CompletionStageFallbackDecorator.java

razum90:
@RobWin Seems to be the TimeLimiter spawning a new thread for the fallback to run in, at https://github.com/resilience4j/resilience4j/blob/master/resilience4j-timelimiter/src/main/java/io/github/resilience4j/timelimiter/internal/TimeLimiterImpl.java#L184

RobWin:
Ah! You are using the timelimiter as well. Didn't know that the fallback method was handling a timeout exception in your case.

RobWin:
Yes, here we need to propagate the Thread Context

razum90:
Ah I see, using the context propagator? If so, how should we go about and structure it? I see it's now only a part of thread-pool-bulkhead

RobWin:
Yes, I guess we need to add the support for context propagator also to the TimeLimiter component.

razum90:
I agree. I guess it would be nice only having to pointing out once which context propagator to use, instead of in both in the thread-pool-bulkhead + timelimiter module. Or do you see a case where you would want to use different context propagators for them?

razum90:
Hey @RobWin, how should we proceed with this? Do you guys have plans on fixing this yourselves or should I spend some time on it?

RobWin:
We are disucssing it internally ... ironically right at this moment,

RobWin:
Currently we are investigating, if our `ThreadPoolBulkhead` could implement `ScheduledExecutorService`  and also used by TimeLimiter/Retry to schedule retry attempts or TimeoutExceptions.

razum90:
Hi @RobWin, did you come to a conclusion with the investigation?

RobWin:
Yes, we are working on this PR -> https://github.com/resilience4j/resilience4j/pull/1204

PR #1148, opened 2020-08-29
SSE springboot test fix

vijaycse:
This fixes the test case failures in CI
https://github.com/resilience4j/resilience4j/pull/1093


vijaycse:
@RobWin  Can you take a look at this PR? I ran several times locally and got few failures sometimes and after this fix , no longer getting any failures.

RobWin:
Thx

PR #1159, opened 2020-09-09
Missing property for writable stack trace in bulk config properties

worldtiki:
Missing property for writable stack trace in bulk config properties


RobWin:
Thx. Could you please add a small test?

Romeh:
Sorry for my late request , can u please extend spring boot 2 integration test in resilience4j spring boot 2 module by adding the new property there and assert it is loaded properly ? in https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring-boot2/src/test/resources/application.yaml#L119

worldtiki:
> https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring-boot2/src/test/resources/application.yaml#L119

Sure :)

RobWin:
thx

Issue #1168, opened 2020-09-17
Timelimiter metrics not calculated when using resilince4j-kotlin


catadincu:
Resilience4j version: 1.5.0

Kotlin version: 1.4.10

Timelimiter metrics are not calculated when using the resilience4j-kotlin module, on the `Timelimiter.decorateSuspendFunction` extension function.

Probably the counter `resilience4j_timelimiter_calls_total` should increment whenever a `TimeoutCancellationException` is thrown in the extension function, however its value remains 0.

RobWin:
Hi, would you like to have a look at it and raise a PR?

abhishekweb:
> Hi, would you like to have a look at it and raise a PR?

Any update on this?

ZzAve:
I'm dealing with the same issue. It seems that https://github.com/resilience4j/resilience4j/pull/1095 addresses this!

ZzAve:
I actually checked out the latest master and found that although the counter do increment, the timeouts aren't registered as timeouts, but as errors

```
resilience4j_timelimiter_calls_total{kind="timeout",name="my-target",} 0.0
resilience4j_timelimiter_calls_total{kind="successful",name="my-target",} 0.0
resilience4j_timelimiter_calls_total{kind="failed",name="my-target",} 103.0
```

What I'd expect is that all errors would be considered an error but the TimeoutExceptions. Allow me to put in an mr?

RobWin:
Yes, please

ZzAve:
The behaviour's a little more fine fineTuned than I expected, although mostly related to kotlin's way of dealing with `CancellationException`s

I'm wrapping my timeLimiter into a circuitbreaker, such that I can stop hammering an upstream service if it's getting really slow. 

```
circuitbreaker
   timelimiter
       <codeBlock calling an upstream service>
```

Now, if the timeLimiter returns in an error state (calling the the `onError` method), the circuitbreaker doesn't pick up the underlying event as an Error as well, and there are no metrics on the event as well, which is something I don't really get (yet). It seems that an error in terms of a timelimiter should never really occur?

Because of the intricate behaviour of kotlin's `TimeoutCancellationException`s, for which I want to be quite explicit in my usage, I did the following:

```kotlin
suspend fun <T>  executeCircuitBreakingTimeLimitedFun(
    circuitBreaker: CircuitBreaker,
    timeLimiter: TimeLimiter,
    description: String,
    block: suspend () -> T
): T {
[...]

PR #1095, opened 2020-07-24
Allows metrics collection for coroutines

piyushmor:
Currently we lack the metrics for the Kotlin time limiter. This PR aims to address that.

Issue #1123, opened 2020-08-17
Circuitbreaker doesn´t open when nested TimeLimiter throws TimeOutCancellationException


dennisschroeder:
Resilience4j version: 1.5.0

Java version: 11
Kotlin version: 1.3.71

**Actual behavior**
A Circuit breaker does **NOT** open when a nested TimeLimiter throws TimeOutCancellationException.

```kotlin
import assertk.assertThat
import assertk.assertions.isEqualTo
import io.github.resilience4j.circuitbreaker.CircuitBreaker
import io.github.resilience4j.kotlin.circuitbreaker.CircuitBreakerConfig
import io.github.resilience4j.kotlin.circuitbreaker.executeSuspendFunction
import io.github.resilience4j.kotlin.timelimiter.executeSuspendFunction
import io.github.resilience4j.timelimiter.TimeLimiter
import kotlinx.coroutines.TimeoutCancellationException
import kotlinx.coroutines.delay
import kotlinx.coroutines.runBlocking
import org.junit.jupiter.api.Test
import org.junit.jupiter.api.TestInstance
import org.junit.jupiter.api.assertThrows
import java.time.Duration

@TestInstance(TestInstance.Lifecycle.PER_CLASS)
internal class ResilienceNestingTest {

    @Test()
    fun `circuit breaker does not open when time limiter throws TimeOutCancellationException`() {
        val circuitBreakerTestConfig = CircuitBreakerConfig {
            slidingWindowSize(2)
            failureRateThreshold(50F)
        }

        val circuitBreakerUnderTest = CircuitBreaker.of("test", circuitBreakerTestConfig)
        val timeLimiterUnderTest = TimeLimiter.of(Duration.ofMillis(1000))

        assertThat(circuitBreakerUnderTest.state).isEqualTo(CircuitBreaker.State.CLOSED)

        repeat(2) {
            assertThrows<TimeoutCancellationException> {
                runBlocking {
                    circuitBreakerUnderTest.executeSuspendFunction {
                        timeLimiterUnderTest.executeSuspendFunction {
                            delay(1100)
                        }
                    }
                }
            }
        }

        assertThat(circuitBreakerUnderTest.state).isEqualTo(CircuitBreaker.State.CLOSED)
    }

    @Test()
    fun `circuit breaker opens when RuntimeException gets thrown by wrapped time limiter`() {
        val circuitBreakerTestConfig = CircuitBreakerConfig {
            slidingWindowSize(2)
[...]

Issue #758, opened 2019-11-20
Decorate supplier/callable fails for kotlin exceptions


maczikasz:
Thanks for raising a Resilience4j issue.
Please provide a brief description of your problem along with the versions you are using.
If possible, please also consider putting together a complete JUnit test that reproduces the issue.

Resilience4j version: 1.1.0

Java version: Kotlin

Problem description:
Retry will not retry on Kotlin exceptions for normal non suspending or non flow koltin functions that will throw Kotlin exceptions, the assumption that supplier/callable throws only RuntimeException is valid in Java world but not in Kotlin

`
import io.github.resilience4j.retry.Retry

class IamARuntimeException : RuntimeException("")
class IamAKotlinException : Exception("")

fun main() {
    val workingRetry = Retry.ofDefaults("i-work")
    val bustRetry = Retry.ofDefaults("i-dont")

    var iWillBeIncremented = 0
    var iAmTouchedOnlyOnce = 0

    println("A working retry")
    try {
        workingRetry.executeSupplier { iWillBeIncremented++; print(iWillBeIncremented); throw IamARuntimeException(); iWillBeIncremented }
    } catch (e: IamARuntimeException) {
    }

    println("\nNow a busted retry")
    try {
        bustRetry.executeSupplier { iWillBeIncremented++; print(iAmTouchedOnlyOnce); throw IamAKotlinException(); iAmTouchedOnlyOnce }
    } catch (e: IamAKotlinException) {
    }

    println(iWillBeIncremented)
    println(iAmTouchedOnlyOnce)
}
`

RobWin:
Hi,
thank you.
What is your suggestion?

marcoferrer:
I think his case relates to the fact that Kotlin doesnt have a concept of checked exceptions. Any method invocation could potentially throw an instance of `Throwable`. This is the reason why Kotlins `runCatching` method in the stdlib catches `Throwable`

RobWin:
How would you fix this?

piyushmor:
shouldn't you use `executeCheckedSupplier ` instead of `executeSupplier` to get the desired behavior? Also, in context of Kotlin, it is better to use `executeFunction`. TBH, I don't really get the issue you are facing

PR #1181, opened 2020-09-29
Timelimiter metrics/events: Register timeouts as timeouts and all other exceptions as errors

ZzAve:
Also see https://github.com/resilience4j/resilience4j/issues/1168 and https://github.com/resilience4j/resilience4j/pull/1095

The changes I'm making to resilience:
- A timelimiter rethrows any exception
- A timelimiter records `TimeoutCancellationException`s as `java.util.concurrent.TimeoutException`s, such that the event handler recognises this as a timeout.
- A timelimiter records any other exception as an error.
- A timelimiter records successfull results if no exception occured during execution

### DIFF
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
@@ -47,8 +47,7 @@ suspend fun <T> TimeLimiter.executeSuspendFunction(block: suspend () -> T): T =
         }
     } catch (t: Throwable) {
         if (isCancellation(coroutineContext, t)) {
-            val timeoutException = createdTimeoutExceptionWithName(name)
-            timeoutException.stackTrace = t.stackTrace
+            val timeoutException = TimeLimiter.createdTimeoutExceptionWithName(name, t)
             onError(timeoutException)
             throw timeoutException
         }
@@ -74,8 +73,4 @@ suspend fun <T> TimeLimiter.executeSuspendFunction(block: suspend () -> T): T =
  */
 fun <T> TimeLimiter.decorateSuspendFunction(block: suspend () -> T): suspend () -> T = {
     executeSuspendFunction(block)
-}
-
-private fun createdTimeoutExceptionWithName(name: String?): TimeoutException {
-    return TimeoutException(String.format("TimeLimiter '%s' recorded a timeout exception.", name))
 }
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
@@ -22,6 +22,7 @@ import io.github.resilience4j.kotlin.isCancellation
 import io.github.resilience4j.timelimiter.TimeLimiter
 import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.withTimeout
+import java.util.concurrent.TimeoutException
 import kotlin.coroutines.coroutineContext
 
 /**
@@ -45,9 +46,9 @@ suspend fun <T> TimeLimiter.executeSuspendFunction(block: suspend () -> T): T =
         }
     } catch (t: Throwable) {
         if (isCancellation(coroutineContext, t)) {
-            onError(t)
+            onError(TimeoutException(t.message))
         } else {
-            onSuccess()
+            onError(t)
         }
         throw t
     }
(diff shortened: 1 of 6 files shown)


RobWin:
I guess we never noticed this misbehavior, because the unit tests don't check the metrics or the number of published events.

Could you please enhance the `CoroutineTimeLimiterTest.kt` ?

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
```diff
@@ -45,9 +46,9 @@ suspend fun <T> TimeLimiter.executeSuspendFunction(block: suspend () -> T): T =
         }
     } catch (t: Throwable) {
         if (isCancellation(coroutineContext, t)) {
-            onError(t)
+            onError(TimeoutException(t.message))
```
RobWin, line 49: Would it even make sense to rethrow a  `TimeoutException`? 
It might solve the following issue -> https://github.com/resilience4j/resilience4j/issues/1123

RobWin (changes requested):

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
ZzAve: The nesting part in Kotlin is indeed a problem. It's something that I solved locally in my application as not to mess the timeLimiter error signals, which, as the comments in the code suggest actually throw a coroutine `TimeoutCancellationException`.

Catching the exception and mapping it to a `java.util.concurrent.TimeoutException` does change the behaviour, and people currently handling CancellationExceptions should be made aware of this change.

Nevertheless, I am in favor of this change, which would also allow me to get rid of the helper function that I wrote here: https://github.com/resilience4j/resilience4j/issues/1168#issuecomment-700704364

RobWin: Do you also have a solution for https://github.com/resilience4j/resilience4j/issues/688

RobWin: Yes, we have to document this behavior in the JavaDoc.
In general this problem is related to issue https://github.com/resilience4j/resilience4j/issues/758

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
```diff
@@ -45,10 +47,12 @@ suspend fun <T> TimeLimiter.executeSuspendFunction(block: suspend () -> T): T =
         }
     } catch (t: Throwable) {
         if (isCancellation(coroutineContext, t)) {
-            onError(t)
-        } else {
-            onSuccess()
+            val timeoutException = TimeoutException(t.message)
```
RobWin, line 50: I think we should not loose the root cause here, or?

Can you use `TimeoutException(t)`.
Does that work in Kotlin?

RobWin (changes requested):

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
RobWin: Do we need to add `@Throws(TimeoutException::class)` to the function?

ZzAve: TimeoutException only has two constructors:
```
    public TimeoutException() 
    public TimeoutException(String message) 
```

Not really any way to add a cause or anything to it. That is a bit unfortunate though.  and I agree a stack trace would be very valuable.

Of course, we can always use extension since TimeoutException extends Exception , which does have a 'cause' variable.

RobWin: It's a long time since I used Kotlin: Why can't you use the constructors of the parent Java class in Kotlin?

RobWin: True.
We could do

```
val timeoutException = TimeoutException(t.message)
timeoutException.stackTrace = t.stackTrace
```

RobWin (approved):

RobWin:
Thank you. Looks fine to me.

ZzAve:
I extracted the logic for the TimeoutException so it exists only once. It does live in an interface though, I'm not sure if that's a code smell or not?

RobWin:
Thank you

PR #1183, opened 2020-09-29
feat(micronaut): update intercept and test cases

pollend:
I move the proceed call into the executionCompletionStage. it should still handle it for the fallback in proceed returns null right?

### DIFF
--- a/resilience4j-micronaut/src/main/java/io/github/resilience4j/mirconaut/circuitbreaker/CircuitBreakerInterceptor.java
+++ b/resilience4j-micronaut/src/main/java/io/github/resilience4j/mirconaut/circuitbreaker/CircuitBreakerInterceptor.java
@@ -36,6 +36,7 @@ import io.reactivex.Flowable;
 import javax.inject.Singleton;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 
 @Singleton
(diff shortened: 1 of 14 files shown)


resilience4j-micronaut/src/main/java/io/github/resilience4j/mirconaut/circuitbreaker/CircuitBreakerInterceptor.java
```diff
@@ -83,11 +83,7 @@ public Object intercept(MethodInvocationContext<Object, Object> context) {
         ReturnType<Object> rt = context.getReturnType();
         Class<Object> returnType = rt.getType();
         if (CompletionStage.class.isAssignableFrom(returnType)) {
-            Object result = context.proceed();
-            if (result == null) {
-                return result;
-            }
-            return this.fallbackCompletable(circuitBreaker.executeCompletionStage(() -> ((CompletableFuture<?>) result)), context);
+            return this.fallbackCompletable(circuitBreaker.executeCompletionStage(() -> ((CompletableFuture<?>) context.proceed())), context);
         } else if (Publishers.isConvertibleToPublisher(returnType)) {
```
Romeh, line 87: Same for reactive types please

Romeh (changes requested):

Romeh:
@pollend Per our discussion which was fun ;) , fallback logic just need to be revisited to make sure it cover the checked exception case as well

Romeh (approved):

Issue #1044, opened 2020-06-25
Retry: Exponential backoff with constant behaviour after certain time


pbarsotti-glovo:
I came across some situations where I needed to retry using an exponential approach with an initial aggressive policy (multiply by 1.5 or 2) to avoid retrying too soon, but after a while of retrying, the wait times become too large to be useful.
I ended creating a custom interval function based on the basic exponential backoff and including a `maxInterval` value. The idea is to grow exponentially until the `maxInterval` threshold is reached and then return a constant value (`maxInterval`).

I wonder if someone has faced the same situation and think that this could be a good addition to the existing library.

RobWin:
Yes, sounds interesting.
Would you like to implement such a feature?

pbarsotti-glovo:
Sure. Let me find some time and I'll submit a PR.

paezba:
I just submitted the PR. Thanks!

alecl:
This is a nice feature and I only found it digging through the source as it wasn't on https://resilience4j.readme.io/docs/retry

PR #1185, opened 2020-10-01
Issue #1044: Added maxInterval to limit exponential backoff - Spring properties (#1057)

hexmind:


resilience4j-framework-common/src/main/java/io/github/resilience4j/common/retry/configuration/RetryConfigurationProperties.java
```diff
@@ -231,49 +252,62 @@ private void configureRetryIntervalFunction(InstanceProperties properties,
...
+        /**
          * exponential backoff multiplier value
          */
         private Double exponentialBackoffMultiplier;
 
-        /*
+        /**
+         * exponential max interval value
+         */
+        private Long exponentialMaxIntervalMillis;
```
RobWin, line 302: Can we use Duration and call it `maxWaitDuration`?

RobWin (changes requested):

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/retry/configuration/RetryConfigurationProperties.java
hexmind: good idea, done

RobWin (approved):

Issue #1043, opened 2020-06-25
Dynamic configuration for retryConfig


ranarula:
Resilience4j version: 1.5.0

Java version: 11

Thanks for raising a Resilience4j issue.
Is there a way to dynamically update the interval function for the retry module. 

Use case : I want to perform retries based on the RETRY_AFTER for my http requests. Currently the only way to configure retry is by pre-configuring an IntervalFunction on the retryConfig. 

I have configured the `resultPredicate` to look for the header to retry but I would want to retry based on the time specified in the header field. 
I was wondering if there is a way to configure the RetryConfig for this scenario.

RobWin:
Hi,
that means you want to reconfigure the `resultPredicate` at runtime?

ranarula:
not exactly - 

I use the result predicate to verify if we need to dynamically delay (by checking the value of the response header `Retry-After`) The delay value to be used is from the value sent in the header. 

Please check section 14.37 for https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html for more details.

My query was if there is a way to override the IntervalFunction to use this value and retry after the time as received in the `Retry-After` header.

RobWin:
No, but it would a nice new feature. Even if the IntervalFuntion would be mutable, it wouldn't work, because the function is used for all requests which are retried. As far as I understand you want to have a different intervalfunction per request/response header, correct? And the `Retry-After` header can be different for every response?

RobWin:
For example:

`(Integer attempt, T result) -> Long waitDuration`

ranarula:
Right exactly, a BiFunction support for IntervalFunction will do the trick

cosminseceleanu:
Hi @RobWin, any news on this? Is this feature on the roadmap?

cosminseceleanu:
@RobWin I want to open a pull request but before starting to write some code I have question. Should I use a single function as you mentioned or two functions 'intervalOnResultFunction' and `intervalOnExceptionPredicate`?
I'm asking because for predicates the current API uses two.

RobWin:
Sry for the late response. 
I think I would even prefer two functions:
`intervalFunctionOnResult` and `intervalFunctionOnException`.

If a result must be retried, either `intervalFunctionOnResult` or as a fallback `interfalFunction` is used. 
If an exception must be retried, either `intervalFunctionOnException` or as a fallback `interfalFunction` is used. 

What do you think?

cosminseceleanu:
[...]

PR #1200, opened 2020-10-09
Issue #1043 Added IntervalBiFunction to calculate wait duration based on result/exception

cosminseceleanu:
Pull request for issue https://github.com/resilience4j/resilience4j/issues/1043

Added `interface IntervalBiFunction<T> extends BiFunction<Integer, Either<T, Throwable>, Long>` to calculate retry wait interval on result or exception. When `IntervalBiFunction` is not set then the it will use the `IntervalFunction` as fallback.

An use case for this feature is to be able to respect HTTP `Retry-After` header.

Questions: 
- `IntervalBiFunction` should be placed in `resilience4j-core` or `resilience4j-retry`?
- `RetryConfig.intervalFunction` should be marked as deprecated? `RetryConfig.IntervalBiFunction` is the only one used to build a `Retry` instance. -> yes

### DIFF
--- a/resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
+++ b/resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
@@ -1,0 +1,18 @@
+package io.github.resilience4j.core;
+
+import io.vavr.control.Either;
+
+import java.util.function.BiFunction;
+
+/**
+ * An IntervalBiFunction which can be used to calculate the wait interval. The input parameters of the bi
+ * function is the number of attempts (attempt) and either result or exception, the output parameter is the wait interval in
+ * milliseconds. The attempt parameter starts at 1 and increases with every further attempt.
+ */
+@FunctionalInterface
+public interface IntervalBiFunction<T> extends BiFunction<Integer, Either<T, Throwable>, Long> {
+
+    static <T> IntervalBiFunction<T> ofIntervalFunction(IntervalFunction f) {
+        return (attempt, either) -> f.apply(attempt);
+    }
+}
--- a/resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
+++ b/resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
@@ -10,7 +10,7 @@ import java.util.function.BiFunction;
  * milliseconds. The attempt parameter starts at 1 and increases with every further attempt.
  */
 @FunctionalInterface
-public interface IntervalBiFunction<T> extends BiFunction<Integer, Either<T, Throwable>, Long> {
+public interface IntervalBiFunction<T> extends BiFunction<Integer, Either<Throwable, T>, Long> {
 
     static <T> IntervalBiFunction<T> ofIntervalFunction(IntervalFunction f) {
         return (attempt, either) -> f.apply(attempt);
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
@@ -19,12 +19,14 @@
 package io.github.resilience4j.retry;
 
 
+import io.github.resilience4j.core.IntervalBiFunction;
 import io.github.resilience4j.core.IntervalFunction;
 import io.github.resilience4j.core.lang.Nullable;
 import io.github.resilience4j.core.predicate.PredicateCreator;
 
 import java.io.Serializable;
 import java.time.Duration;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
@@ -35,6 +37,7 @@ public class RetryConfig implements Serializable {
     public static final long DEFAULT_WAIT_DURATION = 500;
     public static final int DEFAULT_MAX_ATTEMPTS = 3;
     private static final IntervalFunction DEFAULT_INTERVAL_FUNCTION = numOfAttempts -> DEFAULT_WAIT_DURATION;
+    private static final IntervalBiFunction DEFAULT_BI_INTERVAL_FUNCTION = IntervalBiFunction.ofIntervalFunction(DEFAULT_INTERVAL_FUNCTION);
     private static final Predicate<Throwable> DEFAULT_RECORD_FAILURE_PREDICATE = throwable -> true;
 
     @SuppressWarnings("unchecked")
@@ -49,6 +52,8 @@ public class RetryConfig implements Serializable {
 
     private int maxAttempts = DEFAULT_MAX_ATTEMPTS;
     private IntervalFunction intervalFunction = DEFAULT_INTERVAL_FUNCTION;
+
+    private IntervalBiFunction intervalBiFunction = DEFAULT_BI_INTERVAL_FUNCTION;
 
     // The final exception predicate
     private Predicate<Throwable> exceptionPredicate;
@@ -90,6 +95,17 @@ public class RetryConfig implements Serializable {
         return intervalFunction;
     }
 
+    /**
+     * Return the IntervalBiFunction which calculates wait interval based on result or exception
+     *
+     * @param <T> The type of result.
+     * @return the interval bi function
+     */
+    @SuppressWarnings("unchecked")
+    public <T> IntervalBiFunction<T> getIntervalBiFunction() {
+        return intervalBiFunction;
+    }
+
     public Predicate<Throwable> getExceptionPredicate() {
         return exceptionPredicate;
     }
@@ -117,6 +133,9 @@ public class RetryConfig implements Serializable {
         @Nullable
         private Predicate<T> retryOnResultPredicate;
 
+        @Nullable
+        private IntervalBiFunction<T> intervalBiFunction;
+
         @SuppressWarnings("unchecked")
         private Class<? extends Throwable>[] retryExceptions = new Class[0];
         @SuppressWarnings("unchecked")
@@ -132,6 +151,7 @@ public class RetryConfig implements Serializable {
             this.intervalFunction = baseConfig.intervalFunction;
             this.retryOnExceptionPredicate = baseConfig.retryOnExceptionPredicate;
             this.retryOnResultPredicate = baseConfig.retryOnResultPredicate;
+            this.intervalBiFunction = baseConfig.intervalBiFunction;
             this.retryExceptions = baseConfig.retryExceptions;
             this.ignoreExceptions = baseConfig.ignoreExceptions;
         }
@@ -176,6 +196,18 @@ public class RetryConfig implements Serializable {
          */
         public Builder<T> intervalFunction(IntervalFunction f) {
             this.intervalFunction = f;
+            return this;
+        }
+
+        /**
+         * Set a function to modify the waiting interval after a failure based on attempt number and result or exception.
+         * When is set will discard values configured via {@link Builder#intervalFunction} or {@link Builder#waitDuration}.
+         *
+         * @param f Function to modify the interval after a failure
+         * @return the RetryConfig.Builder
+         */
+        public Builder<T> intervalBiFunction(IntervalBiFunction<T> f) {
+            this.intervalBiFunction = f;
             return this;
         }
 
@@ -251,6 +283,8 @@ public class RetryConfig implements Serializable {
             config.retryExceptions = retryExceptions;
             config.ignoreExceptions = ignoreExceptions;
             config.exceptionPredicate = createExceptionPredicate();
+            config.intervalBiFunction = Optional.ofNullable(intervalBiFunction)
+                .orElse(IntervalBiFunction.ofIntervalFunction(intervalFunction));
             return config;
         }
 
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
@@ -179,7 +179,7 @@ public class RetryConfig implements Serializable {
 
         public Builder<T> waitDuration(Duration waitDuration) {
             if (waitDuration.toMillis() >= 0) {
-                this.intervalFunction = (x) -> waitDuration.toMillis();
+                this.intervalBiFunction = (attempt, either) -> waitDuration.toMillis();
             } else {
                 throw new IllegalArgumentException(
                     "waitDuration must be a positive value");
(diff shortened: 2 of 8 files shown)


cosminseceleanu:
@RobWin can you take a look?

RobWin:
Yes, I'm on vacation.

resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.function.BiFunction;
+
+/**
+ * An IntervalBiFunction which can be used to calculate the wait interval. The input parameters of the bi
+ * function is the number of attempts (attempt) and either result or exception, the output parameter is the wait interval in
+ * milliseconds. The attempt parameter starts at 1 and increases with every further attempt.
+ */
+@FunctionalInterface
+public interface IntervalBiFunction<T> extends BiFunction<Integer, Either<T, Throwable>, Long> {
```
RobWin, line 13: By convention, Either's Left represents a failure case and the Right represents a success.
T and `Throwable` must be switched.

RobWin (changes requested):

resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
cosminseceleanu: didn't know, will change 👍

RobWin: Sry, maybe it is even more clear, if we use `Try` instead of `Either`.

That means: `Either<T, Throwable>`  --> `Try<T>`
and `Either.right(...)` --> `Try.success(...)`
and `Either.left(...)` --> `Try.failure(...)`

and  `Throwable result = retryAttempt.isSuccess() ? null : retryAttempt.getCause();`

RobWin (changes requested):
Sry, maybe it is even more clear, if we use `Try` instead of `Either`.

That means: `Either<T, Throwable>` --> `Try<T>`
and `Either.right(...)` --> `Try.success(...)`
and `Either.left(...)` --> `Try.failure(...)`

and `Throwable throwable = result.isFailure() ? result.getCause() : null;`

cosminseceleanu:
Not sure that is more clear:

1. I think `Try` should be used to clearly identify a potential failure in the computation, but in this case there is no computation. And by computation I mean something like `Try.of(httpClient::call)` which is clearly not the case here. 
2. `Try` has `recover` features which may be misleading for clients

`Either represents a value of two possible types. An Either is either a Left or a Right.` - from vavr java docs. This is clear our case.

RobWin:
Ok, you convinced me :)

RobWin (approved):

resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
```diff
@@ -0,0 +1,18 @@
+package io.github.resilience4j.core;
+
+import io.vavr.control.Either;
```
dlsrb6342, line 3: We are trying to remove `vavr` dependency in 2.0 branch.
Is there any substitute for`Either`?

RobWin: We have to provide our own implementation in 2.0

dlsrb6342:
Like #1214, Doesn't need to check both `IntervalFunction` and `IntervalBiFunction` are set?

RobWin:
Yes, this makes sense as well

cosminseceleanu:
My takeaway from https://github.com/resilience4j/resilience4j/issues/1214 is that if `IntervalFunction` and `IntervalBiFunction` are used together then when `build()` is invoked `IllegalStateException` should be thrown, right?

resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
cosminseceleanu: just for my personal curiosity, why do you want to get rid of `vavr`?

RobWin:
Yes, exactly

resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java
RobWin: Because a lot of users requested it, unfortunately.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -150,7 +171,7 @@ public Builder(RetryConfig baseConfig) {
                 this.intervalFunction = (x) -> waitDuration.toMillis();
             } else {
                 throw new IllegalArgumentException(
-                    "waitDurationInOpenState must be a positive value");
```
cosminseceleanu, line 153: is this considered a breaking change?

I think it was copy pasted from circuit breaker 😄

RobWin: No :) we should fix the message :)

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -90,6 +95,18 @@ public int getMaxAttempts() {
         return intervalFunction;
```
RobWin, line 95: I think we should deprecate `getIntervalFunction`, since it shouldn't be used anymore in our  `RetryImpl`. Makes it more obvious where it is still used. 
Comparable to `getWaitDurationInOpenState` in `CircuitBreakerConfig`

RobWin (changes requested):

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
cosminseceleanu: This was one of my questions from PR description 😃

RobWin: I misread the PR description. I thought you want to deprecate `IntervalFunction` itself. 
Too many topics in parallel, sorry.

cosminseceleanu:
https://travis-ci.org/github/resilience4j/resilience4j/builds/737387964 - I don't understand why it fails, on my local machine it works. Is just a flaky test?

RobWin:
I restarted the test. Let's see

RobWin:
Yes, it was a flaky test

resilience4j-retry/src/test/java/io/github/resilience4j/retry/RetryConfigBuilderTest.java
```diff
@@ -168,6 +181,13 @@ public void shouldBuilderCreateConfigEveryTime() {
         assertThat(config1.getMaxAttempts()).isEqualTo(5);
     }
 
+    @Test(expected = IllegalStateException.class)
+    public void intervalFunctionUsedWithIntervalBiFunctionShouldFail() {
```
RobWin, line 185: Could you please add a final test that if neither intervalBiFunction nor intervalFunction is set, that the default setting is used in the RetryConfig?

RobWin (changes requested):

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -24,19 +24,21 @@
...
+import io.github.resilience4j.core.IntervalBiFunction;
 import io.github.resilience4j.retry.event.*;
 import io.vavr.CheckedConsumer;
 import io.vavr.collection.HashMap;
 import io.vavr.collection.Map;
+import io.vavr.control.Either;
 import io.vavr.control.Option;
 import io.vavr.control.Try;
 
+import java.util.Optional;
```
dlsrb6342, line 36: I think this is unused import.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -243,14 +280,19 @@ public Builder(RetryConfig baseConfig) {
         }
 
         public RetryConfig build() {
+            if (intervalFunction != null && intervalBiFunction != null) {
+                throw new IllegalStateException("The intervalFunction was configured twice which could result in an" +
+                    " undesired state. Please use either intervalFunction or intervalBiFunction.");
+            }
             RetryConfig config = new RetryConfig();
-            config.intervalFunction = intervalFunction;
+            config.intervalFunction = Optional.ofNullable(intervalFunction).orElse(IntervalFunction.ofDefaults());
```
dlsrb6342, line 288: `config.intervalFunction` isn't used anymore in `RetryImpl`. 
If set only `IntervalFunction`, I think code should be like following 

```java
config.intervalBiFunction = Optional.ofNullable(intervalBiFunction)
                                    .orElse(IntervalBiFunction.ofIntervalFunction(intervalFunction));
```

dlsrb6342 (changes requested):

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
cosminseceleanu: this means that `config.intervalFunction` should be marked as deprecated and nullable

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
cosminseceleanu: no worries

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
RobWin: Yes, I think that's fine

cosminseceleanu: Hmm, I tried this and found an interesting use case, the following code will throw `IllegalStateException`

```java
RetryConfig baseConfig = RetryConfig.ofDefaults();
RetryConfig retryConfig = RetryConfig.from(baseConfig)
            .intervalFunction(IntervalFunction.of(100L))
            .build();
```

Because `baseConfig.intervalBiFunction = bi function of interval defaults` and since `config.intervalBiFunction` is the only used by `Retry`, I thought that only this value should be copied from base config on `RetryConfig.from(baseConfig)`, therefore `retryConfig.intervalBiFunction = baseConfig.intervalBiFunction`

I'm kind of stuck with this, any idea?

cosminseceleanu: some code
```java
private IntervalBiFunction<T> createIntervalBiFunction() {
            if (intervalFunction != null && intervalBiFunction != null) {
                throw new IllegalStateException("The intervalFunction was configured twice which could result in an" +
                    " undesired state. Please use either intervalFunction or intervalBiFunction.");
            }
            if (intervalFunction == null && intervalBiFunction == null) {
               return IntervalBiFunction.ofDefaults();
            }
            if (intervalFunction != null) {
                return IntervalBiFunction.ofIntervalFunction(intervalFunction);
            }

            return intervalBiFunction;
        }
```

```java
public Builder(RetryConfig baseConfig) {
            this.maxAttempts = baseConfig.maxAttempts;
            this.retryOnExceptionPredicate = baseConfig.retryOnExceptionPredicate;
            this.retryOnResultPredicate = baseConfig.retryOnResultPredicate;
            this.retryExceptions = baseConfig.retryExceptions;
            this.ignoreExceptions = baseConfig.ignoreExceptions;
            this.intervalBiFunction = baseConfig.intervalBiFunction;
        }
```

RobWin: Maybe we just use a int counter. If the counter == 2, then the user has used both methods.

cosminseceleanu: never mind, I think I was too tired

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
```diff
@@ -150,7 +182,7 @@ public Builder(RetryConfig baseConfig) {
                 this.intervalFunction = (x) -> waitDuration.toMillis();
```
dlsrb6342, line 182: Doesn't need to set `IntervalBiFunction`? 
```suggestion
                this.intervalBiFunction = (attempt, either) -> waitDuration.toMillis();
```

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
RobWin: This is how it was implemented in CircuitBreakerConfig now -> https://github.com/resilience4j/resilience4j/pull/1216/files

cosminseceleanu: I was trying to avoid that kind of implementation. 

What I've done is this: whenever `intervalFunction` is set then when creating a new builder with `RetryConfig.from` only `intervalFunction` is copied from base config. Because `intervalBiFunction` is always present and when `build` is invoked will be created from `intervalFunction`. 

See the implementation :)

resilience4j-retry/src/main/java/io/github/resilience4j/retry/RetryConfig.java
cosminseceleanu: done

cosminseceleanu:
```java
io.github.resilience4j.circuitbreaker.monitoring.events.CircuitBreakerHystrixStreamEventsTest > streamAllEvents FAILED

    java.lang.AssertionError at CircuitBreakerHystrixStreamEventsTest.java:81

Error receiving SSE: {}reactor.netty.http.client.PrematureCloseException: Connection prematurely closed DURING response

2020-10-23 08:18:35.047  INFO 8494 --- [extShutdownHook] o.s.s.concurrent.ThreadPoolTaskExecutor  : Shutting down ExecutorService 'applicationTaskExecutor'

2020-10-23 08:18:35.053  INFO 8494 --- [extShutdownHook] o.s.s.concurrent.ThreadPoolTaskExecutor  : Shutting down ExecutorService 'applicationTaskExecutor'

2020-10-23 08:18:35.071  INFO 8494 --- [extShutdownHook] o.s.s.concurrent.ThreadPoolTaskExecutor  : Shutting down ExecutorService 'applicationTaskExecutor'

2020-10-23 08:18:37.076  INFO 8494 --- [extShutdownHook] o.s.s.concurrent.ThreadPoolTaskExecutor  : Shutting down ExecutorService 'applicationTaskExecutor'
```

seems like a flaky test

RobWin (approved):

RobWin:
Would you like to add a documentation to: https://resilience4j.readme.io/docs/retry
And I think we should create another PR and allow the configuration of a IntervalBiFunction in Spring Boot2.

slawekjaranowski:
After it - using `waitDuration` and `intervalFunction` together on `RetryConfig.Builder` cause `IllegalStateException`.
Probably it is correct behavior but not documented.

PR #1206, opened 2020-10-14
Change timeLimiterExecutorService to instance field

amoffet:
Enhancement 1203 -
When running integration tests that refresh the Spring application context, the aspect is closed with the context. When the context is refreshed, the static Executor is not reinstantiated. Consequently, subsequent invocations of a @Timelimiter annotated method fail with a java.util.concurrent.RejectedExecutionException (The scheduler is shut down). The executor needs to be recreated each time the aspect is recreated with the context. Since the aspect is a singleton, the effect of the change should essentially be the same as if the field were static.


RobWin:
Do you have time to fix the issue?

amoffet:
> Do you have time to fix the issue?

Will do today.

amoffet:
Done.  Please let me know if you would like anything changed.

dlsrb6342 (approved):

RobWin:
Thx

Issue #1116, opened 2020-08-10
CircuitBreakers events in actuactor end-point doesn't works after updating in runtime 


JuanMorenoS:
Resilience4j version: 1.5.0
Java version: 11

Hi!

I want to update the circuit-breakers configuration at runtime, I have a schedule that consumes the DB and I update the configuration every 5 minutes (similar to this issue https://github.com/resilience4j/resilience4j/issues/850). To update the CBs I use the `replace` function in the `CircuitbreakerRegistry`, but after the first update, I stop seeing the CBs events at the actuator end point.

RobWin:
Hi,

do you use Spring Boot?
Did you set up resilience4j-micrometer yourself? 
Could you please show me how you did it.

JuanMorenoS:
Hi @RobWin 

Yes, I use the last version of spring-boot 2.3.2 and no I use the R4J default config 

Shure!

That's how I update circuit breakers:
```
CircuitBreaker oldCircuitBreaker = circuitBreakerRegistry.circuitBreaker(name);
CircuitBreaker updatedCircuitBreaker = getLastVersion(oldCircuitBreaker);
circuitBreakerRegistry.replace(name, updatedCircuitBreaker);
```
After replacement, the end point of the actuator shows the oldest CB object

I was checking, and notice that `EventConsumerRegistry<CircuitBreakerEvent>` is never updated.

RobWin:
Hi,
yes you are right. That's a bug / missing feature.

RobWin:
As an alternative you could have a look at Spring Cloud Config.

RobWin:
Would you like to raise a PR?

MrCoderYu:
> Hi @RobWin
> 
> Yes, I use the last version of spring-boot 2.3.2 and no I use the R4J default config
> 
> Shure!
> 
> That's how I update circuit breakers:
> 
> ```
> CircuitBreaker oldCircuitBreaker = circuitBreakerRegistry.circuitBreaker(name);
> CircuitBreaker updatedCircuitBreaker = getLastVersion(oldCircuitBreaker);
> circuitBreakerRegistry.replace(name, updatedCircuitBreaker);
> ```
> 
> After replacement, the end point of the actuator shows the oldest CB object
> 
> I was checking, and notice that `EventConsumerRegistry<CircuitBreakerEvent>` in `CircuitBreakerEventsEndpoint` is never updated.

[...]

PR #1210, opened 2020-10-18
Issue #1116 Handled EntryReplacedEvent in CircuitBreakerRegistry

hexmind:


RobWin (approved):

RobWin:
Do you have to fix this for Retry, RateLimiter and Bulkhead as well?

hexmind:
Not yet, but I see, there is similar problem.

Issue #1214, opened 2020-10-18
When used together, waitDurationInOpenState and waitIntervalFunctionInOpenStat waitDurationInOpenState will be overridden


likhoman:
Resilience4j version:1.5.0

Java version:1.8

Hello, i found bug, it seems to me, and i want to discuss it with you.

I think it is a good idea when a parameter that can be associated with another parameter will be checked. When used together, **waitDurationInOpenState** and **waitIntervalFunctionInOpenStat**, **waitDurationInOpenState** will be overridden.

I create [commit](https://github.com/resilience4j/resilience4j/compare/master...likhoman:bugfix/override-waitDurationInOpenState), please review it and tell me what you think about it.

Hope for feedback. Thanks!

RobWin:
Hi, thank you.
I'm not sure whether throwing an `IllegalArgumentException` or `IllegalStateException` is correct.
Neither an illegal argument is used nor an illegal state is produced. 
The method `waitDurationInOpenState` is rather a simplification (and predecessor) of `waitIntervalFunctionInOpenState`. 
We should adjust the JavaDoc that both methods shouldn't be used at the same time
and we can add a validation that, if they are used at the same time, they produce the same end result (state).
What do you think?

likhoman:
Hi  @RobWin .
Thanks a lot for the quick response! 
I agree that `IllegalArgumentException` and `IllegalStateException` is not correct. Maybe want a specific type of error like `ConfigException` or something  similar?
Where do you want to add verification? When calling a method that sets a property, or when calling an assembly method?
Can I develop this algorithm, or don't you need my help? If you don't need my help it's okay.

Have a nice day!

RobWin:
I think validation/verification should happen when the `build()` method is invoked.

Yes, your help is much appreciated.

RobWin:
merged.
Thank you very much

likhoman:
@RobWin thanks for giving the opportunity to make a commit

RobWin:
Contributions are welcome ;)

PR #1216, opened 2020-10-19
Issue #1214: When used together, waitDurationInOpenState and waitIntervalFunctionInOpenStat waitDurationInOpenState will be overridden

likhoman:
@RobWin i prepared a PR. Please review

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -251,7 +251,7 @@ public class CircuitBreakerConfig implements Serializable {
             .ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD);
         private Duration maxWaitDurationInHalfOpenState = Duration
             .ofSeconds(DEFAULT_WAIT_DURATION_IN_HALF_OPEN_STATE);
-        private byte setWaitInOpenStateCounter = 0;
+        private byte waitInOpenStateCounter = 0;
 
 
         public Builder(CircuitBreakerConfig baseConfig) {
@@ -360,7 +360,7 @@ public class CircuitBreakerConfig implements Serializable {
                     "waitDurationInOpenState must be at least 1[ms]");
             }
             this.waitIntervalFunctionInOpenState = IntervalFunction.of(waitDurationInMillis);
-            setWaitInOpenStateCounter++;
+            waitInOpenStateCounter++;
             return this;
         }
 
@@ -382,7 +382,7 @@ public class CircuitBreakerConfig implements Serializable {
         public Builder waitIntervalFunctionInOpenState(
             IntervalFunction waitIntervalFunctionInOpenState) {
             this.waitIntervalFunctionInOpenState = waitIntervalFunctionInOpenState;
-            setWaitInOpenStateCounter++;
+            waitInOpenStateCounter++;
             return this;
         }
 
@@ -763,9 +763,10 @@ public class CircuitBreakerConfig implements Serializable {
         }
 
         private IntervalFunction validateWaitIntervalFunctionInOpenState() {
-            if (setWaitInOpenStateCounter > 1) {
-                throw new IllegalStateException("The waitIntervalFunction was configured twice which could result in an undesired state. " +
-                    "Please use either waitIntervalFunctionInOpenState or waitDurationInOpenState.");
+            if (waitInOpenStateCounter > 1) {
+                throw new IllegalStateException("The waitIntervalFunction was configured multiple times " +
+                    "which could result in an undesired state. Please verify that waitIntervalFunctionInOpenState " +
+                    "and waitDurationInOpenState are not used together.");
             }
             return waitIntervalFunctionInOpenState;
         }
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -251,7 +251,7 @@ public class CircuitBreakerConfig implements Serializable {
             .ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD);
         private Duration maxWaitDurationInHalfOpenState = Duration
             .ofSeconds(DEFAULT_WAIT_DURATION_IN_HALF_OPEN_STATE);
-        private boolean waitIntervalFunctionInOpenStateAlreadySet = false;
+        private byte setWaitInOpenStateCounter = 0;
 
 
         public Builder(CircuitBreakerConfig baseConfig) {
@@ -360,7 +360,7 @@ public class CircuitBreakerConfig implements Serializable {
                     "waitDurationInOpenState must be at least 1[ms]");
             }
             this.waitIntervalFunctionInOpenState = IntervalFunction.of(waitDurationInMillis);
-            waitIntervalFunctionInOpenStateAlreadySet = true;
+            setWaitInOpenStateCounter++;
             return this;
         }
 
@@ -382,6 +382,7 @@ public class CircuitBreakerConfig implements Serializable {
         public Builder waitIntervalFunctionInOpenState(
             IntervalFunction waitIntervalFunctionInOpenState) {
             this.waitIntervalFunctionInOpenState = waitIntervalFunctionInOpenState;
+            setWaitInOpenStateCounter++;
             return this;
         }
 
@@ -762,7 +763,7 @@ public class CircuitBreakerConfig implements Serializable {
         }
 
         private IntervalFunction validateWaitIntervalFunctionInOpenState() {
-            if (waitIntervalFunctionInOpenStateAlreadySet) {
+            if (setWaitInOpenStateCounter > 1) {
                 throw new IllegalStateException("You have already the durationInOpenState value from the waitDurationInOpenState method." +
                     "Please don't use waitIntervalFunctionInOpenState with waitDurationInOpenState");
             }
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -344,7 +344,7 @@ public class CircuitBreakerConfig implements Serializable {
          * CircuitBreaker should stay open, before it switches to half open. Default value is 60
          * seconds.
          * <p>
-         * <b>ATTENTION!!!</b> Do not use with {@link #waitIntervalFunctionInOpenState(IntervalFunction)}!
+         * Do not use with {@link #waitIntervalFunctionInOpenState(IntervalFunction)}!
          * Please, when using, make sure not to override the value set earlier from the
          * {@link #waitIntervalFunctionInOpenState(IntervalFunction)}
          *
@@ -371,7 +371,7 @@ public class CircuitBreakerConfig implements Serializable {
          * <p>
          * A custom interval function is useful if you need an exponential backoff algorithm.
          * <p>
-         * <b>ATTENTION!!!</b> Do not use with {@link #waitDurationInOpenState(Duration)}!
+         * Do not use with {@link #waitDurationInOpenState(Duration)}!
          * Please, when using, make sure not to override the value set earlier from the
          * {@link #waitDurationInOpenState(Duration)}
          *
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -764,8 +764,8 @@ public class CircuitBreakerConfig implements Serializable {
 
         private IntervalFunction validateWaitIntervalFunctionInOpenState() {
             if (setWaitInOpenStateCounter > 1) {
-                throw new IllegalStateException("You have already the durationInOpenState value from the waitDurationInOpenState method." +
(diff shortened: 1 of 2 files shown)


RobWin:
I'm currently not sure if this might not break backwards-compatibility.
wdyt @Romeh @dlsrb6342 @hexmind ?

likhoman:
Sorry, i try  fixing this...

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -749,5 +760,13 @@ public CircuitBreakerConfig build() {
                 .orElseGet(() -> recordExceptionPredicate != null ? recordExceptionPredicate
                     : DEFAULT_RECORD_EXCEPTION_PREDICATE);
         }
+
+        private IntervalFunction validateWaitIntervalFunctionInOpenState() {
+            if (waitIntervalFunctionInOpenStateAlreadySet) {
```
dlsrb6342, line 765: If I set config like below, doesn't it throw exception?
```java
    CircuitBreakerConfig.custom()
            .waitDurationInOpenState(Duration.ofMillis(3333))
            .build();
```

likhoman: give me time i will fix it today

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -342,6 +343,10 @@ public Builder writableStackTraceEnabled(boolean writableStackTraceEnabled) {
          * Configures an interval function with a fixed wait duration which controls how long the
          * CircuitBreaker should stay open, before it switches to half open. Default value is 60
          * seconds.
+         * <p>
+         * <b>ATTENTION!!!</b> Do not use with {@link #waitIntervalFunctionInOpenState(IntervalFunction)}!
```
hexmind, line 347: This <b>ATTENTION!!!</b> is too strong :) 
I would remove it or replace with more formal "Be aware that ..."

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -749,5 +761,13 @@ public CircuitBreakerConfig build() {
                 .orElseGet(() -> recordExceptionPredicate != null ? recordExceptionPredicate
                     : DEFAULT_RECORD_EXCEPTION_PREDICATE);
         }
+
+        private IntervalFunction validateWaitIntervalFunctionInOpenState() {
+            if (setWaitInOpenStateCounter > 1) {
+                throw new IllegalStateException("You have already the durationInOpenState value from the waitDurationInOpenState method." +
```
RobWin, line 767: Could you please change the message to:
"The waitIntervalFunction was configured twice which could result in an undesired state. Please use either waitIntervalFunctionInOpenState or  waitDurationInOpenState."

RobWin (changes requested):

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -251,6 +251,7 @@ public String toString() {
             .ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD);
         private Duration maxWaitDurationInHalfOpenState = Duration
             .ofSeconds(DEFAULT_WAIT_DURATION_IN_HALF_OPEN_STATE);
+        private byte setWaitInOpenStateCounter = 0;
```
dlsrb6342, line 254: How about use `Duration waitDurationInOpenState` field and check like `if (waitDurationInOpenState != null && waitIntervalFunctionInOpenState != null)` at `build()`?

likhoman: Maybe i move initialization  `waitIntervalFunctionInOpenState` in `builder()` and rename `validateWaitIntervalFunctionInOpenState` to `createWaitIntervalFunctionInOpenState`?

```java
 private IntervalFunction createWaitIntervalFunctionInOpenState() {
            if (waitDurationInOpenState != null && waitIntervalFunctionInOpenState != null ) {
                throw new IllegalStateException("The waitIntervalFunction was configured twice which could result in an undesired state. " +
                    "Please use either waitIntervalFunctionInOpenState or waitDurationInOpenState.");
            }
            return waitDurationInOpenState != null ? IntervalFunction.of(waitDurationInOpenState.toMillis()) : waitIntervalFunctionInOpenState;
        }
```

RobWin: Yes, seems fine to me.

dlsrb6342: Yes exactly. what do you think @RobWin ?

dlsrb6342: Yes exactly.

likhoman: Sorry, i have to clarify one more solution
I want use ` private long waitDurationInMillis = -1L;` instead of `Duration waitDurationInOpenState`.
What would will not calculate `waitDurationInOpenState.toMillis()` twice. 
Because in `public Builder waitDurationInOpenState(Duration waitDurationInOpenState) ` need validation `waitDurationInMillis < 1` what would rise throw developer can saw where method throw exception  in stack trace.
```java
private IntervalFunction createWaitIntervalFunctionInOpenState() {
            final boolean wasItSetToWaitDurationInMillis = waitDurationInMillis > 0;
            if (wasItSetToWaitDurationInMillis && waitIntervalFunctionInOpenState != null ) {
                throw new IllegalStateException("The waitIntervalFunction was configured twice which could result in an undesired state. " +
                    "Please use either waitIntervalFunctionInOpenState or waitDurationInOpenState.");
            }
            return wasItSetToWaitDurationInMillis ? IntervalFunction.of(waitDurationInMillis) : waitIntervalFunctionInOpenState;
        }

```

RobWin: I think this optimization is not needed. Shouldn't be performance critical.

dlsrb6342: Agree with Robert.

likhoman: one more clarification, again) sorry
Could remind you that `waitIntervalFunctionInOpenState` has default value.
And if we use `if (wasItSetToWaitDurationInMillis && waitIntervalFunctionInOpenState != null ) ` we should remove default value for `waitIntervalFunctionInOpenState`.

My attempt
```java
        private IntervalFunction createWaitIntervalFunctionInOpenState() {
            if (waitDurationInOpenState != null && waitIntervalFunctionInOpenState != null) {
                throw new IllegalStateException("The waitIntervalFunction was configured twice which could result in an undesired state. " +
                    "Please use either waitIntervalFunctionInOpenState or waitDurationInOpenState.");
            }

            if (waitDurationInOpenState != null) {
                return IntervalFunction.of(waitDurationInOpenState.toMillis());
            }

            return waitIntervalFunctionInOpenState == null
                ? IntervalFunction.of(Duration.ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD))
                : waitIntervalFunctionInOpenState;
        }
```

But, when we use `CircuitBreakerConfig.from(CircuitBreakerConfig config)`

```java
        CircuitBreakerConfig baseConfig = custom()
            .waitDurationInOpenState(Duration.ofSeconds(100))
            .slidingWindowSize(1000)
            .permittedNumberOfCallsInHalfOpenState(100)
            .writableStackTraceEnabled(false)
            .automaticTransitionFromOpenToHalfOpenEnabled(true)
            .failureRateThreshold(20f).build();

        CircuitBreakerConfig extendedConfig = from(baseConfig)
            .waitDurationInOpenState(Duration.ofSeconds(20))
            .build();
```
We get Fail. If we remove default value for `waitIntervalFunctionInOpenState` we can loose backwards-compatibility.

I could try something else for the proposed solution... What you think?

RobWin: Yes, we must not loose  backwards-compatibility

RobWin: Maybe we just go back and use an int counter. If the counter == 2, then the user has used both methods.

dlsrb6342: How about two boolean values like `isWaitDurationInOpenStateSet` and `isWaitIntervalFunctionInOpenStateSet`?

RobWin: Counter seems simpler to me :)

RobWin: Maybe we add a waitIntervalBiFunction in the future as well, comparable to Retry.

dlsrb6342: I just worried about set same value twice like following. This is not a normal case.
```
CircuitBreakerConfig.custom()
    .waitDurationInOpenState(Duration.ofSeconds(100))
    .waitDurationInOpenState(Duration.ofSeconds(200))
    .build()
```

If it is ok, counter is also fine to me.

RobWin: Ok, I think we should not check this, because this could be done on any config property. If he overrides the same property, it's the fault of the user ;)

likhoman: Please could you consider combining the two solutions together:

```java
  private IntervalFunction createWaitIntervalFunctionInOpenState() {
            if (setWaitInOpenStateCounter > 1) {
                if (waitDurationInOpenState != null && waitIntervalFunctionInOpenState != null) {
                    throw new IllegalStateException("The waitIntervalFunction was configured twice which could result in an undesired state. " +
                        "Please use either waitIntervalFunctionInOpenState or waitDurationInOpenState.");
                }
            }
            if (waitDurationInOpenState != null) {
                return IntervalFunction.of(waitDurationInOpenState.toMillis());
            }

            return waitIntervalFunctionInOpenState == null
                ? IntervalFunction.of(Duration.ofSeconds(DEFAULT_SLOW_CALL_DURATION_THRESHOLD))
                : waitIntervalFunctionInOpenState;
        }
```
First `setWaitInOpenStateCounter > 1` avoid overridden Second `waitDurationInOpenState != null && waitIntervalFunctionInOpenState != null` avoid case which described @dlsrb6342

likhoman: oh, no my bad(

RobWin: Let's keep it simple. 
`waitIntervalFunctionInOpenState` has a default value.
It can be overwritten, but we check a counter and if counter > 1, then throw an exception in the build method.

likhoman: @RobWin let's focus on the solution with equal 2 or greater than 1?

RobWin: Greater than 1 is fine. 
We can change it to `The waitIntervalFunction was configured multime times which could result in an undesired state.`

likhoman: @RobWin what did you mean `multime`? Maybe you mean `multiple`?

RobWin: Yes, sorry ;)

likhoman: Thanks

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfigTest.java
```diff
@@ -443,6 +443,39 @@ public void testToString() {
...
+    @Test(expected = IllegalStateException.class)
+    public void shouldNotUseWitIntervalFunctionInOpenStateAndWaitDurationInOpenStateTogether() {
+        custom()
+            .waitDurationInOpenState(Duration.ofMillis(3333))
+            .waitIntervalFunctionInOpenState(IntervalFunction.of(Duration.ofMillis(1234)))
+            .build();
+    }
+
+    @Test
+    public void test2() {
```
RobWin, line 455: Could you please rename or delete this test?

RobWin (changes requested):

RobWin (approved):

Issue #1221, opened 2020-10-20
RateLimiter Question/Enhancement: Reduce memory footprint of high cardinality keys


laksnv:
I'd like to rate limit access to an operation that's relatively expensive. My intention is to prevent some customers from over-utilizing my application since that could potentially result in degraded performance for other customers (multi-tenant application).

My application creates a wide key space for each of my customer; about 100 keys on average. I'd like to rate limit on these customer keys rather than on the customer identifier itself. If I have 5,000 customers using my application, this could translate to 5,000 * 100 = 500,000 Rate limiter keys.

Since stale/unused entries in RateLimiter registry aren't cleaned up automatically (please correct my if I'm wrong), the memory footprint becomes significant.

I understand that I can override RateLimiterRegistry and perform cleanup on my side, but I feel this is not an uncommon use-case for a Rate Limiter and would like to get your thoughts on introducing 2 new config keys:
1. `expiryTime` - determines the duration after which each unused rate limiter key will be marked as stale.
2. `cleanupInterval` - determines the rate at which stale keys are cleaned up.

RobWin:
Hi,
simplest solution is to override the RegistryStore. Please see: 
https://resilience4j.readme.io/docs/examples-4#override-the-registrystore

```
RateLimiterRegistry rateLimiterRegistry = RateLimiterRegistry.custom()
  .withRegistryStore(new CacheRateLimiterRegistryStore())
  .build();
```

You can use any Cache implementation to implement the `RegistryStore` interface.

RobWin:
What we could do is provide a default implementation based on [caffeine](https://github.com/ben-manes/caffeine).

laksnv:
Thanks for getting back! Yes, that is exactly what I had in mind. I did the same thing when I integrated with circuit breaker (pre-RegistryStore).

laksnv:
Hi Rob, would you like me to create a PR or is someone already working on this?

RobWin:
You could create a PR. But it should be a new module.

laksnv:
My initial thought was to implement a Caffeine backed registry store in `resilience4j-cache` module, which users can use to build their RateLimiterRegistry.

Are you concerned about adding Caffeine as a dependency in `resilience4j-cache` module?

RobWin:
Yes, the purpose of resilience4j-cache is a different one and just have a dependency to JCache, but no implementation of JCache.

[...]

PR #1225, opened 2020-10-24
Issue #1221: Implement a RegistryStore backed by JCache implementation

laksnv:
https://github.com/resilience4j/resilience4j/issues/1221


resilience4j-cache/src/main/java/io/github/resilience4j/cache/CacheRegistryStore.java
```diff
@@ -0,0 +1,79 @@
...
+    public CacheRegistryStore(Cache<String, E> cacheStore) {
+        this.cacheStore = cacheStore;
+    }
+
+    @Override
+    public E computeIfAbsent(String key, Function<? super String, ? extends E> mappingFunction) {
+        try {
+            return cacheStore.invoke(key, new AtomicComputeProcessor<>(), mappingFunction);
+        } catch (EntryProcessorException e) {
+            throw new RuntimeException(e.getCause());
```
RobWin, line 27: I think we should throw a custom exception:

Instead of wrapping it in a `RuntimeException` or `CacheException`, we could define a `CacheRegistryStoreException`.

RobWin (changes requested):

resilience4j-cache/src/main/java/io/github/resilience4j/cache/CacheRegistryStore.java
laksnv: Thanks for the review!

I didn't create a new type to keep the behavior in sync with the in-memory store that throws underlying ConcurrentHashMap exception type: https://github.com/resilience4j/resilience4j/blob/3dc2eadb654514b5d503317dbf80a776a892931c/resilience4j-core/src/main/java/io/github/resilience4j/core/registry/InMemoryRegistryStore.java#L35

Do you think it makes sense to create a `RegistryStoreException` that can be used by all store implementations?

RobWin: >      * @throws RuntimeException or Error if the remappingFunction does so,
     *         in which case the mapping is unchanged

Yes, you are right. Then let's keep it as it is.

RobWin (approved):

PR #1232, opened 2020-10-28
Add result predicate to CircuitBreaker

RobWin:


resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
```diff
@@ -556,6 +555,16 @@ static CircuitBreaker of(String name,
      */
     void onSuccess(long duration, TimeUnit durationUnit);
```
dlsrb6342, line 556: Doesn't need to deprecate `onSuccess`? I think `onResult` can cover `onSuccess`.

RobWin: I wasn't sure. I don't really like methods where you have to provide a `null` parameter, because it's optional. I prefer to use Method Overloading.
But I thought giving the overloaded method a name like `onResult` explains the intention better than `onSuccess`.

dlsrb6342: Understood 👍

dlsrb6342 (approved):

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java
Romeh: but maybe we still need to add @deprecated so we can move it from version 2 if it is not still needed ?

RobWin: It's still needed. There are some functions which can be decorated and don't return a value. Like Runnable. Or types like Completable, Maybe.

jpdev01:
@RobWin how to add this property in yml/properties? 
currently there is only being able to add `record-failure-predicate`

Issue #1228, opened 2020-10-24
RateLimiter.closePeriod - new method block any new permits for the given period


walec51:
There are situation when despite using a RateLimiter the underlining called service will say that you passed the maximum number of calls for a given period. For example an application that uses RateLimiter for its calls to a REST API can be restated and in consequence be not aware of the number previous calls done to that API. In this case despite having a properly defined RateLimiter config in accordance to the documented REST API call limit quota it can cause and over usage of the API. When that happens it would be good to close the current period in a given rate limiter so that no new permissions will be granted during it.

To achieve this I propose to add two methods:

```
rateLimiter.closePeriod();
```

and a configuration option that would close the period in a decorated call would throw an exception that meets some configured criteria

```
RateLimiterConfig.from(...)
                .limitRefreshPeriod(Duration.ofMinutes(1))
                .limitForPeriod(1200)
                .closePeriodOnException(e -> e instanceof HttpStatusAware && ((HttpStatusAware) e).getHttpStatus() == 429)
                .build()
```

I'm willing to prepare a pull request if this was to be accepted.

RobWin:
Hi,

what do you think about calling it the new method:

`rateLimiter.drainPermissions()`

The method acquires and returns all permits that are available. Like `reservePermission`, but without a "number of permits" parameter.

I wonder if we should allow to the user to configure a result handler (`Either<Throwable,T> result`) so that he can even change other config parameters dynamically, eg. by using `rateLimiter.changeLimitForPeriod(newLimit)`. We could even utilize the existing event publisher and introduce a new `RateLimiterOnResultEvent`.

```
rateLimiter.getEventPublisher().onResult(event -> {
            Either<Throwable, T> result = event.getResult();
            // check result
            if(... ){
                rateLimiterRegistry
                    .rateLimiter(event.getRateLimiterName()).drainPermissions();
            });
        });
```

But indeed a short cut would be nice.

RateLimiterConfig.from(...)
                .limitRefreshPeriod(Duration.ofMinutes(1))
                .limitForPeriod(1200)
                .drainPermissionsOnException(e -> e instanceof HttpStatusAware && ((HttpStatusAware) e).getHttpStatus() == 429)

walec51:
ok, I'll use `drainPermissions` and `drainPermissionsOnResult`

PR #1240, opened 2020-11-08
added RateLimiter.drainPermissions method

walec51:
First part of https://github.com/resilience4j/resilience4j/issues/1228

### DIFF
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/RateLimiterOnDrainEvent.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/RateLimiterOnDrainEvent.java
@@ -18,24 +18,14 @@
  */
 package io.github.resilience4j.ratelimiter.event;
 
-import java.time.ZonedDateTime;
+public class RateLimiterOnDrainEvent extends AbstractRateLimiterEvent {
 
-/**
- * An event which is created by a {@link io.github.resilience4j.ratelimiter.RateLimiter}.
- */
-public interface RateLimiterEvent {
+    public RateLimiterOnDrainEvent(String rateLimiterName, int numberOfPermits) {
+        super(rateLimiterName, numberOfPermits);
+    }
 
-    String getRateLimiterName();
-
-    Type getEventType();
-
-    int getNumberOfPermits();
-
-    ZonedDateTime getCreationTime();
-
-    enum Type {
-        FAILED_ACQUIRE,
-        SUCCESSFUL_ACQUIRE,
-        DRAINED
+    @Override
+    public Type getEventType() {
+        return Type.SUCCESSFUL_ACQUIRE;
     }
 }
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
@@ -20,6 +20,7 @@ package io.github.resilience4j.ratelimiter.internal;
 
 import io.github.resilience4j.ratelimiter.RateLimiter;
 import io.github.resilience4j.ratelimiter.RateLimiterConfig;
+import io.github.resilience4j.ratelimiter.event.RateLimiterOnDrainEvent;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnFailureEvent;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnSuccessEvent;
 import io.vavr.collection.HashMap;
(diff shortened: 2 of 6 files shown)


resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/RateLimiterOnDrainEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class RateLimiterOnDrainEvent extends AbstractRateLimiterEvent {
+
+    public RateLimiterOnDrainEvent(String rateLimiterName, int numberOfPermits) {
+        super(rateLimiterName, numberOfPermits);
+    }
+
+    @Override
+    public Type getEventType() {
+        return Type.SUCCESSFUL_ACQUIRE;
```
RobWin, line 29: DRAINED

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/RateLimiterOnDrainEvent.java
```diff
@@ -0,0 +1,31 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+package io.github.resilience4j.ratelimiter.event;
+
+public class RateLimiterOnDrainEvent extends AbstractRateLimiterEvent {
```
RobWin, line 21: RateLimiterOnDrainedEvent 
wdyt?

RobWin (changes requested):

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/event/RateLimiterOnDrainEvent.java
walec51: fixed

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
```diff
@@ -128,20 +129,34 @@ public long reservePermission(final int permits) {
...
             return modifiedState.nanosToWait;
         }
 
-        publishRateLimiterEvent(false, permits);
+        publishRateLimiterAcquisitionEvent(false, permits);
         return -1;
     }
 
+    @Override
+    public void drainPermissions() {
```
storozhukBM, line 147: Please don't change state directly, but use the same approach as in `updateStateWithBackOff` https://github.com/resilience4j/resilience4j/blob/a2534c515eeb9361b755bbaab2210990f99ac123/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java#L157

storozhukBM (changes requested):

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
walec51: I don't think that is a good idea

to drain the number of permits to zero precisely in a thread safe manner I have to read the number of permits left and subtract that amount in one atomic operation

`updateStateWithBackOff` doesn't do that - explain why using it would be better?

storozhukBM: `updateStateWithBackOff` uses `compareAndSwap` operation, so it is atomic, consistent and will do completely fine.
`getAndUpdate` is also fine with one exception that it will reduce throughput of contended operations on `state` for Java 8 (starting from Java 9 it is fine), for additional details, please read the doc comment on `updateStateWithBackOff` method.

walec51: reimplemented drain using compareAndSet

please check if this impl is ok for you

I also added a unit test for an drain edge case

walec51:
@RobWin I think you broke the master branch https://travis-ci.org/github/resilience4j/resilience4j/builds/743885837

and now my branch is broken as well after resolving a commit

RobWin:
Should be fixed now

walec51:
@RobWin @storozhukBM can we merge this now?

RobWin:
I will do it when I'm back from quarantine

RobWin (approved):

PR #1245, opened 2020-11-19
Prevent the decoration of default methods in feign client

celcius112:
Previously a default method calling an interface method would trigger the decorator twice. For instance two circuit-breaker events would be fired, one when calling the default method and one when the default method calls the actual interface method.

See the feign documentation: https://github.com/OpenFeign/feign#static-and-default-methods

### DIFF
--- a/resilience4j-feign/src/main/java/io/github/resilience4j/feign/FeignDecorators.java
+++ b/resilience4j-feign/src/main/java/io/github/resilience4j/feign/FeignDecorators.java
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright 2019
+ * Copyright 2020
  *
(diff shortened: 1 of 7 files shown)


RobWin:
Seems two tests failed
```
io.github.resilience4j.feign.FeignDecoratorsTest > testWithCircuitBreaker FAILED
    java.lang.NullPointerException at FeignDecoratorsTest.java:48
io.github.resilience4j.feign.FeignDecoratorsTest > testWithRateLimiter FAILED
    java.lang.NullPointerException at FeignDecoratorsTest.java:66
```

celcius112:
@RobWin my bad, I have amended the commit

RobWin (approved):

PR #1264, opened 2020-12-05
rate limiter - drain permissions on result config option

walec51:
rest of the code outlined in https://github.com/resilience4j/resilience4j/issues/1228


walec51:
needs a little more javadoc and test

however I'd like to get the implementation approved before I write them

@RobWin @storozhukBM please take a look when you have the time

RobWin:
Thank you for your PR.

Could you please introduce hooks like `onResult` and `onError` to the RateLimiter interface.
This would be analog to the design of the CircuitBreaker, Bulkhead and Retry. 
See: [CircuitBreaker](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java#L547-L566)

Inside of the hooks we can invoke the `drainPermissionsIfNeeded` check.
In the `drainPermissionsOnResult` function I would prefer to use `<Either<Throwable, Object>` over `CallsResult `, as we use it in [IntervalBiFunction](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalBiFunction.java)

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
```diff
@@ -185,7 +191,14 @@ static RateLimiter ofDefaults(String name) {
     ) {
         return () -> {
             waitForPermission(rateLimiter, permits);
-            return supplier.get();
+            try {
+                F result = supplier.get();
+                drainIfNeeded(rateLimiter, CallsResult.success(result));
```
RobWin, line 196: Please introduce and use an `onResult` hook

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
```diff
@@ -185,7 +191,14 @@ static RateLimiter ofDefaults(String name) {
...
+                return result;
+            } catch (Exception exception) {
+                drainIfNeeded(rateLimiter, CallsResult.failure(exception));
```
RobWin, line 199: Please introduce and use an `onError` hook

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java
```diff
@@ -161,6 +173,21 @@ public Builder writableStackTraceEnabled(boolean writableStackTraceEnabled) {
...
+         * Allows you to check the result of a call decorated by this rate limiter and make a decision
+         * should we drain all the permissions left it the current period. Useful in situations when
+         * despite using a RateLimiter the underlining called service will say that you passed the maximum number of calls for a given period.
+         *
+         * @param drainPermissionsOnResult your function should return true when the permissions drain
+         *                                 should happen
+         * @return the RateLimiterConfig.Builder
+         * @see RateLimiter#drainPermissions()
+         */
+        public Builder drainPermissionsOnResult(Function<CallsResult, Boolean> drainPermissionsOnResult) {
```
RobWin, line 186: Please use `Either<Throwable, Object>` instead of `CallsResult`

RobWin (changes requested):

amoffet:
The response to reviews on this pull request seems to be stalling...  The release of 1.7.0 seems to be predicated upon it.  Any progress?

walec51:
@amoffet I'll polish this PR on the weekend

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java
walec51: are you sure? CallsResult has a nicer API which enables cleaner configuration 

please take a look at my unit test:

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/RateLimiterWithConditionalDrainIntegrationTest.java

the configuration part would not look so simple if I used `Either<Throwable, Object>`

RobWin: Yes, I would prefer to use Either for now.
We are working on a 2.0 branch which uses Java 11 and has no Vavr dependency anymore. We will replace Either by our own implementation at many places.  This would be the best time to introduce a simpler API.

earce:
@walec51 wanted to bump this back up

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java
walec51: I've tried refactoring the code to `Either` but `Either` cannot handle a successful call return of a function that returns `void`

this is needed in situations like

```
    static CheckedRunnable decorateCheckedRunnable(RateLimiter rateLimiter, int permits,
        CheckedRunnable runnable) {

        return () -> {
            waitForPermission(rateLimiter, permits);
            try {
                runnable.run();
                drainIfNeeded(rateLimiter, CallsResult.success());
            } catch (Exception exception) {
                drainIfNeeded(rateLimiter, CallsResult.failure(exception));
                throw exception;
            }
        };
    }
```

what should I use here? `Either.right(null)` ?

and then in the application code which check for a certain condition should I do this:

```
if (result.isRigth() && result.get() != null) {
```

?

besides that the API started looking ugly with template mayhem such as:

```
private final Function<Either<? extends Throwable, Object>, Boolean> drainPermissionsOnResult;
```

I do not think this is the right approach.

Furthermore where should I put helper methods like `isSuccessfulAndReturned`:

```
        RateLimiter limiter = RateLimiter.of("someLimiter", RateLimiterConfig.custom()
            .limitForPeriod(5)
            .limitRefreshPeriod(Duration.ofHours(1))
            .drainPermissionsOnResult(
                callsResult -> callsResult.isSuccessfulAndReturned(
                    ResponseWithOverloadIndication.class,
                    response -> response.isPotentialOverload()))
            .build());
```

?

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
walec51: where should I introduce it? in `RateLimiterEventProcessor` ?

if so then how should I pass the event to it from a static method? event processors are privet fields in rate limiters implementations

RobWin: See https://github.com/resilience4j/resilience4j/pull/1264#issuecomment-739824598

I'm talking about new public methods in the `RateLimiter` interface analog to the design of the CircuitBreaker, Bulkhead and Retry.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java
RobWin: We can add a `ResultUtil` class.

```
ResultUtil.isSuccessfulAndReturned(eitherResult, responseWithOverloadIndication.class,  response -> response.isPotentialOverload())
```

As I said, when we remove Vavr from Resilience4j 2.0, we will have a look at the whole code base and might introduce a call like you suggest. But then we will use it everywhere and consistently. 
In v1.x, we are using `Either` when a user needs to check if the call was a sucess or failure.

In v1.x anew `ResultUtil` helper class could also be utilized in `IntervalBiFunction`.

If you don't like it, as an alternative we could introduce:

`RateLimiterConfig.drainPermissionsOnResult` and `RateLimiterConfig.drainPermissionsOnException` comparable to `RetryConfig.retryOnException` and `RetryConfig.retryOnResult`.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
walec51: done

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java
walec51: removed `CallsResult` introduced `Either`

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
walec51: done

RobWin (changes requested):

RobWin:
Thank you. 
PR looks fine now.

RobWin:
Ok, everything looks fine now.
Now we need to enhance Reactor  `RetyOperator` and RxJava2&3 `RateLimiterTransformer`

walec51:
> Now we need to enhance Reactor `RateLimiterOperator` and RxJava2&3 `RateLimiterTransformer`

shouldn't this be a topic of a separate PR?

RobWin:
We can create another PR, but usually when we introduce a new feature we make sure that all decorators and also Reactor, RxJava2 and RxJava3 are not to be forgotten.

RobWin (approved):

PR #1286, opened 2020-12-28
Retry failing tests up to 3 times

joschi:
The resilience4j test suite has been very flaky for me and I never got it to complete without failing tests on my machine (current `master` branch).

Since the failing tests were different in every run and were mostly related to timing problems, I think introducing the Test Retry Gradle plugin makes sense in this case.

References:
- https://blog.gradle.org/gradle-flaky-test-retry-plugin
- https://github.com/gradle/test-retry-gradle-plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,6 +79,7 @@ configure(project.coreProjects) {
         retry {
             maxRetries = 3
             maxFailures = 20
+            failOnPassedAfterRetry = true
         }
     }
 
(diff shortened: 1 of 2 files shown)


joschi:
With the latest change (11075ae41e7d33a479647e18235dec7f1420a992), builds will fail when there are flaky tests.

Gradle test report with flaky test

RobWin:
Komisch. Der Test dürfte niemals mehr als 1 Minute dauern O_O

RobWin (approved):

RobWin:
thx

PR #1291, opened 2021-01-04
Removed vavr dependency from latest master commits to branch 2.0

KrnSaurabh:
pulled latest commits from master to branch 2.0
Removed Vavr dependency introduced in latest commits from master.

@RobWin  Kindly review. I need to manually merge this PR to 2.0 branch since it has conflicts.


resilience4j-core/src/main/java/io/github/resilience4j/core/functions/Either.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.function.Function;
+
+/*
+ * Either represents a value of two possible types. An Either is either a {@link Left} or a
+ * {@link Right}.
+ * This interface is similar to {@link io.vavr.control.Either}.
+ */
+public interface Either<L, R> {
```
KrnSaurabh, line 31: @RobWin I have created our custom `Either` that replaces `vavr Either`. This class has some utility methods taken from `vavr Either` . Let me know if this approach is fine.

PR #1293, opened 2021-01-07
Issue #1281 Added Retry.failAfterMaxAttempts

### LINKED ISSUE #1281: [Idea] Retry - throw on max retries reached
Resilience4j version: `1.6.1`

Java version: `openjdk 15 2020-09-15`

**Here's a feature suggestion which I would gladly open a PR for:**
Whenever the `Retry` logic reaches it's configured `maxAttempts`, it would be nice for a configurability option to throw an exception, instead of returning the last result back to the caller.

I found it unintuitive to configure a `retryOnResult`, only to find out that after maxAttempt was reached, the result was still being returned. In my mind, if you consider a certain result an error to be retried, it doesn't suddenly become a non-error just cause you ran out of retries :)

This probably depends on the use-case though, but having some kind of `throwOnCompletion` or `throwOnMaxFailedRetries` configuration would be awesome.

Here's my use case:

```java
final Retry retry = Retry.of("clientCall", RetryConfig
    .<HttpResponse<String>>custom()
    .retryOnException(e -> e instanceof IOException)
    .retryOnResult(response -> response.statusCode() == 503)
    .build()
);

final HttpResponse<String> resp = retry.executeCheckedSupplier(() -> client.send(
    request, HttpResponse.BodyHandlers.ofString()
));

return resp.body();
```

`resp` would still be the last attempts response, even though `retryOnResult` would have kicked in.

My suggestion would be to config it up like:
```java
final Retry retry = Retry.of("clientCall", RetryConfig
    .<HttpResponse<String>>custom()
    .retryOnException(e -> e instanceof IOException)
    .retryOnResult(response -> response.statusCode() == 503)
->  .throwOnMaxFailedRetries(response -> new RuntimeException("Call failed, retries exhausted. status=" + response.statusCode()))
    .build()
);
```

Which would make `retry.executeCheckedSupplier()` throw the configured exception, if the result still matches the `retryPredicate`. 

What do you think about this?

izrobin:
Adds property `failAfterMaxAttempts` (boolean) to the `RetryConfig`.
This property will be inspected on the `Retry.Context :: onComplete()`, and if found to be `true`, it will throw a `MaxRetriesExceeded` exception.

If `false`, behaviour should stay unaffected.

One more notable change made in this PR is the switch of calls in `AsyncRetryBlock :: onResult`.
This to make it more in-line with the synchronous result checking. Plus the fact that we can't complete our promise/future before we inspect if we would receive a `MaxRetriesExceeded` exception from `onComplete`.

The difference here is that we'll publish events before resolving our promise/future. Do you see that to be a problem? Happy to discuss!

Lastly, I bumped the `assertj` dependency, in order to be able to call
```java
CompletionStage.failsWithin(5, TimeUnit.SECONDS)
    .withThrowableOfType(ExecutionException.class)
```

Closes #1281 

Also: Let me know if I could help out updating the readme / user guide somehow. Couldn't find any documentation (other than javadoc) in this repo

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
@@ -678,8 +678,12 @@ public interface Retry {
             final long delay = retryContext.onResult(result);
 
             if (delay < 1) {
-                promise.complete(result);
-                retryContext.onComplete();
+                try {
+                    retryContext.onComplete();
+                    promise.complete(result);
+                } catch (Exception e) {
+                    promise.completeExceptionally(e);
+                }
             } else {
                 scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
             }
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
@@ -20,11 +20,12 @@ package io.github.resilience4j.retry.internal;
 
 import io.github.resilience4j.core.EventConsumer;
 import io.github.resilience4j.core.EventProcessor;
+import io.github.resilience4j.core.IntervalBiFunction;
 import io.github.resilience4j.core.lang.Nullable;
 import io.github.resilience4j.retry.MaxRetriesExceeded;
+import io.github.resilience4j.retry.MaxRetriesExceededException;
 import io.github.resilience4j.retry.Retry;
 import io.github.resilience4j.retry.RetryConfig;
-import io.github.resilience4j.core.IntervalBiFunction;
 import io.github.resilience4j.retry.event.*;
 import io.vavr.CheckedConsumer;
 import io.vavr.collection.HashMap;
@@ -162,15 +163,15 @@ public class RetryImpl<T> implements Retry {
                 if (currentNumOfAttempts >= maxAttempts) {
                     failedAfterRetryCounter.increment();
                     Throwable throwable = Option.of(lastException.get())
-                        .getOrElse(lastRuntimeException.get());
-                    RuntimeException maxRetriesExceeded = new MaxRetriesExceeded(
-                        "Max retries reached for retry=" + name
-                    );
-                    publishRetryEvent(() -> new RetryOnErrorEvent(name, currentNumOfAttempts,
-                        throwable != null && !failAfterMaxAttempts ? throwable : maxRetriesExceeded));
+                        .orElse(Option.of(lastRuntimeException.get()))
+                        .filter(p -> !failAfterMaxAttempts)
+                        .getOrElse(new MaxRetriesExceeded(
+                            "max retries is reached out for the result predicate check"
+                        ));
+                    publishRetryEvent(() -> new RetryOnErrorEvent(name, currentNumOfAttempts, throwable));
 
                     if (failAfterMaxAttempts) {
-                        throw maxRetriesExceeded;
+                        throw MaxRetriesExceededException.createMaxRetriesExceededException(RetryImpl.this);
                     }
                 } else {
                     succeededWithoutRetryCounter.increment();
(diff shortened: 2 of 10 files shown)


resilience4j-retry/src/test/java/io/github/resilience4j/retry/internal/CompletionStageRetryTest.java
```diff
@@ -141,25 +139,30 @@ public void shouldRetryInCaseOfAnExceptionAtAsyncStage() {
         assertThat(result).isEqualTo("Hello world");
     }
 
-    private void shouldCompleteFutureAfterAttemptsInCaseOfExceptionAtSyncStage(int noOfAttempts) {
```
izrobin, line 144: This method was unused

RobWin:
Thank you. I will review it.
Documentation resides here: https://resilience4j.readme.io/

The page https://resilience4j.readme.io/docs/retry has a `SUGGEST EDITS` button which allows you to contribute documentation enhancements.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -161,9 +163,15 @@ public void onComplete() {
                     failedAfterRetryCounter.increment();
                     Throwable throwable = Option.of(lastException.get())
                         .getOrElse(lastRuntimeException.get());
+                    RuntimeException maxRetriesExceeded = new MaxRetriesExceeded(
```
RobWin, line 166: I think we should create a new exception `MaxRetriesExceededException` similar  to `CallNotPermittedException` or `BulkheadFullException`, if we really throw it as part of the Public API now.

It should also store the causing retry name as a field similar to the `CallNotPermittedException` and provide a getter method. 
I guess we also need a `writableStackTrace` config property so that users can disable the stacktrace which is sometimes filling the log files too much.

RobWin (changes requested):
Please also have a look at the Reactor `RetryOperator` and Rxjava 2 and 3 `RetryTransformer`.
Might be a little bit more tricky, if you don't know reactive streams. Ping me if you need help.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
izrobin: Sure, that sounds fair.
What to do with the similarly named `MaxRetriesExceeded` ? 

Also, (I maybe have understood this wrong), but if `throwable` on the row above is not null, (eg. if attempt 2 out of 3 got an ignorable exception), we'll still send it in the `MaxRetriesExceeded` event, even though we failed because we reached maximum nbr of attempts. That seems like a bug

izrobin:
>Please also have a look at the Reactor RetryOperator and Rxjava 2 and 3 RetryTransformer.

All of these should have been handled by the change in `Retry.AsyncContext`? Or am I misunderstanding something? I even have tests for them. You're right, I don't have any experience with reactive streams but I know asynchronous coding

RobWin:
Ture, I missed the RetryTransfomerTests. Sry.
Could you please add a test to RetryOperatorTest

izrobin:
Hm, the tests that failed on Travis works locally. Are they flakey? Could you re-trigger a build?

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
RobWin: I think we have to keep `MaxRetriesExceeded`  for backward compatibility :( But we shouldn't reuse the exception.

Regarding line 164:
1) Let's assume the first call fails because of an exception. And 2nd, 3rd calls fail because of the result. In that case the `RetryOnErrorEvent` would contain the last thrown exception instead of `MaxRetriesExceeded`. It seems to be a bug. You are right.

1) Let's assume the first call fails because of the result. And 2nd, 3rd calls fail because of an exception. In that case the `RetryOnErrorEvent` should contain the last thrown exception.

izrobin: Regarding point 1. We wouldn't even reach the `onComplete` method, unless the last retry attempt return a non-exceptional result. Sooo, I'm not sure what the `lastRuntimeException` is supposed to mean. But let's keep it as-is.

I agree on your last statement, let's make it so to be less confusing.

izrobin:
Ok, ready for a second review @RobWin ! 🙏

resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
```diff
@@ -678,8 +678,12 @@ private void onResult(T result) {
...
 
             if (delay < 1) {
-                promise.complete(result);
-                retryContext.onComplete();
+                try {
+                    retryContext.onComplete();
+                    promise.complete(result);
+                } catch (Exception e) {
+                    promise.completeExceptionally(e);
+                }
```
izrobin, line 686: This is the only change introduced by this PR (Regardless of if the new `failAfterMaxAttempts` is set). Let me know if you see any issues with this

RobWin:
I will review on monday. Thank your for your effort.

RobWin (approved):

RobWin:
Thank you for this PR

izrobin:
Thank you @RobWin ! 🙏

izrobin:
I'll update the documentation as soon as we have a new release.

PR #1300, opened 2021-01-18
Issue #1228:  Added RateLimiter drainIfNeeded to RxJava3 (#1264)

hexmind:


resilience4j-rxjava3/src/main/java/io/github/resilience4j/rxjava3/ratelimiter/operator/FlowableRateLimiter.java
```diff
@@ -62,8 +63,13 @@ protected void subscribeActual(Subscriber<? super T> downstream) {
         }
 
         @Override
-        public void hookOnError(Throwable t) {
-            // NoOp
+        protected void hookOnError(Throwable e) {
+            rateLimiter.drainIfNeeded(Either.left(e));
```
RobWin, line 67: Please use `rateLimiter.onError()` instead

resilience4j-rxjava3/src/main/java/io/github/resilience4j/rxjava3/ratelimiter/operator/FlowableRateLimiter.java
```diff
@@ -62,8 +63,13 @@ protected void subscribeActual(Subscriber<? super T> downstream) {
...
+        }
+
+        @Override
+        protected void hookOnNext(T value) {
+            rateLimiter.drainIfNeeded(Either.right(value));
```
RobWin, line 72: Please use `rateLimiter.onResult(value)` instead

RobWin (changes requested):

resilience4j-rxjava3/src/main/java/io/github/resilience4j/rxjava3/ratelimiter/operator/FlowableRateLimiter.java
RobWin: drainIfNeeded must not be part of the public API. We have to fix it

hexmind: > Since java 9, you will be able to add private methods and private static method in interfaces.
What we can do before migration to Java9+ is `drainIfNeeded ` deprecation* or delegation:
```
    default void onResult(Object result) {
        DrainPermissionsOnResult.drainIfNeeded(this, Either.right(result));
    }
```

We can also add `void drainPermissions(Predficate<Either<? extends Throwable, ?>> drainPermissionsOnResult)`
but it an ugly replacing of drainIfNeeded method by new one.


*) 
I am ok with a deprecation (making private in the future):
```
@Deprecated(forRemoval = true)
default void drainIfNeeded(Either<? extends Throwable, ?> callsResult) {
```

RobWin: We can make it private now. It's a new method and not part of any release.

hexmind: We cannot. It is why I mentioned Java9+ feature:
`public interface RateLimiter`
`targetCompatibility = "1.8"`

> Since java 9, you will be able to add private methods and private static method in interfaces.

RobWin (approved):

PR #1306, opened 2021-01-21
Adaptive Bulkhead 2 Draft

hexmind:


resilience4j-bulkhead/src/test/java/io/github/resilience4j/bulkhead/adaptive/internal/amid/LogEntry.java
```diff
@@ -7,6 +7,7 @@
 /**
  * @author romeh
  */
+@Deprecated
 public class LogEntry implements Comparable<LogEntry> {
```
RobWin, line 11: Do we still need this class?

RobWin (approved):

PR #1321, opened 2021-01-29
Add name to configs, allow inheritance of RetryConfig and CircuitBreakerConfig

fededonna:


Romeh:
@fededonna  the build is failing on this PR , u had a chance to check why ?

RobWin:
Hi,
why do we need config names?

RobWin:
RetryConfigBuilderTest is failing

PR #1339, opened 2021-02-18
Issue #1334 Fix bug instance properties with base config

### LINKED ISSUE #1334: Retry max attempts configuration not resolved from base config
Base configuration of max attemps is not resolved properly from base config here is an example
default-config has configured max attempts = 4
all-exceptions-config extends default-config and overrides only retryExceptions
maxAttempts configuration is not taken from baseConfig and it's resolved to 3, which is default

I did not test other configuration properties

configuration: 
```
resilience4j.retry:
  configs:
    default-config:
      maxAttempts: 4
      waitDuration: 100ms
      retryExceptionPredicate: io.github.resilience4j.circuitbreaker.RecordFailurePredicate
      retryExceptions:
        - java.io.IOException

    all-exceptions-config:
      baseConfig: default-config
      retryExceptions:
        - java.lang.Exception

  backends:
      retryBackendE:
         baseConfig: all-exceptions-config
```

following assertion fails on max attempt check
```
Retry overriddenException = retryRegistry.retry("retryBackendE");
assertThat(overriddenException.getRetryConfig().getMaxAttempts()).isEqualTo(4);
assertThat(overriddenException.getRetryConfig().getExceptionPredicate().test(new UnsupportedOperationException()))
            .isTrue();
```

dlsrb6342:
* Fix #1334 
* Divide `ThreadPoolBulkheadConfigurationPropertiesTest` from `BulkheadConfigurationPropertiesTest`

### DIFF
--- a/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/bulkhead/configuration/BulkheadConfigurationProperties.java
+++ b/resilience4j-framework-common/src/main/java/io/github/resilience4j/common/bulkhead/configuration/BulkheadConfigurationProperties.java
@@ -53,7 +53,7 @@ public class BulkheadConfigurationProperties extends CommonProperties {
         CompositeCustomizer<BulkheadConfigCustomizer> compositeBulkheadCustomizer,
         String instanceName) {
         ConfigUtils.mergePropertiesIfAny(baseProperties, instanceProperties);
-        BulkheadConfig baseConfig = buildBulkheadConfig(BulkheadConfig.custom(), baseProperties,
+        BulkheadConfig baseConfig = createBulkheadConfig(baseProperties,
             compositeBulkheadCustomizer, instanceName);
         return buildBulkheadConfig(BulkheadConfig.from(baseConfig), instanceProperties,
             compositeBulkheadCustomizer, instanceName);
--- a/resilience4j-framework-common/src/test/java/io/github/resilience4j/common/bulkhead/configuration/ThreadPoolBulkheadConfigurationPropertiesTest.java
+++ b/resilience4j-framework-common/src/test/java/io/github/resilience4j/common/bulkhead/configuration/ThreadPoolBulkheadConfigurationPropertiesTest.java
@@ -80,33 +80,28 @@ public class ThreadPoolBulkheadConfigurationPropertiesTest  {
 
         //When
         //Then
-        try {
-            assertThat(bulkheadConfigurationProperties.getBackends().size()).isEqualTo(2);
-            // Should get default config and core number
-            ThreadPoolBulkheadConfig bulkhead1 = bulkheadConfigurationProperties
-                .createThreadPoolBulkheadConfig("backendWithDefaultConfig",
-                    compositeThreadPoolBulkheadCustomizer());
-            assertThat(bulkhead1).isNotNull();
-            assertThat(bulkhead1.getCoreThreadPoolSize()).isEqualTo(3);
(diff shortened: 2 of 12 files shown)


Landrev (approved):
Good job mate.

swalendzik (approved):

resilience4j-framework-common/src/test/java/io/github/resilience4j/common/bulkhead/configuration/ThreadPoolBulkheadConfigurationPropertiesTest.java
```diff
@@ -0,0 +1,151 @@
...
+            assertThat(bulkhead2.getQueueCapacity()).isEqualTo(2);
+            // Unknown backend should get default config of Registry
+            ThreadPoolBulkheadConfig bulkhead3 = bulkheadConfigurationProperties
+                .createThreadPoolBulkheadConfig("unknownBackend",
+                    compositeThreadPoolBulkheadCustomizer());
+            assertThat(bulkhead3).isNotNull();
+            assertThat(bulkhead3.getCoreThreadPoolSize())
+                .isEqualTo(ThreadPoolBulkheadConfig.DEFAULT_CORE_THREAD_POOL_SIZE);
+        } catch (Exception e) {
+            System.out.println(
```
Landrev, line 107: I think that log instead of System.out.println would be more elegant.

dlsrb6342: I removed try catch. 
Thank you.

resilience4j-framework-common/src/main/java/io/github/resilience4j/common/bulkhead/configuration/BulkheadConfigurationProperties.java
```diff
@@ -53,7 +53,7 @@ private BulkheadConfig buildConfigFromBaseConfig(InstanceProperties baseProperti
         CompositeCustomizer<BulkheadConfigCustomizer> compositeBulkheadCustomizer,
         String instanceName) {
         ConfigUtils.mergePropertiesIfAny(baseProperties, instanceProperties);
-        BulkheadConfig baseConfig = buildBulkheadConfig(BulkheadConfig.custom(), baseProperties,
+        BulkheadConfig baseConfig = createBulkheadConfig(baseProperties,
```
Romeh, line 56: this will not create a loop between createBulkheadConfig() and buildConfigFromBaseConfig() ? as inside buildConfigFromBaseConfig()  u call createBulkheadConfig() and inside createBulkheadConfig() u can  buildConfigFromBaseConfig() if there is base configuration ?

dlsrb6342: Yes. There is a loop between `buildConfigFromBaseConfig()`and `createBulkheadConfig()`. 
Everytime we build BulkheadConfig, Need to check if there is `baseConfig`.

Romeh: Then when it will stop that loop of checks ?

dlsrb6342: If there is no base config in instance properties, it will stop.

Issue #1138, opened 2020-08-25
Add a Refill Based Rate Limiter implementation


gkatzioura:
Resilience4j version: 1.3.0

This is a feature proposal for a rate Limiter that issues permissions gradually based on the time that has passed within a period.
The Rate Limiter has a max capacity of permits refilled periodically. Each period has duration in nanoseconds. On a permit request the permits that should have been replenished are calculated based on the nanos passed and the ratio of nanos per period divided by permissions per period.Also this rate limiter has a burst capacity (max capacity). Thus this rate limiter instead of giving permissions based on the amount of permissions per period, gives permissions based on a permission replenish rate (similar to a token bucket).

PR #1371, opened 2021-03-14
Feature/rate limiter registry

gkatzioura:
This is a pull request for the ticket #1138

As discussed on #1139 this pull request is only for the Rate Limiter registry. Thus it contains only the registry mechanism as well as some inheritance to create atomic based rate limiters.


RobWin:
Thx. 
Sry I was on sick leave for 4 weeks. I will have a look at the PR.

Issue #1368, opened 2021-03-10
How to change the RateLimiter limitRefreshPeriod？


ZhuoSiChen:
Resilience4j version: 1.7.0

Java version: 8
As the titile。
What I had find is io.github.resilience4j.ratelimiter.RateLimiter#changeTimeoutDuration 
and 
io.github.resilience4j.ratelimiter.RateLimiter#changeLimitForPeriod
but didn’t find the method to io.github.resilience4j.ratelimiter.RateLimiter#limitRefreshPeriod

dlsrb6342:
@ZhuoSiChen 
I think we can add a method. 
@RobWin @Romeh wdyt?

Romeh:
fine by me @RobWin  ?

storozhukBM:
@ZhuoSiChen  can you please explain why do you want to change limitRefreshPeriod in runtime and what is your use case?

ZhuoSiChen:
Because the people had the day and night,
I was write a offline program and I want to larger the RateLimiter on the night。

storozhukBM:
@ZhuoSiChen can you provide your exact settings of limit and period for night and day?

ZhuoSiChen:
```java
		RateLimiter night = RateLimiter.of("night",
		                                   RateLimiterConfig.custom()
		                                                    .limitRefreshPeriod(Duration.ofSeconds(
				                                                    1))
		                                                    .limitForPeriod(5)
		                                                    .timeoutDuration(Duration.ofMinutes(
				                                                    1))
		                                                    .build());
		RateLimiter day = RateLimiter.of("day",
		                                 RateLimiterConfig.custom()
		                                                  .limitRefreshPeriod(Duration.ofSeconds(
				                                                  2))
		                                                  .limitForPeriod(3)
		                                                  .timeoutDuration(Duration.ofMinutes(
				                                                  1))
		                                                  .build());
```

RobWin:
Better use the RateLimiterRegistry in your code to get an RateLimiter instance.
Then you could replace the ratelimiter instance in the RateLimiterRegistry whenever it's necessary.

ZhuoSiChen:
I was used like this
```
 Flux.just(response)
                .transformDeferred(RateLimiterOperator.of(rateLimiter)));
```

RobWin:
Use our Spring Boot starter and inject the RateLimiterRegistry bean, or create a Singleton RateLimiterRegistry.

```
[...]

PR #1386, opened 2021-03-23
add changeLimitRefreshPeriod method for RateLimiter

ZhuoSiChen:
#1368

### DIFF
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
@@ -28,6 +28,7 @@ import io.vavr.collection.Map;
 
 import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.UnaryOperator;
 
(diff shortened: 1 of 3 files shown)


resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
```diff
@@ -100,15 +101,30 @@ public void changeLimitForPeriod(final int limitForPeriod) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void changeLimitRefreshPeriod(Duration refreshPeriod) {
+        RateLimiterConfig newConfig = RateLimiterConfig.from(state.get().config)
+            .limitRefreshPeriod(refreshPeriod)
+            .build();
+        nanoTimeStart.compareAndSet(nanoTimeStart.get(), nanoTime());
+        state.updateAndGet(currentState -> new State(
```
storozhukBM, line 113: These two consecutive atomic operations aren't actually atomic on their own, and this can lead to inconsistencies of calculations that happen between these operations. I'd try to avoid changing nanoTimeStart altogether.

storozhukBM (changes requested):

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java
ZhuoSiChen: I think the nanoTimeStart is part of the state.
wdyt

dlsrb6342:
Decided not to add method

Issue #1414, opened 2021-05-02
SpEL expressions should support passing runtime method arguments to bean methods


masch712:
Resilience4j version: 1.7.0

Java version: 11.0.6

I want my Bulkhead name to be the return value of passing the @Bulkhead-annotated method's runtime arguments to a Spring bean method.  Currently, the SpelResolver class does not support this.

For example, consider this Spring bean:
```java
@Service
public class BackendService {

  @Autowired
  BulkheadService bulkheadService;

  @Bulkhead(name="@bulkheadService.getBulkheadNameFor(#target)")
  public String doSomething(String target) {
    return "did something to " + target;
  }
}
```

If the `doSomething` method were called at runtime with target `github-api`,  I'd expect the resolved Bulkhead name to be the return value of `bulkheadService.getBulkheadNameFor("github-api")`.

Here's a jUnit test that for this fails when run in the resilience4j repo:

First, define a DummySpelBean interface:
```java
public interface DummySpelBean {
    String getBulkheadName(String param1);
}
```

The test:
```java
@RunWith(SpringJUnit4ClassRunner.class)
@SpringBootTest(classes = TestApplication.class, properties = "property=backend")
public class SpelResolverBeanMethodTest {
    private SpelResolver sut;

    @Autowired
    private ConfigurableBeanFactory configurableBeanFactory;

    @MockBean(name="dummySpelBean")
    DummySpelBean dummySpelBean;

    @Before
    public void setUp() {
        sut = new SpelResolver(new SpelExpressionParser(), new StandardReflectionParameterNameDiscoverer(), configurableBeanFactory);
        sut.setEmbeddedValueResolver(new EmbeddedValueResolver(configurableBeanFactory));
    }
    @Test
    public void beanMethodSpelTest() throws Exception {
        String testExpression = "@dummySpelBean.getBulkheadName(#parameter)}";
        String testMethodArg = "argg";
        String bulkheadName = "sgt. bulko";

        SpelResolverTest target = new SpelResolverTest();
        Method testMethod = target.getClass().getMethod("testMethod", String.class);

        given(dummySpelBean.getBulkheadName(testMethodArg)).willReturn(bulkheadName);

        String result = sut.resolve(testMethod, new Object[]{testMethodArg}, testExpression);

        then(dummySpelBean).should(times(1)).getBulkheadName(testMethodArg);
        assertThat(result).isEqualTo(bulkheadName);
    }
}

```

[...]

PR #1415, opened 2021-05-02
Issue #1414: Add support for calling bean methods with args from the annotated method

masch712:
I also extract a simple interface from SpelResolver to enable library consumers to cleanly define their own SpelResolver bean if they want to override the DefaultSpelResolver


RobWin (approved):

RobWin:
Thanks.
PR is ready?

masch712:
@RobWin yes, this PR is ready for review

PR #1429, opened 2021-05-13
Issue #1428: changes time limiter exception handling

### LINKED ISSUE #1428: Timelimiter for suspendFunction is throwing timeout exception even when the coroutine was cancelled due to a different reason
Resilience4j version: 1.7.0

Java version: 11

Timelimiter for suspendFunction is throwing timeout exception even when the coroutine was cancelled due to different reason.

Suppose due to some business logic, we need to cancel a particular "time limited" suspend function using something like `Defferred.cancel(BusinessCancellationException("Do not require this computation anymore"))`, then this should not be treated as a failure for the timelimiter and should not be converted to a timeout exception but currently it does.

code here converts all cancellation exceptions to timeoutExceptions:

```kotlin

 catch (t: Throwable) {
        if (isCancellation(coroutineContext, t)) {
            val timeoutException = TimeLimiter.createdTimeoutExceptionWithName(name, t)
            onError(timeoutException)
            throw timeoutException
        }

        onError(t)
        throw t
    }
```

I would propose to specifically do this only for TimeoutCancellationException.

piyushmor:
Signed-off-by: Piyush Mor <piyush.mor@gmail.com>

fixes #1428

### DIFF
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/timelimiter/TimeLimiter.kt
@@ -47,12 +47,15 @@ suspend fun <T> TimeLimiter.executeSuspendFunction(block: suspend () -> T): T =
         }
     } catch (t: Throwable) {
         if (isCancellation(coroutineContext, t)) {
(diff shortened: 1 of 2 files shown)


piyushmor:
@RobWin can you please check this?

RobWin (approved):

Issue #1485, opened 2021-07-05
Resilience4j-rxjava2 is needed when using reactor


LenaKucharczuk:
Resilience4j version: 1.7.1

Java version: 11

I'm using rate limiter in micronaut app with reactor. I included resilience4j-reactor but I get the error below when I don't include resilience4j-rxjava2 in my dependencies too (which contains io.github.resilience4j.ratelimiter.operator.RateLimiterOperator class). 
I adjusted the micronaut demo to reproduce the problem: [link](https://github.com/LenaKucharczuk/resilience4j-micronaut-demo/tree/reactor-needs-rxjava)

I assumed that resilience4j-rxjava2 shouldn't be needed when I use reactor. Is it a problem with my configuration or a bug?

```
19:20:30.108 [default-nioEventLoopGroup-1-2] ERROR i.m.h.s.netty.RoutingInBoundHandler - Unexpected error occurred: io/github/resilience4j/ratelimiter/operator/RateLimiterOperator
java.lang.NoClassDefFoundError: io/github/resilience4j/ratelimiter/operator/RateLimiterOperator
	at io.github.resilience4j.micronaut.ratelimiter.RateLimiterInterceptor.intercept(RateLimiterInterceptor.java:84)
	at io.micronaut.aop.chain.MethodInterceptorChain.proceed(MethodInterceptorChain.java:96)
	at resilience4j.micronaut.demo.service.$BusinessAServiceDefinition$Intercepted.fluxSuccess(Unknown Source)
	at resilience4j.micronaut.demo.controller.BackendAController.fluxSuccess(BackendAController.java:35)
	at resilience4j.micronaut.demo.controller.$BackendAControllerDefinition$$exec3.invokeInternal(Unknown Source)
	at io.micronaut.context.AbstractExecutableMethod.invoke(AbstractExecutableMethod.java:151)
	at io.micronaut.context.DefaultBeanContext$4.invoke(DefaultBeanContext.java:482)
	at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:255)
	at io.micronaut.web.router.RouteMatch.execute(RouteMatch.java:118)
	at io.micronaut.http.server.netty.RoutingInBoundHandler.lambda$buildResultEmitter$12(RoutingInBoundHandler.java:1406)
	at io.reactivex.internal.operators.flowable.FlowableDefer.subscribeActual(FlowableDefer.java:35)
[...]

PR #1489, opened 2021-07-09
feature: add publisher extension to support rxJava2/rxJava3/Reactor for micronaut

pollend:
still need to do some more testing but this should be a good start to get this mostly working.  the only thing I'm not entirly sure about is this fallbackPublisher that I have was seeing how I could avoid the extra wrapping around the method. 

any clue how I can test this across these variations? any gradle configuration magic to configure this for the different test.

here are the changes to address the comments here: https://github.com/resilience4j/resilience4j/issues/1485

### DIFF
--- a/resilience4j-micronaut/build.gradle
+++ b/resilience4j-micronaut/build.gradle
@@ -16,11 +16,18 @@ dependencies {
     implementation "io.micronaut:micronaut-validation"
     implementation "io.micronaut:micronaut-router"
 
+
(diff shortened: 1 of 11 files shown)


RobWin:
Is this PR ready to be reviewed?

pollend:
Yes

pollend:
is there anything outstanding I need to resolve @RobWin

RobWin (approved):

PR #1499, opened 2021-07-16
Add new `active.thread.count` metric

### LINKED ISSUE #1496: Available (or currently executing) threads metric from ThreadPoolBulkhead
Resilience4j version:
1.7.1

Java version:
11

We have a few places where we are using the ThreadPoolBulkhead pattern and export metrics using Micrometer to Influx. One metric we would like to see would be either the number of currently in use threads, or the number of potentially remaining threads that could be used (max size - executing). The latter would mimic the available concurrent calls metric that gets generated from the semaphore bulkhead. Not having this metric makes it more difficult to know when a thread pool is sized incorrectly. Could this be added? Or do I have a way of inferring this somehow?

jjathman:
Fixes #1496.

One of the tests for this seemed a little flakey due to a timing issue, I think it's due to `ExecutorService.getActiveCount()` not being guaranteed to always provide the exact answer. Open to suggestions about how to guarantee it always working.

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkhead.java
@@ -312,6 +312,13 @@ public interface ThreadPoolBulkhead extends AutoCloseable {
          * @return the queue capacity
          */
         int getQueueCapacity();
+
+        /**
+         * Returns the number of actively executing tasks.
+         *
+         * @return the number of executing tasks
+         */
+        int getActiveThreadCount();
     }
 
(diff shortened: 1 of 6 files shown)


jjathman:
One other note, it looks like the `ThreadPoolBulkhead` is missing from the metrics page. Happy to contribute that as well, but I'm not really sure where that is stored. https://resilience4j.readme.io/docs/micrometer

RobWin:
Documentation is using readme.io
There is a button on every site where you can suggest edits.
I can then review and merge them.

jjathman:
@RobWin thanks, I made some suggested updates to the documentation. If you approve the PR I can add the new metric to the documentation once that's been released as well.

RobWin (approved):

PR #1501, opened 2021-07-21
Add new `available.thread.count` metric to `ThreadPoolBulkhead`

### LINKED ISSUE #1500: Add ThreadPoolBulkheadMetric for Available Thread Count
Resilience4j version:
1.7

Java version: any

This is a followup to #1496 to add one more metric which is the available number of threads (maximum - active threads). This will make it easier for creating dashboards for applications which utilize both the the semaphore and thread pool bulkheads. The current metrics measure opposite things (remaining available capacity vs. active work). Having a metric produced from both bulkhead types will make consistent visualizations much easier. 

PR to come.

jjathman:
Fixes #1500 

Allows for tracking the max threads remaining which can execute work. Similar to `available.concurrent.calls` from `Bulkhead`

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkhead.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkhead.java
@@ -319,6 +319,14 @@ public interface ThreadPoolBulkhead extends AutoCloseable {
          * @return the number of executing tasks
          */
         int getActiveThreadCount();
+
+        /**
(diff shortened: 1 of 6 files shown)


RobWin (approved):

RobWin:
Hi, it seems some tests are failing

jjathman:
> Hi, it seems some tests are failing

@RobWin I'm not very familiar with GitHub's CI. Is there a way to see the full test results including the stack trace?

jjathman:
@RobWin Think I've got the tests fixed now. Not sure how I missed that. Apologies.

Issue #1514, opened 2021-08-09
Missing metrics when replacing circuit-breakers in a watched registry when using resilience4j-prometheus


KoxAlen:
Resilience4j version: 1.7.1

Java version: 14

When we call replace in the registry the replaced circuit-breaker does't add samples on the histogram `resilience4j_circuitbreaker_calls` metric.

Probably because the collector is only hooking the add event and not the replace one too
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-prometheus/src/main/java/io/github/resilience4j/prometheus/collectors/CircuitBreakerMetricsCollector.java#L44

PR #1515, opened 2021-08-10
Issue #1514: Fixed CircuitBreakers prometheus metrics when replacing entry in the registry

KoxAlen:
We found a related bug to this when trying your library, so I tried to submit a PR for the fix of #1514.

Basically the Prometheus collector is only watching for added events but not for replace ones.

Tell me if I missed anything about how to contribute.


RobWin (approved):

RobWin:
Thx

Issue #1517, opened 2021-08-12
Fallback method signature out of sync causes primary method call failure


priyaaggarwal24:
Resilience4j version: 1.5.0

Java version: 11

Issue: When fallback method signature is not in sync with primary method signature, resilience4j throws this error at runtime and fails primary method call; it should instead use some default strategy for fallback. 
These two methods can go out of sync with time. With primary method call failure at runtime being the only way to detect them out of sync, there is huge risk associated with maintaining methods annotated with @CircuitBreaker. 
Additionally, testing circuitbreaker fallback requires integration tests, having tests for all usages of circuitbreakers in the app is build time and effort expensive.

Exception stacktrace:
Caused by: java.lang.NoSuchMethodException: class Output class Repository.getByIdFallback(class Id,class java.lang.Throwable)
	at io.github.resilience4j.fallback.FallbackMethod.create(FallbackMethod.java:92)
	at io.github.resilience4j.circuitbreaker.configure.CircuitBreakerAspect.circuitBreakerAroundAdvice(CircuitBreakerAspect.java:115)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethodWithGivenArgs(AbstractAspectJAdvice.java:644)
	at org.springframework.aop.aspectj.AbstractAspectJAdvice.invokeAdviceMethod(AbstractAspectJAdvice.java:633)
	at org.springframework.aop.aspectj.AspectJAroundAdvice.invoke(AspectJAroundAdvice.java:70)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:175)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:749)
	... 76 common frames omitted

PR #1519, opened 2021-08-12
Issue #1518 - Eliminated code duplication around resolving fallback method and using it

priyaaggarwal24:
Eliminated duplicate code around fallback method handling in following classes

1. CircuitBreakerAspect
2. RateLimiterAspect
3. BulkheadAspect
4. TimeLimiterAspect
5. RetryAspect

Extracted duplicate code for following and put in FallbackMethodExecutor class; replaced FallbackMethodDecorators dependency with FallbackMethodExecutor in all *Aspect classes
 - Resolving fallback method
- execution path when fallback method exists
- execution path when fallback method name is empty

PR #1522, opened 2021-08-13
Issue #1517 - Don't fail primary method execution when fallback method is not found

priyaaggarwal24:
(Depends on branch of PR https://github.com/resilience4j/resilience4j/pull/1519 for removing duplicate code)

Fixes issue #1517 

Since primary and fallback method can get out of sync during software maintenance, not finding fallback method should not result in runtime error even before primary method is executed. We should choose a sensible default strategy that doesn't harm the application in case of misconfiguration, especially when it can be detected only at runtime or via integration tests (they are expensive so are written sparingly).

If fallbackMethod is not found or is invalid, 
 - then catches the NoSuchMethodException
 - Emits a warning log
 - proceeds with primary function execution without FallBackDecorator
 
This is similar to the approach we take when fallbackMethodValue in annotation is either unspecified or is empty string.


RobWin (approved):

RobWin:
Could you please have a look at the failing test?

priyaaggarwal24:
Looking into it @RobWin

RobWin:
Sry, I deleted the comment.
I think the error was introduces in a prev. PR

PR #1504, opened 2021-07-28
Adding SynchronousQueue support to FixedThreadPoolBulkhead

KimDoubleB:
From #1503 

I modified the code related to the above issue.
- If queueCapacity is 0, use SynchronousQueue, and if it is greater than 0, use ArrayBlockingQueue.

PR #1536, opened 2021-09-10
Fix test fail (spring boot test)

KimDoubleB:
Hi.
I make this PR because there was a test issue in the previously merged PR(#1504).

Once look at the test code, test error doesn't seem to be an issue that occurred in #1504 PR.

As BACKEND_D is used in several places, it seems that `testBulkheadAutoConfigurationThreadPoolContextPropagation` test fails or is successful according to the test order.
So, I created BACKEND_E and used it independently to prevent problem.

In addition, a test error occurred randomly in the `testBulkheadAutoConfigurationRxJava2` test.
I modified it to operate normally using `CountDownLatch`.


RobWin:
Thank you

RobWin (approved):

Issue #1387, opened 2021-03-25
New feture: open CircuitBreaker for a time based on the decorated calls response


walec51:
Problem:

We have a client class that uses resiliance4j decorators. The underlining API, which the client calls, can ban our IP for some time for multiple reasons. The underlining API in its error response returns the duration of the given ban and expects that we will stop any further communication during it. Not doing so could result in an even longer ban.

Solution:

This is an ideal use case for a circuit breaker as the bans duration can exceed the period of any rate limiter that is also applied. However resilience4j circuit breaker lacks configuration options and some implementation mechanisms to handle this. Therefore I propose to add the following API:

```
CircuitBreakerConfig.from(...)
                // ...
                .forceOpenOnResult(Either<Throwable, T> result -> {
                    if (result.getLeft() instanceof SomeSpecificException) {
                        SomeSpecificException e = (SomeSpecificException) result.getLeft();
                        return ForceOpenChecksResult.neededUntil(e.getBannedUntil()); // Instant
                    }
                    return ForceOpenChecksResult.notNeeded();
                })
                .build()
``` 

`ForceOpenChecksResult` could also have a creator method called `ForceOpenChecksResult.needed()` that would force open the circuit breaker for the time configured in `waitDurationInOpenState` attribute.

I'm willing to implement this feature if it will be accepted. We need this feature in this project: https://github.com/knowm/XChange where we already use resilience4j for rate limiting and retries.

RobWin:
I like the idea that you can handle results and based on the result you can force a transition to open. 
Comparable to `RateLimiterConfig.drainPermissionsOnResult` predicate, we could add a `CircuitBreakerConfig.forceOpenOnResult()` predicate.

walec51:
@RobWin to handle my use case I would need something more then a predicate - I'm thinking more of an `Function<ForceOpenChecksResult, Either<Throwable, T>>` as shown in the example

walec51:
in `ForceOpenChecksResult` one would pass if a force open is needed and optionally until which point in time should it be opened

RobWin:
Ok, 
some ideas:

1) We have to add a `private void transitionToOpenState(long waitDurationInMillis)` to the `CircuitBreakerStateMachine` and add another constructor to `CircuitBreakerStateMachine.OpenState` so that you can use a fixed waitDurationInMillis and not use the `waitIntervalFunctionInOpenState` from the CircuitBreakerConfig.

[...]

PR #1560, opened 2021-10-10
#1387 open CircuitBreaker for a time based on the decorated calls response

walec51:
implementation of #1387


RobWin:
LGTM. Thank you

walec51:
added javadocs and two more unit tests

PR #1564, opened 2021-10-13
Issue #1554: If "default" config defined, CircuitBreakers configurations will inherit from it by default unless overridden setting specific baseConfig property

### LINKED ISSUE #1378: resilience4j-spring-boot2: Create circuit breaker based on default configuration
Resilience4j version: 1.6.1.

I have default resilience4j configuration:
```
resilience4j.circuitbreaker:
    configs:
        default:
            slidingWindowSize: 100
            permittedNumberOfCallsInHalfOpenState: 10
            waitDurationInOpenState: 10000
            failureRateThreshold: 60
            eventConsumerBufferSize: 10
```

and some specific:
```
resilience4j.circuitbreaker:
    instances:
        backendA:
            slidingWindowSize: 50
```
In that case, the circuit breaker was created based on a specific configuration without using the default configuration. 
To override the default configuration I should use `baseConfig` property.

As for me it more reasonable to override the default configuration without `baseConfig` property. And in case, if  `baseConfig` is presented, doesn't override the default configuration and override configuration from `baseConfig` property. 

The same for all modules - TimeLimiter/Bulkhead/ThreadPoolBulkhead ets.

### LINKED ISSUE #1554: [resilience4j-spring-boot2] Set register-health-indicator property to true by default
Resilience4j version: 1.7.0

Java version: openjdk-11.0.12-1

Hello.

We'd like to set `register-health-indicator` property to true by default for every circuit-breaker defined in yaml, but to do that, we need to define a baseConfig and made all circuit-breakers inherit from that config using `baseConfig` property.

Would it be possible to define some kind of special `defaultConfig` that is always inherited by all circuit-breakers without needing to set  `baseConfig` property for each?

We develop a internal company framework on top of spring-boot and sometimes it's very useful for us to provide some default config values for all our users (internal application developers).

If this is not possible... would you please make this method protected so we could easily extend it and modify its behaviour?

https://github.com/resilience4j/resilience4j/blob/7ce1227f9d120fd5175ca9abf9927a26d9fbabfc/resilience4j-spring-boot2/src/main/java/io/github/resilience4j/circuitbreaker/monitoring/health/CircuitBreakersHealthIndicator.java#L100-L104

Thanks!

JoseLora:
Closes #1554, Closes #1378

### DIFF
--- a/resilience4j-spring-boot2/src/test/resources/application.yaml
+++ b/resilience4j-spring-boot2/src/test/resources/application.yaml
@@ -8,7 +8,7 @@ resilience4j.retry:
     tag1: tag1Value
     tag2: tag2Value
   configs:
-    default:
+    sharedConfig:
       maxAttempts: 4
       waitDuration: 100ms
       retryExceptionPredicate: io.github.resilience4j.circuitbreaker.RecordFailurePredicate
@@ -19,9 +19,9 @@ resilience4j.retry:
         - io.github.resilience4j.circuitbreaker.IgnoredException
   backends:
     retryBackendA:
-      baseConfig: default
+      baseConfig: sharedConfig
     retryBackendB:
-      baseConfig: default
+      baseConfig: sharedConfig
     retryBackendC:
       maxAttempts: 3
       waitDuration: 100ms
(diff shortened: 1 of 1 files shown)


RobWin:
Thank you. LGTM.
Could you also change Retry/RateLimiter/...?

JoseLora:
Hi @RobWin. Done replicating similar changes to RateLimiter, Retry, TimeLimiter, Bulkhead and ThreadPoolBulkhead.

PR #1596, opened 2021-12-11
bump vertx to 4.x

brunolellis:
Update Vert.x dependency version to 4.2.1.

`Future.future()` is deprecated since 3.8.x and removed after 4.x version.


RobWin:
We will remove Vertx support from 2.x

Issue #1588, opened 2021-11-22
[Enhancement request] Support for Prometheus tags 


mdsalmeida92:
Currently, the only option to have custom tags is to use resilience4j-micrometer package.

Are there any plans to implement this in the resilience4j-prometheus?

RobWin:
Hi, no there are no plans right now.
But feel free to create a PR. We need any support.

PR #1601, opened 2021-12-20
[ISSUE-1588] Support prometheus tags

mdsalmeida92:
This is a pull request for the issue #1588 to add custom tags as metric labels to the Prometheus module


RobWin:
Hi, could you please also add a test

RobWin (approved):

PR #1627, opened 2022-02-09
Restore Java 8 compatibility to DropWizards Metrics Module

mikebell90:
* Replaced usages of `List.of` with `Collections.singletonList`

Other usages of List.of, Set.of, and Map.of are scattered throughout Resilience4j,
however the others are in test classes, so not necessarily a big deal.


mikebell90:
HA! I checked master instead of the 1.7.1 tag. So it looks like Java 8 compatiblity is being removed in Resilience4j 2? So this is probably deliberate, in which case, feel free to close.

RobWin:
Hi, yes you are right.
In resilience4j we migrated a lot of code to Java 11 and removed Vavr.

Issue #1624, opened 2022-02-04
CircuitBreaker.executeSuspendFunction unable to register CancellationException


ollis:
Resilience4j version: 1.7.1

Java version: 11.0.2-open

I want to wrap some [Ktor Client](https://ktor.io/docs/client.html) calls with a CircuitBreaker. The [timeout feature](https://ktor.io/docs/timeout.html) throws `HttpRequestTimeoutException` in case of timeouts. This class extends `CancellationException` and `CircuitBreaker.executeSuspendFunction` will therefore ignore it. Could you add a configuration option that made it possible to record specified instances of CancellationException as failures?

RobWin:
Hi, would you like to create a PR?

ollis:
Hi, I can give it a try, hopefully during this week.

PR #1633, opened 2022-02-22
Issue #1624

ollis:
Removed special treatment of CancellationException in CircuitBreaker

### DIFF
--- a/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
+++ b/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
@@ -20,14 +20,26 @@ package io.github.resilience4j.kotlin.circuitbreaker
 
 import io.github.resilience4j.circuitbreaker.CircuitBreaker
 import io.github.resilience4j.kotlin.isCancellation
-import java.util.*
 import java.util.concurrent.TimeUnit
+import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.coroutineContext
 
 /**
  * Decorates and executes the given suspend function [block].
+ * Cancellation errors are ignored.
  */
-suspend fun <T> CircuitBreaker.executeSuspendFunction(block: suspend () -> T): T {
+suspend fun <T> CircuitBreaker.executeSuspendFunction(
+    block: suspend () -> T
+): T = executeSuspendFunction({ t, c -> isCancellation(c, t) }, block)
+
+/**
+ * Decorates and executes the given suspend function [block].
+ * Release permission without recording error if [ignoreThrowablePredicate] is true.
+ */
+suspend fun <T> CircuitBreaker.executeSuspendFunction(
+    ignoreThrowablePredicate: (Throwable, CoroutineContext) -> Boolean,
+    block: suspend () -> T
+): T {
     acquirePermission()
     val start = System.nanoTime()
     try {
@@ -36,7 +48,7 @@ suspend fun <T> CircuitBreaker.executeSuspendFunction(block: suspend () -> T): T
         onResult(durationInNanos, TimeUnit.NANOSECONDS, result)
         return result
     } catch (exception: Throwable) {
-        if (isCancellation(coroutineContext, exception)) {
+        if (ignoreThrowablePredicate(exception, coroutineContext)) {
             releasePermission()
         } else {
             val durationInNanos = System.nanoTime() - start
@@ -45,6 +57,14 @@ suspend fun <T> CircuitBreaker.executeSuspendFunction(block: suspend () -> T): T
         throw exception
     }
 }
+
+/**
+ * Decorates and executes the given suspend function [block].
+ * All types of throwable including cancellation exceptions are recorded with [CircuitBreaker.onError]
+ */
+suspend fun <T> CircuitBreaker.executeSuspendFunctionAndRecordCancellationError(block: suspend () -> T): T =
+    executeSuspendFunction({ _, _ -> false }, block)
+
 
 /**
  * Decorates the given *suspend* function [block] and returns it.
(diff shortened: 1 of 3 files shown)


resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
```diff
@@ -36,12 +33,8 @@ suspend fun <T> CircuitBreaker.executeSuspendFunction(block: suspend () -> T): T
         onResult(durationInNanos, TimeUnit.NANOSECONDS, result)
         return result
     } catch (exception: Throwable) {
-        if (isCancellation(coroutineContext, exception)) {
```
RobWin, line 39: I think we cannot just remove this code snippet.
There were could reasons to release a permissions when a coroutine was cancelled.

I assume sometimes you want to handle a `CancellationException` and release a permission and 
sometimes you want to handle a `HttpRequestTimeoutException` and track it as an error.

RobWin (changes requested):

resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt
ollis: I think this undocumented feature breaks the semantics/API of the circuit breaker. I was caught by
surprise when exceptions were silently swallowed. Shouldn't you use the configuration option `ignoreExceptions` if you wanted this behaviour?

RobWin: No, this release permission functionality is needed when a call is done in HALF_OPEN state.
Every call in HALF_OPEN state decreases the number of permitted calls (`permittedNumberOfCalls
InHalfOpenState`). But if someone cancels the call, we want to increase the number of permitted calls again.

It's implemented similar in Futures, RxJava and Reactor:

Future:
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreaker.java#L1088-L1091

Reactor:
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/circuitbreaker/operator/CircuitBreakerSubscriber.java#L73-L81

But in Java a `TimeoutException` does not inherit from a `CancellationException`.

RobWin: But I agree the exception should not be swalloed but rethrown,

RobWin: Not sure about it. It feels wrong that you need to add `CancellationException` to the list of ignored exceptions so that the number of permitted calls is decreased when the CircuitBreaker is in HALF_OPEN state. The idea was that request cancellation works without additional configuration.
It's not neccessary for Futures, RxJava and Reactor. I don't want to introduce this just for Kotlin Ktor.  It would also be a non-backwards compatible change for resilience4j-kotlin. 

It also feels strange that `HttpRequestTimeoutException` inherits from `CancellationException`. What's the reasoning behind it? Does the exception cancel the coroutine which is executing the call using the Ktor client library?

ollis: But you cant make it work without additional configuration since it is totally dependent on the context if a cancellation is an error, success or should be ignored? I dont think you can always assume that this should be treated as ignore as it is implemented now.  And `HttpRequestTimeoutException` from Ktor Client is a good example of that. A http request is cancelled if it takes too long time, and in my context I want to count that as an error so the CircuitBreaker trips if too many requests times out. The exception is given as a detailed cancellation cause, see https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/cancel.html

RobWin: For backwards-compatibility we could add a new extension function instead of modifying the existing one? Wdyt?

ollis: Yes, that is OK for me, but then we are left with one of the hard problems in computer science: naming 😄

RobWin: `executeSuspendFunctionAndIgnoreCancellation` ?

ollis: I think that is a bit confusing since it actually describes the current implementation of `executeSuspendFunction`? Maybe `executeSuspendFunctionAndRecordCancellationError`could work?

ollis: Or `executeSuspendFunctionAndRecordAllThrowables`?

RobWin: I prefer `executeSuspendFunctionAndRecordCancellationError` or `executeSuspendFunctionAndRecordCancellationException`

ollis: I made a new commit with `executeSuspendFunctionAndRecordCancellationError`.

RobWin (approved):

RobWin:
Could you rebases the latest changes?
Should fix the unit test.

ollis:
OK, I hope it works now.

Issue #1253, opened 2020-12-01
How to modify/close the retried object if there's any 


Chr3is:
We use resilience4j to retry requests to our backend. For example we want to retry the request if the response object contains a status 500 (jaxrs). However I haven't seen a solution to call .close() on the response object  before a new try is done. I'm missing an optional postAction-BiConsumer with the current attemptCount & the received object T (the response in this case) which can be registered in the retryconfig.  

eg:

```
(attemptCount, response) -> {
   if(attemptCount < (maxAttempts - 1)){
        response.close();
   }
}
```

Resilience4j version:
1.6.1 (resilience4j-retry)

Java version:
8

RobWin:
Hi,
can't you automatically unmarshalls the response?

```
Customer customer = client.target("...")
                          .get(Customer.class);

```

In that case the client framework converts any HTTP error code into an exception. JAX-RS 2.0 has added an exception hierarchy for various HTTP error conditions. 

You can then then retry based on the exception which is thrown. In your case it should be an `InternalServerErrorException`.

Chr3is:
This is not an option for us because we invoke for a response (retry until we receive a response object under the given conditions or after the last try) and then evaluate for example the headers. The post validation is not and should not be part of the retry mechanism. To free the connections we have to call .close() or bufferEntity(). A quick hack would be following:

```
RetryConfig.Builder<Response>()
                    .maxAttempts(maxAttempts)
                    .retryOnResult(response -> { 
                                   if(response.getStatusInfo().getFamily() == SERVER_ERROR){
                                       response.bufferEntity();
                                       return true;
                                   }
                                  return false;
                     })
                    .retryExceptions(ProcessingException.class)
                    .build();
```
but this is not clean, because no one would expect a bufferEntity() call in the retryOnResult predicate.

The whole thing is part of our clientlibrary which is used by our clients to communicate to the backends and they want/need the response (not closed) in an error case.

If we have 3 attempts following should be possible during the retry cycle:
1. attempt (status 500) -> response.close()
2. attempt (status 500) -> response.close()
3. attempt (status 500) -> response.bufferEntity()

we would then receive the response object, do some post validation, wrap it with an exception and return the exception to the caller who would then be able to do his things on the buffered response.

Chr3is:
@RobWin do you need more information?

RobWin:
Can you create a proposal?

Chr3is:
[...]

PR #1634, opened 2022-02-24
Issue #1253

Chr3is:
Create possibility to perform post actions on the result if it needs to be retried. For example if you want to invoke .close() on a response object before the next attempt is performed.

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.LongAdder;
+import java.util.function.BiConsumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
@@ -49,6 +50,10 @@ public class RetryImpl<T> implements Retry {
     private final RetryEventProcessor eventProcessor;
     @Nullable
     private final Predicate<T> resultPredicate;
+
+    @Nullable
+    private final BiConsumer<Integer, T> consumeResultBeforeRetryAttempt;
+
     private final String name;
     private final RetryConfig config;
     private final Map<String, String> tags;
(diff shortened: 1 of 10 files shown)


Chr3is:
@RobWin is something missing before the review can be done?

RobWin:
Yes, time :(
I'm currently very busy

Chr3is:
@RobWin Alright take your time! And thanks for replying.

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -177,6 +183,9 @@ public void onComplete() {
         public boolean onResult(T result) {
             if (null != resultPredicate && resultPredicate.test(result)) {
                 int currentNumOfAttempts = numOfAttempts.incrementAndGet();
+                if(postActionOnRetryBiConsumer != null){
+                    postActionOnRetryBiConsumer.accept(currentNumOfAttempts, result);
```
RobWin, line 187: What should happen if the consumer itself throws an exception?
I think we should catch the exception and log it, but not rethrow it.

RobWin: Shouldn't we execute the consumer only when a retry attempt needs to be performed? I think the consumer must not be invoked when no retry attempt must be performed, or? In that case the normal code flow gets the result and can do whatever is necessary.
We could move this line of code in front of  `waitIntervalAfterFailure(currentNumOfAttempts, Either.right(result));`, or?
We could also rename `postActionOnRetryBiConsumer` to `consumeResultBeforeRetryAttempt`. Wdyt?

RobWin (changes requested):

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
Chr3is: I though possible exception in the postAction should be handled by the creator of the function. If we catch the exception and just log it, it might be unrecognized. The behavior is at the moment equal to the resultPredicate test isn't it?

I agree with you. We should not perform the postAction if no retry is needed and the renaming to "consumeResultBeforeRetryAttempt" would clarify this behavior. However the postAction is currently only performed if the resultPredicated returned true e.g. we already have the desired behavior don't we?

Chr3is:
@RobWin what do you think about my review questions?

RobWin:
Which one?
I think I answered all questions, or?

Chr3is:
@RobWin I had some questions left after your review: https://github.com/resilience4j/resilience4j/pull/1634#discussion_r860647957

RobWin:
Ooh, I forgot to press the submit button of my review.
I answered many days ago ;)

resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
Chr3is: You're right if we rename the consumer to consumeResultBeforeRetryAttempt we should not invoke the consumer after the last retry. The user can do whatever he wants with the returned object in the normal code-flow. I will move the consumer accordingly.

Chr3is:
@RobWin should be fine now

RobWin:
Thank you

PR #1671, opened 2022-04-23
Update AtomicRateLimiter.java

chin-chin:
let's make our life more easy :)


chin-chin:
Hi, I put this object into remote cache (Redis) and have some problems with serialization this object. Suppose I'm not alone

storozhukBM:
I don't think this is the right way to do it. AtomicRateLimiter is not designed to be serialized, and I suspect that permissions reservation logic won't work correctly in this setup.

chin-chin:
Yep, I see  - serialization fulfilled without any problem after that, but deserialization fulfilled with a wrong way, some important properties init with a default values :). I supposed to go on serialize with the standart way but restore values with my own logic. The object appeared to be to complex for automating  restoration. Tnx for cooperation, you save me a lot of time.

Issue #1683, opened 2022-05-09
SemaphoreBasedRateLimiter  could cause memory leak


aCoder2013:
Resilience4j version:
1.7.1
Java version:
1.8

SemaphoreBasedRateLimiter use scheduledExecutorService to refresh permit, but it didn't provide a way to shut down or close it, even if there are no references  to SemaphoreBasedRateLimiter  instance,  the schedule tasks are still running,  so SemaphoreBasedRateLimiter can't be garbage collected.

> In our use case , we'll create millions of SemaphoreBasedRateLimiter instance, so after some long running,  jvm will be filled with SemaphoreBasedRateLimiter instances, and cause too much full gc, jvm is basically dead

```java

    private void scheduleLimitRefresh() {
        scheduler.scheduleAtFixedRate(
            this::refreshLimit,
            this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
            this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
            TimeUnit.NANOSECONDS
        );
    }
```

storozhukBM:
`SemaphoreBasedRateLimiter` gives you an API to provide your own `ScheduledExecutorService` so you can share it across different limiters and close it when necessary.
https://github.com/resilience4j/resilience4j/blob/0e4964c44bd7366ab2da3391ac72deeb8e3ae956/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java#L86

storozhukBM:
A default limiter constructor is created for permanent limiters that will live throughout all application lifetime.

aCoder2013:
Thanks for your reply ! we actually share the same ScheduledExecutorService,  so  we may need the ScheduledFuture returned from `scheduleAtFixedRate` and close the specific RateLimitter

storozhukBM:
Yes, you can make a PR with close() method implementation for `SemaphoreBasedRateLimiter` if you want.
In general I'd suggest you to use default `AtomicRateLimiter` in such complex cases.

aCoder2013:
Thanks, i'll try to make a PR, we've tested  `SemaphoreBasedRateLimiter ` and `AtomicRateLimiter `, AtomicRateLimiter will create so many `State` instances due to its CAS mechanism in our use case, so we prefer `SemaphoreBasedRateLimiter `.

storozhukBM:
[...]

PR #1684, opened 2022-05-10
Issue #1683: SemaphoreBasedRateLimiter could cause memory leak 

aCoder2013:
Resilience4j version:
1.7.1
Java version:
1.8

SemaphoreBasedRateLimiter use scheduledExecutorService to refresh permit, but it didn't provide a way to shut down or close it, even if there are no references  to SemaphoreBasedRateLimiter  instance,  the schedule tasks are still running,  so SemaphoreBasedRateLimiter can't be garbage collected.

> In our use case , we'll create millions of SemaphoreBasedRateLimiter instance, so after some long running,  jvm will be filled with SemaphoreBasedRateLimiter instances, and cause too much full gc, jvm is basically dead

```java

    private void scheduleLimitRefresh() {
        scheduler.scheduleAtFixedRate(
            this::refreshLimit,
            this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
            this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
            TimeUnit.NANOSECONDS
        );
    }
```

### DIFF
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
@@ -29,6 +29,7 @@ import io.github.resilience4j.ratelimiter.event.RateLimiterOnFailureEvent;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnSuccessEvent;
 import io.github.resilience4j.ratelimiter.internal.AtomicRateLimiter;
 
+import java.io.Closeable;
 import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.Callable;
@@ -49,7 +50,7 @@ import static java.util.Collections.emptyMap;
  * necessary until a permit is available, and then takes it. Once acquired, permits need not be
  * released.
  */
-public interface RateLimiter {
+public interface RateLimiter extends Closeable {
 
     /**
      * Creates a RateLimiter with a custom RateLimiter configuration.
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
@@ -25,13 +25,11 @@ import io.github.resilience4j.ratelimiter.event.RateLimiterOnDrainedEvent;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnFailureEvent;
 import io.github.resilience4j.ratelimiter.event.RateLimiterOnSuccessEvent;
 
+import java.io.IOException;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static java.util.Collections.emptyMap;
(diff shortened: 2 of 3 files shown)


resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiter.java
```diff
@@ -49,7 +50,7 @@
  * necessary until a permit is available, and then takes it. Once acquired, permits need not be
  * released.
  */
-public interface RateLimiter {
+public interface RateLimiter extends Closeable {
```
storozhukBM, line 53: Since not all types of rate limiters will require closing, I'd prefer that, for now, only `SemaphoreBasedRateLimiter` would expose the `close` method, and it doesn't need to be a part of the Closable interface.

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -25,13 +25,11 @@
...
 
+import java.io.IOException;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
storozhukBM, line 32: please do not use star imports

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -121,15 +120,19 @@ private ScheduledExecutorService configureScheduler() {
...
+    private ScheduledFuture<?> scheduleLimitRefresh() {
+        return scheduler.scheduleAtFixedRate(
             this::refreshLimit,
             this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
             this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
             TimeUnit.NANOSECONDS
         );
     }
 
+    public ScheduledFuture<?> getScheduledFuture() {
```
storozhukBM, line 132: what is the reason to expose this field?

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -260,6 +263,13 @@ private void publishRateLimiterAcquisitionEvent(boolean permissionAcquired, int
         eventProcessor.consumeEvent(new RateLimiterOnFailureEvent(name, permits));
     }
 
+    @Override
+    public void close() throws IOException {
```
storozhukBM, line 267: Please, add a few words into documentation of this class on how and when the user needs to call this method.

storozhukBM, line 267: This code needs to be covered with some test cases.

storozhukBM (changes requested):

storozhukBM:
Thanks for your PR @aCoder2013 
I have a few comments, please take a look.

aCoder2013:
@storozhukBM thanks for your time  , i'll take a look today and improve the code based on your suggestions

aCoder2013:
I've fixed code style problems you mentioned, could you plz take a look again?  @storozhukBM

resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
```diff
@@ -260,6 +262,18 @@ private void publishRateLimiterAcquisitionEvent(boolean permissionAcquired, int
         eventProcessor.consumeEvent(new RateLimiterOnFailureEvent(name, permits));
     }
 
+    /**
```
storozhukBM, line 265: This one looks very good, I'm ready to merge it, but can you please add a mention about this method to this class doc at the top of a file.

PR #1693, opened 2022-05-19
Issue #1683: SemaphoreBasedRateLimiter could cause memory leak

aCoder2013:
Resilience4j version:
1.7.1
Java version:
1.8

SemaphoreBasedRateLimiter use scheduledExecutorService to refresh permit, but it didn't provide a way to shut down or close it, even if there are no references  to SemaphoreBasedRateLimiter  instance,  the schedule tasks are still running,  so SemaphoreBasedRateLimiter can't be garbage collected.

> In our use case , we'll create millions of SemaphoreBasedRateLimiter instance, so after some long running,  jvm will be filled with SemaphoreBasedRateLimiter instances, and cause too much full gc, jvm is basically dead

```java

    private void scheduleLimitRefresh() {
        scheduler.scheduleAtFixedRate(
            this::refreshLimit,
            this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
            this.rateLimiterConfig.get().getLimitRefreshPeriod().toNanos(),
            TimeUnit.NANOSECONDS
        );
    }
```


RobWin:
LGTM

storozhukBM (approved):
Approve

PR #1698, opened 2022-05-30
Added support for micrometer 1.10 observations

### LINKED ISSUE #1628: Provide observability via Micrometer 2.0.0 & Micrometer Tracing
# Introduction

With Micrometer 2.0.0 we're introducing a concept of an `Observation` ([you can read more about it here](https://micrometer.io/docs/tracing)). An `Observation` wraps an action that we would like to measure. A measurement can be wrapping it in a `Timer` (metrics), creation of a `Span` (tracing) or any other custom behavior such as e.g. logging. 

Let's check the following example

```java
ObservationRegistry registry = new SimpleMeterRegistry();
Observation observation = Observation.start("my.operation", registry);
try {
    observation.scoped(() -> doSomeWorkHere());
} finally {
    observation.stop();
}
```

Depending on the configuration of the `ObservationRegistry` there might be various outcomes when `doSomeWorkHere()` gets executed.

# Benefit

Using the Observation API makes you instrument your code once but you get multiple benefits out of it (e.g. metrics & tracing). Since Micrometer Tracing works with Micrometer we also bridge the span creation to such tracer implementations as OpenZipkin Brave and OpenTelemetry.

# Feature request

We would like to suggest that such instrumentation using Micrometer would happen in this project. Doing so would allow us to transparently continue any traces that occurred on the client side and have consistency in terms of used metrics. You could also emit metrics and traces (and whatever you want really) on the server side.

marcingrzejszczak:
fixes #1628

### DIFF
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -22,7 +22,7 @@ ext {
     spockVersion = '1.3-groovy-2.5'
     feignVersion = '10.2.0'
     reactiveStreamsVersion = '1.0.2'
-    micrometerVersion = '1.10.0-SNAPSHOT'
+    micrometerVersion = '1.10.0'
     hibernateValidatorVersion = '6.0.18.Final'
     wiremockVersion = '2.26.0'
     validationApiVersion = '2.0.1.Final'
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -22,7 +22,7 @@ ext {
     spockVersion = '1.3-groovy-2.5'
     feignVersion = '10.2.0'
     reactiveStreamsVersion = '1.0.2'
-    micrometerVersion = '1.5.1'
+    micrometerVersion = '1.10.0-SNAPSHOT'
     hibernateValidatorVersion = '6.0.18.Final'
     wiremockVersion = '2.26.0'
     validationApiVersion = '2.0.1.Final'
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
That's a good question. With the Observations unless the users provides some custom handler mechanism, if they go with the default Micrometer `TimerObservationHandler` and the tracing one, I would expect they would get a `Timer` around the function execution and a corresponding `Span`.

marcingrzejszczak:
GA is planned for 14.11 at the moment (https://github.com/micrometer-metrics/micrometer/milestone/177). I'll double check that the PR is ready to be merged with the latest changes. Other than that it's not GA, I guess you're ok with the proposed changes?

marcingrzejszczak:
It would be great if Feign would release soon after so that Spring Cloud CircuitBreaker could use it before the 30.11 - it's when we're doing 2022.0.0 GA release (https://github.com/spring-cloud/spring-cloud-release/milestone/115). Do you think we have a chance of going with that schedule?

RobWin:
Yes, I try my best :D

marcingrzejszczak:
cc @ryanjbaxter

marcingrzejszczak:
I've upgraded Micrometer to 1.10.0!

RobWin (approved):

RobWin:
2.0 released

Issue #1810, opened 2022-11-18
question regarding limitation on retry backoff multiplier


krinklesaurus:
Resilience4j version: latest

Java version:
 
    java 11.0.1 2018-10-16 LTS
    Java(TM) SE Runtime Environment 18.9 (build 11.0.1+13-LTS)
    Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.1+13-LTS, mixed mode

Hi,

I was tryin to set the use a custom RetryConfig and also customize the IntervalFunction. I would like to have a maximum wait time somehow, in a way it uses exponential backoff to a certain max and then stick with this max time instead of increasing further. Like 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 10 , 10, 10 instead of ...11, 12, 13.

I think this could be done setting a negative backoff multiplier but this is currently not allowed: https://github.com/resilience4j/resilience4j/blob/master/resilience4j-core/src/main/java/io/github/resilience4j/core/IntervalFunction.java#L255-L259

Is there any specific reason doing this? Mathematically this is possible and I don't see why I'm not alloed to do this :-(

example of how the backoff would look like can be seen here https://exponentialbackoffcalculator.com/ with values Interval 1, Max Retries 50, and Exponential 0.6

thanks in advance

RobWin:
Hi,
no, i don't think that there is specific reasons for it :/
We could fix it in Resilience4j 2.x.
Would you like to create a PR?

PR #1816, opened 2022-11-19
allow any backoff multiplier

krinklesaurus:
Issue #1810 

no reason to limit backoff mulitplier to numbers > 1.0 Negative and numbers < 1.0 should be allowed too.


RobWin:
Thank you

Issue #1856, opened 2022-12-23
Consider setting Thread.currentThread().interrupted() when InterruptedException thrown in Retry execute*()


youngm:
Resilience4j version: 2.0.2

Java version: 17

When using Retry.executeSupplier(), if the thread is interrupted while waiting to retry then the state of the current thread is not interrupted so the invoker has no way of knowing that an interrupt exception was thrown. Resilience4j should set `Thread.currentThread().interrupt()` when an `InterruptedException` is thrown instead of swallowing it so downstream code will know the current thread interrupt state.

I checked Reslience4j 1.7.0 and in that case an InterruptedException is thrown (probably because of a sneakythrows) which isn't correct behavior either. I like how 2.0.2 throws the last failed exception. It should just flag the thread as interrupted also.

Here is a test that I think should pass when working correctly.
```
  @Test
  void test() throws Exception {
    var started = new CountDownLatch(1);
    var done = new CountDownLatch(1);
    var error = new AtomicReference<Throwable>();
    var thread =
        new Thread(
            () -> {
              try {
                Retry.of(
                        "retry",
                        RetryConfig.custom()
                            .maxAttempts(3)
                            .waitDuration(Duration.ofSeconds(10))
                            .build())
                    .executeSupplier(
                        () -> {
                          started.countDown();
                          throw new IllegalStateException("Error");
                        });
              } catch (Throwable e) {
                error.set(e);
                done.countDown();
              }
            });
    thread.start();
    assertThat(started.await(1, TimeUnit.SECONDS)).isTrue();
    assertThat(done.await(2, TimeUnit.SECONDS)).isFalse();
    thread.interrupt();
    assertThat(done.await(1, TimeUnit.SECONDS)).isTrue();
    assertThat(error)
        .hasValueSatisfying(e -> assertThat(e).isInstanceOf(IllegalStateException.class));
    assertThat(thread.isInterrupted()).isTrue();
  }
```

PR #1859, opened 2022-12-27
Set thread interrupted when interrupted from sleep

youngm:
Sets the current thread as interrupted when interrupted from sleep. Fixes https://github.com/resilience4j/resilience4j/issues/1856


youngm:
It could also be argued that if the exception is an InterruptedException we not retry by default. But, this can be handled by the user with an exceptionPredicate so probably less necessary of a change.

Issue #1875, opened 2023-01-23
Add copyright information to Jar.


tobi5775:
Resilience4j version: 2.0.2

Java version: 17

For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar File. Therefore it would be great to have the copyright information included in the Jar File as COPYRIGHT.txt File.

PR #1876, opened 2023-01-23
Issue #1875: add copyright information to jar

tobi5775:
For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar File. Therefore it would be great to have the copyright information included in the Jar File as COPYRIGHT File.


RobWin:
Do you also need the `LICENSE.txt` file?

tobi5775:
We don't necessarily need the license file, but I have added it.

PR #1965, opened 2023-05-29
FallbackMethod supports AOP

SOOHYUN-LIM:
FallbackMethod supports AOP (CircuitBreaker, Cacheable, etc...)

**Motivation**
Fallback method must be supported AOP

**Solution**
 Fallback method changed to use proxy instead of using original object

~~~java
    @CircuitBreaker(name = "test", fallbackMethod = "fallback")
    public String test(String arg) {
        if (arg == null) {
            throw new NullPointerException();
        }
        return "SUCCESS";
    }

    @Cacheable(cacheNames = "fallback")
    public String fallback(NullPointerException e) {
        return "SUCCESS";
    }

    @CircuitBreaker(name = "fallback")
    public String fallback(Exception e) {
        throw new RuntimeException("test..!!");
    }
~~~

### DIFF
--- a/resilience4j-spring/src/main/java/io/github/resilience4j/fallback/FallbackExecutor.java
+++ b/resilience4j-spring/src/main/java/io/github/resilience4j/fallback/FallbackExecutor.java
@@ -28,7 +28,7 @@ public class FallbackExecutor {
         if (StringUtils.hasLength(fallbackMethodName)) {
             try {
                 fallbackMethod = FallbackMethod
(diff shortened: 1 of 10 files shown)


SOOHYUN-LIM:
@RobWin Can you let me know if there is any issue?

RobWin:
Hi,

good idea.
Could you please add some Spring Boot integration test which also shows that the functionality is working as expected?

SOOHYUN-LIM:
@RobWin I have added Spring & Spring6 integration testing.

RobWin (approved):

Issue #1980, opened 2023-06-22
API regression following removal of io.vavr dependencies in v2.0.0


msohaill:
Resilience4j version: 2.0.2

Java version: 17

API regression following [removal](https://github.com/resilience4j/resilience4j/pull/1173/files#diff-232ea4f1ab359fd4e84b2f604be4b76fd864756f281e9d7160e0c219787b45e0) of io.vavr dependencies. The CheckedConsumer no longer has the `unchecked` method which [io.vavr used to support](https://www.javadoc.io/doc/io.vavr/vavr/latest/index.html).

It is a small change I believe: Something similar to [this](https://github.com/resilience4j/resilience4j/blob/63663861c9b929e3742c6ea902697135f8d4de65/resilience4j-core/src/main/java/io/github/resilience4j/core/functions/CheckedSupplier.java#L36) can be added [here](https://github.com/resilience4j/resilience4j/blob/63663861c9b929e3742c6ea902697135f8d4de65/resilience4j-core/src/main/java/io/github/resilience4j/core/functions/CheckedConsumer.java#L27).

Let me know if you need more information!

RobWin:
Hi,

any PR is really welcome.
Good finding.

msohaill:
Wonderful, I've created the above PR #1981 with the changes

PR #1981, opened 2023-06-23
Adding `unchecked` and `andThen` to CheckedConsumer

msohaill:
See #1980. These changes introduce both the `unchecked` and `andThen` methods to the CheckedConsumer interface to mitigate API regression


RobWin:
Thank you

Issue #1761, opened 2022-09-15
onRetry event not published for executeCompletionStage and retryOnResult


huhle:
Resilience4j version: 1.7.1
Java version: openjdk 11.0.15

This issue is very similar (or maybe the same) as https://github.com/resilience4j/resilience4j/issues/978

If retryOnResult is configured and retry is used to decorate CompletableFuture (executeCompletionStage) , there are no onRetry events published in case retry is caused by retryOnResult predicate.

Simple test
```
  private static RetryConfig retryConfig =
      RetryConfig.<Integer>custom()
          .maxAttempts(5)
          .intervalFunction(IntervalFunction.of(100))
          .retryOnResult(
              response -> response < 3)
          .retryOnException(err -> false)
          .build();


  @Test
  public void testSync() throws Exception {
    System.out.println("Running sync");

    List<RetryOnRetryEvent> retryEventList = new ArrayList<>();

    Retry retry = Retry.of("id", retryConfig);
    retry.getEventPublisher()
        .onRetry(retryEventList::add);


    Service s = new Service();
    int result = retry.executeCallable(s::get);

    Assert.assertEquals(3, result);
    Assert.assertEquals(2, retryEventList.size());
  }

  @Test
  public void testAsync() throws Exception {
    System.out.println("Running sync");

    List<RetryOnRetryEvent> retryEventList = new ArrayList<>();

    Retry retry = Retry.of("id", retryConfig);
    retry.getEventPublisher()
        .onRetry(retryEventList::add);


    Service s = new Service();
    ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(1);
    int result = retry.executeCompletionStage(executor, s::getAsync).toCompletableFuture().get();
    executor.shutdown();

    Assert.assertEquals(3, result);
    Assert.assertEquals(2, retryEventList.size());
  }

  private static class Service {
    int counter = 0;

    public int get() {
      counter++;
      System.out.println("Being called (sync). Returning " + counter);
      return counter;
    }

    public CompletableFuture<Integer> getAsync() {
      counter++;
      System.out.println("Being called (async). Returning " + counter);
      return CompletableFuture.completedFuture(counter);
    }

  }
  ```
I believe that the retry events should be called the same for both scenarios, or am I wrong?
Thanks

RobWin:
Yes, you are correct.
Seems to be a bug

RobWin:
Would you like to contribute a bugfix?

[...]

PR #1986, opened 2023-06-30
Issue #1761: Async retry doesn't emit event bugfix

b923:
Hello.

I probably found the solution for Issue #1761.

I also added test which verifies that async retries emits events by now.

Is it possible to add this bugfix into version 1.7.X releases? We are working with java 11 in our internal project.


RobWin (approved):

RobWin:
Unfortunately I won't release any 1.7.x anymore.

Issue #1625, opened 2022-02-04
New module that collects overall information


mkopylec:
Resilience4j version: newest

Java version: doesn't matter

Hi.
What do you think about adding a new module, let's say `resilience4j-status`, which will collect overall metrics (maybe more things like logs or something) for each operation it decorates?
By overall metrics I mean:
counters:
```
resilience4j.status.operation-1.successes
resilience4j.status.operation-1.exceptions
resilience4j.status.operation-1.exceptions.java-lang-exception (exception class name)
```
timers:
```
resilience4j.status.operation-1.duration
```

madhava-sridhar:
there's micrometer module that already exposes these metrics 
https://resilience4j.readme.io/docs/micrometer, is this requirement quite different from what "resilience4j-micrometer" already does ?

RobWin:
Hi,
I think the difference is that the metrics are per circuitbreaker instance or retry instance.
He would like to have a metric per decorated "method", correct?

mkopylec:
Yes, exactly @RobWin

mkopylec:
Maybe the new decorator could be implemented inside resilience4j-micrometer module.

madhava-sridhar:
thanks @RobWin and @mkopylec , this weekend i will give  a try with building decorator for resilience4j-micrometer module.

mkopylec:
@madhava-sridhar Any update on this? :)

mkopylec:
@RobWin I've added a PR for that. Please check it out.

RobWin:
Hi,

I think I would not mix metrics and logging in a single decorator. For logging we can make use of event handlers.

I've implemented an Aspect in one of my projects  which combined a several metrics (Meter and Timer) into a single annotation which could be attached to any public method or class. It measured the execution rates, success/error counts and execution time of methods.
The Aspect added the class name and method name dynamically as tags to the metrics so that I was able to filter in Grafana or Prometheus.
That way I was able to see which methods or classes are the slowest or the methods which are called the most often.

I think that could be a nice addition. Unfortunately I don't have access to the code anymore. 

Would someone be interested to implement this?

mkopylec:
[...]

PR #1989, opened 2023-07-04
Micrometer Timer decorator

mkopylec:
#1625

### DIFF
--- a/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/Timer.java
+++ b/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/Timer.java
@@ -1,0 +1,426 @@
+/*
+ *
+ *  Copyright 2023 Mariusz Kopylec
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *
+ */
+package io.github.resilience4j.micrometer;
+
+import io.github.resilience4j.core.functions.CheckedConsumer;
+import io.github.resilience4j.core.functions.CheckedFunction;
+import io.github.resilience4j.core.functions.CheckedRunnable;
+import io.github.resilience4j.core.functions.CheckedSupplier;
+import io.github.resilience4j.micrometer.internal.TimerImpl;
+import io.micrometer.core.instrument.MeterRegistry;
+
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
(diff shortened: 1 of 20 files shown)


resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/event/TimerOnFailureEvent.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * A TimerEvent which informs that the decorated operation threw an exception.
+ */
+public class TimerOnFailureEvent extends AbstractTimerEvent {
+
+    private final Duration operationDuration;
+
+    public TimerOnFailureEvent(String monitorName, Duration operationDuration) {
+        super(monitorName, SUCCESS);
```
RobWin, line 30: Should be FAILURE or?

mkopylec: Updated

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/Timer.java
```diff
@@ -0,0 +1,428 @@
...
+     * @param supplier the supplier
+     * @param <T>      the supplier result
+     * @return decorated supplier
+     */
+    static <T> CheckedSupplier<T> decorateCheckedSupplier(Timer timer, CheckedSupplier<T> supplier) {
+        return () -> {
+            Context context = timer.createContext();
+            try {
+                T output = supplier.get();
+                context.onSuccess(output);
```
RobWin, line 127: Let's rename `onSuccess` to `onResult` to make it analog to CircuitBreaker or Retry

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/Timer.java
```diff
@@ -0,0 +1,428 @@
...
+     * @param timer    the timer
+     * @param runnable the original runnable
+     * @return decorated runnable
+     */
+    static CheckedRunnable decorateCheckedRunnable(Timer timer, CheckedRunnable runnable) {
+        return () -> {
+            Context context = timer.createContext();
+            try {
+                runnable.run();
+                context.onSuccess(null);
```
RobWin, line 148: Let's change `context.onSuccess(null)` to `context.onSuccess()` to make it analog to CircuitBreaker or Retry.
In that chase the SUCCESS_RESULT_NAME_RESOLVER must not be able to handle null values.

RobWin (changes requested):

andrzejsydor (approved):

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/Timer.java
mkopylec: Done

resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/Timer.java
mkopylec: Instead of `onSuccess()` I added a `onNoResult()` method to indicate that this behaviour fires when the decorated operation returns nothing (is void). I can think of an alternative name like `onVoid()`. Which name do you prefer?

mkopylec: IMO "on success" refers to both, "on result" and "on no result".

RobWin: I prefer onSuccess(), because it is similar to all other implementations

mkopylec: Done

PR #576, opened 2019-08-17
#560 fix Feign fallback from lambda

hexmind:
#560

Issue #560, opened 2019-08-13
Fallback for with Feign does not works


harishmbm:
Hi,

In continuation of my previous post https://github.com/resilience4j/resilience4j/issues/559 I am trying to register a fallback with my Feign client. I am referring the example mentioned [here](https://github.com/resilience4j/resilience4j/tree/master/resilience4j-feign). Following is my config code

```
MyFeignClient requestFailedFallback = () -> "fallback from feign config";

        FeignDecorators decorators = FeignDecorators.builder()
                .withCircuitBreaker(circuitBreaker)
                .withFallback(requestFailedFallback)
                .build();
        return Resilience4jFeign.builder(decorators).contract(new SpringMvcContract())
                .target(MyFeignClient.class, "http://localhost:8080/resilience4j");
```

Whenever there are any exceptions from the Feign, I expect to see the response as "fallback from feign config" but I cannot see this message. I can see the following exception in logs:

```
java.lang.IllegalAccessException: Class io.github.resilience4j.feign.DefaultFallbackHandler can not access a member of class com.resilience4j.config.ResilienceFeignConfig$$Lambda$489/1253466407 with modifiers "public"
```

RobWin:
Hi,
a fallback must implement the interface `MyFeignClient`. Can you show me your Interface?

harishmbm:
Here it is

```
import org.springframework.web.bind.annotation.GetMapping;

public interface MyFeignClient {

    @GetMapping(path = "/sample")
    String feignMessage();

}
```

RobWin:
Why `@GetMapping`? Shouldn't it be `@RequestLine`?

harishmbm:
Hi RobWin,

The behaviour is same with `@RequestLine` as well. Can see the following exception in logs

```
java.lang.IllegalAccessException: Class io.github.resilience4j.feign.DefaultFallbackHandler can not access a member of class com.resilience4j.config.ResilienceFeignConfig$$Lambda$489/940088750 with modifiers "public"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102) ~[na:1.8.0_211]

```

RobWin:
@pwlan Do lamdas not work anymore as a fallback?

pwlan:
I will take a look.

pwlan:
@RobWin Ok, i can reproduce the bug. Could be an easy fix.

`fallbackMethod.setAccessible(true);`

@harishmbm Thanks for the feedback!

[...]

PR #1999, opened 2023-07-18
fix Feign fallback from lambda

kerwin612:
see #560 #2000
Optimization of #576


RobWin:
Hi,
could you please add a test to validate what has been fixed?

RobWin (approved):

RobWin:
Thank you

PR #2007, opened 2023-07-21
Issue #201 timeUnit replaced by currentTimestamp

hexmind:
Because of the currentTimestampFunction and MockClock it's still testable eg. AdaptiveBulkheadDurationTest


RobWin (changes requested):
I wonder why the execution duration parameter is called `start` or `startTime`.
Shouldn't it be `duration` like in the Circuitbreaker interface?
I actually would prefer to have analog interfaces and keep duration and timeUnit.

What's the mean reason why you want remove it?

hexmind:
Circuitbreaker had been created before currentTimestampFunction was added. 
Client instead of providing duration (that is equal difference 'finish-start' ) can provide only 'start' because component (AB) knows 'finish' time. 
To sum up from both client and component point od view it is simpler to provide only 1 argument 'start' instead of  calculating difference (duration) each time (using 2 arguments). 
Of course "Stateless" component can't keep start time (only 'now' equals finish time) to keep it even simpler...

RobWin (approved):

PR #2009, opened 2023-07-27
Timer reactive support

mkopylec:
https://github.com/resilience4j/resilience4j/issues/1625


resilience4j-reactor/build.gradle
```diff
@@ -5,11 +5,13 @@ dependencies {
...
     implementation project(':resilience4j-bulkhead')
     implementation project(':resilience4j-retry')
+    implementation project(':resilience4j-micrometer')
     testImplementation(libraries.reactor)
     testImplementation(libraries.reactor_test)
     testImplementation(libraries.assertj)
     testImplementation(libraries.reactive_streams_tck)
     testImplementation(libraries.blockhound)
     testImplementation project(':resilience4j-test')
+    testImplementation(project(':resilience4j-micrometer').sourceSets.test.output)
```
mkopylec, line 15: Add to use `io.github.resilience4j.micrometer.TimerAssertions` in tests.

resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/micrometer/operator/TimerSubscriber.java
```diff
@@ -0,0 +1,107 @@
...
+        context = requireNonNull(timer, "Timer is null").createContext();
+    }
+
+    @Override
+    protected void hookOnNext(T value) {
+        if (!isDisposed()) {
+            if (singleProducer) {
+                singleProducerResult.set(value);
+            } else {
+                multipleProducerResult.add(new ValueWrapper<>(value));
```
mkopylec, line 58: Concurrent Set is used to collect Flux elements. I'm not sure if this creates any risk of having concurrency problems while processing the reactive stream on a thread pool. I saw that there are Atomic References used all over the project so I decided to use the Concurrent Set.

Those collected elements are passed to Timer's `onResult(...)` method.

mkopylec:
@RobWin Can you take a look please?

RobWin:
Yes, I'm back from vacation in Monday

resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/micrometer/operator/TimerSubscriber.java
RobWin: I think for reactive streams we need to the change how the Timer is working.
The time shouldn't be stopped when a Flux item is emitted, but when the flux is completed successfully or with an error. 
We cannot not capture and store a (possible very long) stream of data in a Set just because of a tag name. 
The questions is how the `onResultTagResolver` should work. 
We could store the first or latest item in a AtomicReference and publish it to `onResult` when the Flux is completed.

Do you have any other smart idea?

RobWin: In our RetryOperator we have a similar issue to handle. 
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/retry/RetryOperator.java#L62

With `upstream.doOnNext(context::handleResult)`  we could store the latest item in some variable.

RobWin:
I added a comment

resilience4j-reactor/src/main/java/io/github/resilience4j/reactor/micrometer/operator/TimerSubscriber.java
mkopylec: >I think for reactive streams we need to the change how the Timer is working.
The time shouldn't be stopped when a Flux item is emitted, but when the flux is completed successfully or with an error.

This is exactly what I'm trying o achieve here. Every reactive stream (Flux, Flow, Observable, etc) should behave this way. Thus that's how I implemented it. If the implementation is somewhat wrong please let me know.

>The questions is how the onResultTagResolver should work.

Well, that's a hard one. Some options could be:
 - pass first and/or last item emitted
 - pass limited number of first/last items (use Set with limited size)
 - pass nothing
 - pass number of emitted items

mkopylec: Sure, that's true. We could also never fire the `onResultTagResolver` for multi-items streams/publishers. We can always fire the `onSuccessTagResolver` for those.

mkopylec: Ok, let's try to establish how`onResultTagResolver` should work. Are you OK with any of the options I proposed?

RobWin: Do you have a good usecase for the `onResultTagResolver`.
Otherwise we could also drop it and only provide a `onSuccessTagResolver`

mkopylec: At first I thought that it can be used for some business metrics. Suppose you have an inbound/incoming operation like "verify user credentials". Let's say it returns the number of already failed credentials verifications. You assume that if users fail to verify their credentials more than 5 times in row they become suspicious. You would like to measure how many suspicious and not suspicious users are trying to use your system. Now, you can do it either by creating a custom metric or by using `onResultTagResolver` which will check the returned number of already failed credentials verifications.

Now I'm not sure if this the right way to go. I guess these kind of business metrics should be handled separately.
If so, I think we should get remove the `onResultTagResolver` but also the `onSuccessTagResolver`.

RobWin: Yes, let's remove them. Makes it easier to go ahead

mkopylec: Ok.
Should I create a separate PR for the removal or just do it here?

RobWin: I would prefer this PR

mkopylec: Done

RobWin (approved):

RobWin:
Thank you

PR #2020, opened 2023-08-25
Timer spring support

mkopylec:
https://github.com/resilience4j/resilience4j/issues/1625


RobWin:
Hi, please tell me when I should start a review

mkopylec:
It's already ready for the review :) If it wasn't I would make it a draft.

mkopylec:
Can we make it this week? :)

mkopylec:
There is even more about the aspect order. Some users may want to measure the raw method execution as well as the overall time. They will actually need two timers on a single method. Maybe we can have many aspect orders for the timer decorator?
>What is your personal opinion why you have chosen to make the timer the clostest decorator to the method.

If you won't measure the raw method execution time you won't be able (in some cases) to actualy find out how long does the raw method execution really is. That's because other decorators like time limiter will cut of the execution after some treshold but the raw method execution may not stop. E.g. consider external HTTP request. If it takes 1s, time limiter is set to 500ms and the timer decorator is set after the time limiter then you will never know how long the request was (the time limiter won't kill the connection and the HTTP request processing will continue until the http client gets the response or times out).
>It would not measure the time the bulkhead, circuitbreaker or retry takes.

We should think how we would like the timer to behave when e.g. circuit breaker is open or bulkhead is full. If the timer will measure raw method it will not collect any metrics in that particular cases. If it will measure the overall execution it will collect times close to zero. For me it's probably better to measure nothing here, cause the method will never be executed in that cases.

RobWin:
Having the possibility to use the same decorator/annotation multiple times on the same method was discussed here: https://github.com/resilience4j/resilience4j/issues/643

I still like the idea very much, but unfortunately I never had time to help because of work, family and illness. 
But we can pick up the topic again if @walec51 is still interested and not angry about me that I didn't have time to respond and other duties in my live :/

PR #2035, opened 2023-10-19
Fix bug in Retry AsyncContext's onResult consumeResultBaeforeRetryAttempt

qnwlqnwlxm:
Fix bug in Retry AsyncContext's onResult consumeResultBaeforeRetryAttempt.

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
@@ -184,20 +184,18 @@ public class RetryImpl<T> implements Retry {
 
         @Override
         public boolean onResult(T result) {
-
             if (null != resultPredicate && resultPredicate.test(result)) {
                 totalAttemptsCounter.increment();
                 int currentNumOfAttempts = numOfAttempts.incrementAndGet();
-
                 if (currentNumOfAttempts >= maxAttempts) {
                     return false;
-                } else {
-                    if (consumeResultBeforeRetryAttempt != null) {
-                        consumeResultBeforeRetryAttempt.accept(currentNumOfAttempts, result);
-                    }
-                    waitIntervalAfterRuntimeException(currentNumOfAttempts, Either.right(result));
-                    return true;
                 }
+
+                if (consumeResultBeforeRetryAttempt != null) {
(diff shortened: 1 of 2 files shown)


resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
```diff
@@ -359,19 +357,20 @@ private long handleOnError(Throwable throwable) {
         public long onResult(T result) {
             if (null != resultPredicate && resultPredicate.test(result)) {
                 totalAttemptsCounter.increment();
-                int attempt = numOfAttempts.incrementAndGet();
-                if (attempt >= maxAttempts) {
-                    if(consumeResultBeforeRetryAttempt != null){
-                        consumeResultBeforeRetryAttempt.accept(attempt, result);
-                    }
```
injae-kim, line 366: (FYI) On `AsyncContextImpl`, `consumeResultBeforeRetryAttempt ` is only invoked when `attempt > maxAttempts` and it seems bug!

On `ContextImplt`,  `consumeResultBeforeRetryAttempt ` is invoked on each retry well.

injae-kim, line 366: (FYI) On `AsyncContextImpl`, `consumeResultBeforeRetryAttempt` is only invoked when `attempt >= maxAttempts` and it seems bug!

On `ContextImplt`,  `consumeResultBeforeRetryAttempt ` is invoked on each retry well.

qnwlqnwlxm:
@RobWin 
Sorry. Instead of running the entire test, I only ran the new test I added.
So I couldn't confirm the issue. Now i'm fix it. 

shouldMarkThreadInterruptedWhenInterruptedDuringRetry method of the SupplierRetryTest class interrupted the thread and did not clear it.
This was causing an InterruptedException in AsyncUtils.awaitResult when calling completionStage.toCompletableFuture().get(timeoutSeconds, TimeUnit.SECONDS).
The fix is to clear the interrupted status of the thread at the end of the shouldMarkThreadInterruptedWhenInterruptedDuringRetry test method.

cc. @injae-kim

RobWin (approved):

RobWin (approved):

Issue #2021, opened 2023-08-30
NPE for private field when triggering fallbackMethod from the Spring @Retry annotation


mikesjan:
Resilience4j version: 2.1.0

Java version: 17

We have a Spring component, that looks something like this:
```java
public class RetryingClient {

    private final Client client;
    private final ApplicationEventPublisher applicationEventPublisher;

    public RetryingClient(Client client, ApplicationEventPublisher applicationEventPublisher) {
        this.client = client;
        this.applicationEventPublisher = applicationEventPublisher;
    }

    @Retry(name = "client", fallbackMethod = "onGetStatusFailed")
    public ClientStatus getStatus() {
        ClientStatus result = client.getStatus();
        applicationEventPublisher.publishEvent(new ConnectionEvent(Status.UP));
        return result;
    }

    private AimStatus onGetStatusFailed(RuntimeException e) {
        applicationEventPublisher.publishEvent(new ConnectionEvent(Status.DOWN));
        throw e;
    }
```

It stopped working in version `2.1.0` with the following error: `Cannot invoke "org.springframework.context.ApplicationEventPublisher.publishEvent(Object)" because "this.applicationEventPublisher" is null"` when retrying fails (executing the `onGetStatusFailed` method).

After debugging, it seems like that the issue is because of this change https://github.com/resilience4j/resilience4j/compare/v2.0.2...v2.1.0#diff-5c30a60f9a316effc456a031670dcc1ea32617a820fb1ce36c44aa1be2a521f2. It uses proxy with all fields null instead of the target with the correctly initialized fields.

RobWin:
Hi,
damn thats correct. Seems our tests weren't sufficient :(

RobWin:
It was introduced by PR: https://github.com/resilience4j/resilience4j/pull/1965

@SOOHYUN-LIM Would you like to have a look at it? otherwise I have to revert this PR :(

mauriciogeneroso:
I was updating the library and found the issue is also happening when using the annotation `@CircuitBreaker` with `fallbackMethod`. It is using the proxy class and the instance objects are null causing `NullPointerException`.

E.g: `@CircuitBreaker(name = "name", fallbackMethod = "fallbackMethod")`

tushar-meesho:
Hey how to resolve this issue?

mauricio-sky:
[...]

PR #2058, opened 2023-11-15
FallbackMethod supports AOP

SOOHYUN-LIM:
- mr: https://github.com/resilience4j/resilience4j/pull/1965  
- bug: https://github.com/resilience4j/resilience4j/issues/2021  

@RobWin Sorry for the delay.  
I have fixed the bug mentioned below and added test cases.   

In Spring AOP, using private methods can be challenging due to the limitations of proxy. To address this issue, I employed different objects based on access modifiers.

please confirm that PR


RobWin (approved):

cmakinen:
When is a release with this fix anticipated? We're facing quite a few issues due to this specific NPE bug too.

mauriciogeneroso:
@cmakinen You need to stay on the previous version 2.0.2 until it be fixed if you wanna use this library without the NPE. That is what my team is doing while we wait for a new release.

cmakinen:
Yes @mauriciogeneroso, that is what we're doing in the interim too.

Issue #2038, opened 2023-10-24
java.lang.StackOverflowError due to infinite loop in CircuitBreakerStateMachine$OpenState.tryAcquirePermission


polkosity:
Resilience4j version: 2.1.0

Java version: any

I've encountered java.lang.StackOverflowError due to infinite loop in CircuitBreakerStateMachine$OpenState.tryAcquirePermission in a production server, which was under strain.

After reviewing the code, I suggest it's caused by failure in the AtomicReference getAndUpdate call (CircuitBreakerStateMachine.java:333), as such failures can occurr due to contention among threads, according to Javadoc.  As an aside, Javadoc also states that the updateFunction should be side-effect free, which is not the case here.

Once the updateFunction fails during an attempted transition from open to half open state, the AtomicReference CircuitBreakerStateMachine.stateReference still refers to an OpenState whose AtomicBoolean OpenState.isOpen has already been set to false (CircuitBreakerStateMachine.java:812), so the transitionToHalfOpenState() call will not be attempted again, additionally retryAfterWaitDuration is already in the past for this OpenState object.  The effect is that OpenState.tryAcquirePermission calls itself recursively (CircuitBreakerStateMachine.java:737) until the StackOverflowError occurs.  

Please see the attached screenshot displaying a partial stack trace of the StackOverflowError, which was generated under resilience4j-circuitbreaker v1.7.1, but the relevant code sections have not changed since then. 

<img width="1011" alt="StackOverflowError" src="https://github.com/resilience4j/resilience4j/assets/6734401/9fd3fa9a-623a-48f6-ae9b-717f2b5ec2b6">

RobWin:
Hi,
thank you for the analysis.
First time that I see this.  Do you have a proposal how to fix it?

storozhukBM:
@polkosity 
https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/atomic/AtomicReference.html#getAndUpdate(java.util.function.UnaryOperator) can't fail, it will be retried until update is successful.

I think the problem is here https://github.com/resilience4j/resilience4j/blob/12f66a31d7b4619ed0ffd1fddee9da683c41b9b7/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java#L812

[...]

PR #2072, opened 2023-12-05
Fix StackOverflowError after scheduled future interruption

Hartigan:
[issue](https://github.com/resilience4j/resilience4j/issues/2038)

Changes:

1. Do not interrupt future if it is running
2. Remove `synchronized`, because code already synchronized with atomic change gate (REVERTED)


fatso83:
If this works, we are really wanting the fix. The bug breaks our production systems non-stop, so we had to remove our circuit breakers. Please review 🙏

RobWin (approved):

fatso83:
Fails CI: `> Cannot perform signing task ':resilience4j-all:signResilience4jPublication' because it has no configured signatory`

PR #2074, opened 2023-12-11
avoid to add  duplicate consumer

ShijunDeng:
CopyOnWriteArrayList may lead to add same eventConsumer and repeatedly publish events


RobWin (approved):

RobWin:
I'm not sure how to handle this change.
Actually it was be design that you can register multiple consumer for the same event. It's up to the client to make sure that not multiple consumer are registered.
This would be a non-backwards compatible change.

ShijunDeng:
> I'm not sure how to handle this change. Actually it was be design that you can register multiple consumer for the same event. It's up to the client to make sure that not multiple consumer are registered. This would be a non-backwards compatible API change.

yes，with `MAP` we can register multiple consumer, but it's should not be  duplicate consumer

PR #2084, opened 2024-01-02
feat(Retry): Added decorate as instance method

### LINKED ISSUE #2029: Include `decorateSupplier` as `Retry` instance method
Resilience4j version: 1.7.1

Java version: 17+

Thanks for raising a Resilience4j issue.
Please replace this text with a brief description of your problem along with the versions you are using.
If possible, please also consider putting together a complete JUnit test that reproduces the issue.

Similar to #325 & #485 could you include a `decorateSupplier` method in the `Retry` interface.

## Current  behavior

Current code looks like this:
```java
return circuitBreaker.executeSupplier(
                Retry.decorateSupplier(retry, responseSupplier)
        );
```

I have bypass this limitation by using an intermediate supplier

```java
return circuitBreaker.executeSupplier(
                // Avoid static reference to `Retry.decorateSupplier`
                () -> retry.executeSupplier(responseSupplier)
        );
```

## Desired behavior

I would like to avoid both static and unnecessary supplier.

```java
  return circuitBreaker.executeSupplier(
                retry.decorateSupplier(responseSupplier)
        );
```

If you agree, I could open the PR. Thanks.

jdvr:
fixes #2029

Added `decorate*` as instance method to reduce some boilerplate. I originally suggested just the `decorateSupplier` but I think that all method should have the same behavior.

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/Retry.java
@@ -119,6 +119,21 @@ public interface Retry {
     }
 
     /**
+     * Decorates CompletionStageSupplier using current instance as context.
+     *
+     * @param scheduler execution service to use to schedule retries
+     * @param supplier  completion stage supplier
+     * @param <T>       type of completion stage result
+     * @return decorated supplier
+     */
+    default <T> Supplier<CompletionStage<T>> decorateCompletionStage(
+            ScheduledExecutorService scheduler,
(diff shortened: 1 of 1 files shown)


RobWin (approved):

jdvr:
@RobWin I have removed `decorateFunction` because it breaks kotlin package public interface.

RobWin (approved):

jdvr:
@RobWin Do I need to do something else to merge this one?

Issue #752, opened 2019-11-18
startTime property in RateLimiter


walec51:
I propose to add a startTime property to RateLimiterConfig so that we can set from when does a RateLimiting period start. This is necessary if someone wants to user RateLimiter both on his server and client. If you have your clocks in synce then with this property one could make limit period overlap on the server and client.

RobWin:
Hi,

yes, good idea.

RobWin:
FYI @storozhukBM

injae-kim:
Hi! I created PR https://github.com/resilience4j/resilience4j/pull/2100 to fix this issue. PTAL when you have some times~ thanks! 😃

Issue #1668, opened 2022-04-19
Fixed window counter implementation


Misha123456:
Resilience4j version: 1.7.0

Java version: 11

Hey, guys. 
How could I get the "Fixed window counter" implementation based on your library? 

The API I need to apply a rate limiter for, is using the "Fixed window counter" algorithm. 
Let's say it allows 10 requests each minute. 
So, within the 2000-01-01T10:00:00 - 2000-01-01T10:01:00 time frame, I can do 10 requests. 
Same for 2000-01-01T10:01:00, ends 2000-01-01T10:02:00. 
And so on. 

When I use AtomicRateLimiter, I can't control the "window" start time. As far as I understood, it depends on class loading time. So, if I configure it to use 1 minute "window" with limitRefreshPeriod(Duration.ofMinutes(1)), and class was loaded 2000-01-01T10:00:29, then "window" will be 2000-01-01T10:00:29 - 2000-01-01T10:01:29. But I need it to be 2000-01-01T10:00:00 - 2000-01-01T10:01:00. 
So, what could I do? 

I mean, except reflection to adjust AtomicRateLimiter.nanoTimeStart :)

RobWin:
Hi,

https://github.com/resilience4j/resilience4j/issues/752

Would you like to create a PR?

injae-kim:
Hi! I created PR https://github.com/resilience4j/resilience4j/pull/2100 to fix this issue. PTAL when you have some times~ thanks! 😃

PR #2100, opened 2024-01-23
Add `startedTime` configuration on `RateLimiter`

injae-kim:
Fixes gh-752, gh-1668

### Motivation
```
The API I need to apply a rate limiter for, is using the "Fixed window counter" algorithm.
Let's say it allows 10 requests each minute.
So, within the 2000-01-01T10:00:00 - 2000-01-01T10:01:00 time frame, I can do 10 requests.

This is necessary if someone wants to user RateLimiter both on his server and client 
... could make limit period overlap on the server and client.
```
- We found that many users want to set some `startedTime` on `RateLimiter`
  - Some user wants to start rate limiter on every minute's start
  - Some user uses rate limiter on both server & client, and wants to sync them to overlap time window

### Modification
- Add `startedTime` configuration on `RateLimiter`

### Result
- Now user can set `startedTime` configuration on `RateLimiter`
- Closes gh-752, gh-1668


resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/RateLimiterConfig.java
```diff
@@ -117,6 +124,14 @@ private static int checkLimitForPeriod(final int limitForPeriod) {
         return limitForPeriod;
     }
 
+    private static Instant checkStartedTime(final Instant startedTime) {
+        requireNonNull(startedTime, STARTED_TIME_MUST_NOT_BE_NULL);
+        if (startedTime.isAfter(Instant.now())) {
+            throw new IllegalArgumentException("StartedTime should be earlier or equal than now");
+        }
+        return startedTime;
+    }
```
injae-kim, line 133: FYI) if we allow to set `startTime` that later than now, we need to fix lots of codes on `AtomicRateLimiter`.

https://github.com/resilience4j/resilience4j/blob/24798b3aa800be28e28207da33f717583179be82/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java#L66
https://github.com/resilience4j/resilience4j/blob/24798b3aa800be28e28207da33f717583179be82/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java#L103-L108

Cause `AtomicRateLimiter` assumes that `startTime` is the time when class is loaded and always `now() > startTime`

https://github.com/resilience4j/resilience4j/blob/24798b3aa800be28e28207da33f717583179be82/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java#L222-L223

And it calculate `currentCycle` based on `currentNanoTime()` (elapsed time after class loaded), so if we set `future` as start time here then `currentCycle` is negative and limiter not works well.

Also, I think it's unclear what limiter should do when `now()` is before `startTime`.

So I named it as `startedTime` and want to set instant that earlier or equal than now~!

PR #2108, opened 2024-02-07
check randomization factor in all root level constructors

jdcaperon:
In this constructor the randomization factor is not checked until call time, this means that the IntervalFunction can throw runtime exceptions after the application has started. This is undesirable as it is generally better for something to fail at startup rather than when the system is already retrying due to other failures. This also brings it into line with the behavior of other constructors.


.idea/codeStyles/Project.xml
```diff
@@ -3,6 +3,9 @@
     <JetCodeStyleSettings>
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </JetCodeStyleSettings>
+    <ScalaCodeStyleSettings>
+      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
+    </ScalaCodeStyleSettings>
```
jdcaperon, line 8: I'll remove this.

jdcaperon, line 8:
```suggestion
```

jdcaperon:
cc @RobWin do you mind running the CI again so I can get this green?

jdcaperon:
@RobWin green now

RobWin (approved):

PR #2113, opened 2024-02-08
Update Feign version

mateusz-bajorek:
Update to the latest Feign version (13.2.1).

Fix `Resilience4jFeign`: Feign 12.5 introduced a change to the `Builder` type, making the `build()` method `final` and instead adding `abstract internalBuild()` that now has to be overridden.

### DIFF
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -22,7 +22,7 @@ ext {
     springBoot3Version = '3.0.0'
     springCloudContext2Version = '3.1.5'
     spockVersion = '1.3-groovy-2.5'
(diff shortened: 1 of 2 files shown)


nilo85:
Meh, found this a little bit too late, I just pushed a PR solving the same problem, however I think using the capability api is leaner and compatible with more versions of Feign (10.9+)

https://github.com/resilience4j/resilience4j/pull/2127/files

mateusz-bajorek:
Ah, your solution @nilo85 is indeed better. I am closing this and hoping for #2127 to be released as soon as possible :)

PR #2117, opened 2024-02-14
Removed the synchronized block because an atomic boolean is already used

koundinyagoparaju:


jatin-bansal-21:
Hi, does the use of synchronized block actually matter here because these operations of circuit breaker are in memory and not I/O operations? 

Also, I am not sure if there are any synchronized blocks when I am decorating my I/O operations with circuit breaker.

dowenliu-xyz:
Notice that the `if` block of the method body itself is **not** atomic (it is compiled as a serial number of instructions). The original `synchronized` keyword guarantees the atomicity of the entire method.

koundinyagoparaju:
> Notice that the `if` block of the method body itself is **not** atomic (it is compiled as a serial number of instructions). The original `synchronized` keyword guarantees the atomicity of the entire method.

Sry for a n00b question :( 
Would the compareAndSet in if condition does not guarantee that the if block is executed atomically?

dowenliu-xyz:
I rechecked the code. 
Since there is no attempt anywhere to change the value from false to true, it should be guaranteed here that only one thread will execute the `transitionToHalfOpenState` method in the code block.
However, I'm not sure if there is any problem if there are multiple threads concurrently executing to the `toHalfOpenState` method. Assume the following scenario:
Thread A executes the if condition and gets true into the if block, but the OS/JVM hangs it; Thread B gets a chance to execute the if condition, but since isOpen is already false, B doesn't get into the if condition. Thread A will continue to execute the transitionToHalfOpenState method after recovery, but the question is, will B, which is not executing that method, have problems executing after toHalfOpenState? I'm not sure.

PR #2127, opened 2024-03-05
Add support for Feign 12.5+

nilo85:
Feign 12.5 made the builder.build() method final, and started internally clone it...

This breaks resilience4j integration with feign and renders  resilience4j completely unused by feign.

In version 10.9 they introduced an api to add "capabilities" to the feign client.

Here I have made changed to the 1.7 branch (the release I need) that implements this new pattern instead, and provides a fallback using old pattern for backwards compatibility down to 10.9.

Merging this means minimum version of feign to use would be 10.9. and works with latest

### DIFF
--- a/libraries.gradle
+++ b/libraries.gradle
@@ -22,7 +22,7 @@ ext {
     ratpackVersion = '1.7.6'
     spockVersion = '1.3-groovy-2.5'
     retrofitVersion = '2.5.0'
-    feignVersion = '10.2.0'
+    feignVersion = '10.9'
     prometheusSimpleClientVersion = '0.9.0'
     reactiveStreamsVersion = '1.0.2'
     micrometerVersion = '1.5.1'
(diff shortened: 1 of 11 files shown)


nilo85:
This is applied to the 1.7 branch, as I need this fixed in 1.7, if this is approved and we create a new 1.7.x release I am happy to port these changes to 2.0 branch =)

mateusz-bajorek:
Yes, porting this to 2.x would be appreciated as well :)

andrzejsydor (approved):

nilo85:
follow up for 2.0 branch https://github.com/resilience4j/resilience4j/pull/2129 (simple cherry-pick seems to be enough)

nilo85:
follow up for master branch https://github.com/resilience4j/resilience4j/pull/2130

AlessandroBagnoli (approved):

PR #2130, opened 2024-03-06
Master branch - Add support for Fegin 12.5+

nilo85:
Follow up from https://github.com/resilience4j/resilience4j/pull/2127

Cherry picked and addressed merge conflicts, TLDR; 
* you had bumped feign to 12.0 so no need for me to modify libraries
* Test file had conflict where you had previously copied the entire long exception where I went for startsWith()

```
BUILD SUCCESSFUL in 5m 21s
214 actionable tasks: 214 executed
```

Feign 12.5 made the builder.build() method final, and started internally clone it...

This breaks resilience4j integration with feign and renders resilience4j completely unused by feign.

In version 10.9 they introduced an api to add "capabilities" to the feign client.

Here I have made changed to the 1.7 branch (the release I need) that implements this new pattern instead, and provides a fallback using old pattern for backwards compatibility down to 10.9.

Merging this means minimum version of feign to use would be 10.9. and works with latest

### DIFF
--- a/resilience4j-feign/README.adoc
+++ b/resilience4j-feign/README.adoc
@@ -14,8 +14,9 @@ resilience4j-feign makes it easy to incorporate "fault tolerance" patterns into 
  
 == Decorating Feign Interfaces
 
-The `Resilience4jFeign.builder` is the main class for creating fault tolerance instances of feign. 
-It extends the `Feign.builder` and can be configured in the same way with the exception of adding a custom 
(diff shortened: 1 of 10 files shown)


nilo85:
Force pushed a new commit message mentioning this cherry-pick results in no change of feign version

AlessandroBagnoli:
Any chance we can get this released?

Issue #2215, opened 2024-10-03
com.test.Connector#getToken has orted by @TimeLimiter return type. CompletionStage expected


rcbandit111:
Resilience4j version: 2.1.0

Java version: 21

I have the following Java code into Spring Boot microservice:

      @TimeLimiter(name = "getToken")
      @CircuitBreaker(name = "getToken", fallbackMethod = "fallbackRequestException")
      public ResponseEntity<Resources> getHandleByToken(HttpHeaders headers, String url) {
        return getByToken(headers, url);
      }
    
    .......
    
    public ResponseEntity<Resources> getByToken(HttpHeaders headers, String url) {
        return invoke(null, url);
      }

I get this error:

    io.github.resilience4j.spring6.timelimiter.configure.IllegalReturnTypeException: org.springframework.http.ResponseEntity com.test.Connector#getHandleByToken has orted by @TimeLimiter return type. CompletionStage expected.
            at io.github.resilience4j.spring6.timelimiter.configure.TimeLimiterAspect.proceed(TimeLimiterAspect.java:106) ~[resilience4j-spring6-2.1.0.jar!/:2.1.0]
            at io.github.resilience4j.spring6.timelimiter.configure.TimeLimiterAspect.lambda$timeLimiterAroundAdvice$0(TimeLimiterAspect.java:90) ~[resilience4j-spring6-2.1.0.jar!/:2.1.0]
            at io.github.resilience4j.spring6.fallback.FallbackExecutor.execute(FallbackExecutor.java:37) ~[resilience4j-spring6-2.1.0.jar!/:2.1.0]
            at io.github.resilience4j.spring6.timelimiter.configure.TimeLimiterAspect.timeLimiterAroundAdvice(TimeLimiterAspect.java:91) ~[resilience4j-spring6-2.1.0.jar!/:2.1.0]
            at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) ~[na:na]
            at java.base/java.lang.reflect.Method.invoke(Method.java:580) ~[na:na]

I tried to add:

    @Bulkhead(name = "getToken", type = Bulkhead.Type.THREADPOOL)

But still I get the same error. Any way to solve the issue without code refactoring into Java methods? Some solution to add or change annotation configurations?

rcbandit111:
@RobWin Can you advice, please?

RobWin:
Hi,
[...]

PR #2229, opened 2024-10-27
Feature #2215: Bulkhead Sync Return Support

charb29:
This feature enhances the `Bulkhead` functionality by allowing it to return plain objects. Previously, the `Bulkhead` was limited to only returning `CompletableFuture` and reactive types such as `Mono` and `Flux`. For further context, please refer to the related issue: [Issue #2215](https://github.com/resilience4j/resilience4j/issues/2215)

### Key Enhancements
1. `PlainObjectBulkheadAspectExt`:
This class provides an aspect for managing method executions that return plain objects while enforcing `Bulkhead` constraints.

2. `ThreadPoolBulkheadAdapter`:
This new class adapts a `ThreadPoolBulkhead` to interface as a Bulkhead for the `bulkheadAspectExt.handle()` method. It facilitates seamless integration with existing bulkhead aspects while retaining the core functionalities of a thread pool bulkhead.


resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/ThreadPoolBulkheadAdapter.java
```diff
@@ -0,0 +1,66 @@
+package io.github.resilience4j.bulkhead.internal;
+
+import io.github.resilience4j.bulkhead.Bulkhead;
+import io.github.resilience4j.bulkhead.BulkheadConfig;
+import io.github.resilience4j.bulkhead.ThreadPoolBulkhead;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+public record ThreadPoolBulkheadAdapter(ThreadPoolBulkhead threadPoolBulkhead) implements Bulkhead {
```
RobWin, line 10: Could you please explain why this class/record is needed?

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/timelimiter/autoconfigure/AbstractTimeLimiterConfigurationOnMissingBean.java
```diff
@@ -58,6 +58,12 @@ public CompositeCustomizer<TimeLimiterConfigCustomizer> compositeTimeLimiterCust
         return new CompositeCustomizer<>(customizers);
     }
 
+    @Bean
+    @ConditionalOnMissingBean
+    public TimeLimiterProperties timeLimiterProperties() {
```
RobWin, line 63: Why do we need to add TimeLimiterProperties here?

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public class PlainObjectBulkheadAspectExt implements BulkheadAspectExt {
+
+    private static final Logger logger = LoggerFactory.getLogger(PlainObjectBulkheadAspectExt.class);
+    private final TimeLimiterRegistry timeLimiterRegistry;
+    private final ScheduledExecutorService executorService;
+
+    public PlainObjectBulkheadAspectExt(TimeLimiterRegistry timeLimiterRegistry) {
+        this.timeLimiterRegistry = timeLimiterRegistry;
+        this.executorService = Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
```
RobWin, line 30: Have a look here: https://github.com/resilience4j/resilience4j/blob/master/resilience4j-spring/src/main/java/io/github/resilience4j/timelimiter/configure/TimeLimiterAspect.java#L54-L68

The TimeLimiter needs a special `ContextAwareScheduledThreadPoolExecutor` so that context information can be passed from Thread to Thread

resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public class PlainObjectBulkheadAspectExt implements BulkheadAspectExt {
+
+    private static final Logger logger = LoggerFactory.getLogger(PlainObjectBulkheadAspectExt.class);
+    private final TimeLimiterRegistry timeLimiterRegistry;
+    private final ScheduledExecutorService executorService;
+
+    public PlainObjectBulkheadAspectExt(TimeLimiterRegistry timeLimiterRegistry) {
+        this.timeLimiterRegistry = timeLimiterRegistry;
+        this.executorService = Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
```
RobWin, line 30: See review comment above

RobWin (changes requested):

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,139 @@
...
+            return handleThreadPoolBulkhead(proceedingJoinPoint, threadPoolBulkhead, methodName);
+        }
+
+        return handleSemaphoreBulkhead(proceedingJoinPoint, bulkhead, methodName);
+    }
+
+    private Object handleSemaphoreBulkhead(ProceedingJoinPoint proceedingJoinPoint, Bulkhead bulkhead, String methodName) {
+        TimeLimiter timeLimiter = timeLimiterRegistry.timeLimiter(bulkhead.getName());
+        try {
+            Supplier<CompletableFuture<Object>> futureSupplier = createBulkheadFuture(proceedingJoinPoint, bulkhead);
```
RobWin, line 69: You cannot time out a SemphoreBulkhead, since it is not running on a separate thread.
The SemphoreBulkhead already returns a plain object type.  It is handled here: https://github.com/resilience4j/resilience4j/blob/ce6cf63905099b57daec391a6b0a0d6dabbbbcf5/resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java#L194

If we want to time out a call which is waiting for a permission, we would need to work with Semaphore timeouts.

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,139 @@
...
+        } catch (Throwable ex) {
+            logGenericException(methodName, ex);
+            throw ex;
+        }
+    }
+
+    private Object handleThreadPoolBulkhead(ProceedingJoinPoint proceedingJoinPoint, ThreadPoolBulkhead threadPoolBulkhead, String methodName) {
+        TimeLimiter timeLimiter = timeLimiterRegistry.timeLimiter(threadPoolBulkhead.getName());
+        try {
+            CompletableFuture<Object> completableFuture = threadPoolBulkhead.executeCallable(() -> {
```
RobWin, line 88: This is wrong.
It already executes the call before you have wrapped it with a TimeLimiter.
Use `decorateCallable` instead

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,139 @@
...
+                try {
+                    return proceedingJoinPoint.proceed();
+                } catch (Throwable throwable) {
+                    return completeFutureExceptionally(throwable);
+                }
+            }).toCompletableFuture();
+
+            return timeLimiter.executeCompletionStage(executorService, () -> completableFuture)
```
RobWin, line 96: You can skip `() -> completableFuture`, if you use decorateCallable above

RobWin (changes requested):

resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/internal/ThreadPoolBulkheadAdapter.java
charb29: I created that class as an adapter so that we didn't have to create a separate `ThreadPoolBulkheadAspectExt` interface and implementation and to avoid having to complicate the `PlainObjectBulkheadAspectExt` class by needing to handle both bulkhead types directly.

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/timelimiter/autoconfigure/AbstractTimeLimiterConfigurationOnMissingBean.java
charb29: I added the `TimeLimiterProperties` bean because the `PlainObjectBulkheadAspectExt` depends on the `TimeLimiterRegistry`, the `TimeLimiterRegistry` itself requires `TimeLimiterConfigurationProperties` (which extends `TimeLimiterProperties`) for its configuration, and without this bean declaration, the dependancy chain breaks and causes failures in multiple tests, specifically in `BulkheadConfigurationOnMissingBeanTest`

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Determines if the aspect can handle the specified return type of the method.
+     *
+     * @param returnType the AOP method return type class
+     * @return {@code true} if the method has a plain object return type; {@code false} otherwise.
+     */
+    @Override
+    @SuppressWarnings("rawtypes")
+    public boolean canHandleReturnType(Class returnType) {
+        return !returnType.equals(CompletableFuture.class)
```
RobWin, line 49: Here we need to be careful.
What happens if Spring Reactor is not on the classpath? If someone is not using Spring Reactor this might not work.
Another idea could be to have a list of BulkheadExtensions which are tested in a row. It's called the Chain of Responsibility pattern.
If we make sure that the ComplatebleFuture or Spring Reactor, RxJava Extensions are tested first and this is the last extension in the chain as a fallback.

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
```diff
@@ -0,0 +1,152 @@
...
+     * Handle the execution of a method wrapped in a bulkhead context.
+     *
+     * @param proceedingJoinPoint Spring AOP proceedingJoinPoint
+     * @param bulkhead the configured bulkhead
+     * @param methodName the method name
+     * @return the result object from the method execution
+     * @throws Throwable if an exception occurs during the method execution
+     */
+    @Override
+    public Object handle(ProceedingJoinPoint proceedingJoinPoint, Bulkhead bulkhead, String methodName) throws Throwable {
```
RobWin, line 64: We don't need to handle the SemaphoreBulkhead in this extensions. 
SemaphoreBulkhead  is handled in the BulkheadAspect.

RobWin (changes requested):

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java
```diff
@@ -229,7 +230,20 @@ private Object proceedInThreadPoolBulkhead(ProceedingJoinPoint proceedingJoinPoi
...
                 backend);
         }
+
         ThreadPoolBulkhead threadPoolBulkhead = threadPoolBulkheadRegistry.bulkhead(backend);
+        if (bulkheadAspectExts != null && !bulkheadAspectExts.isEmpty()) {
+            for (BulkheadAspectExt bulkheadAspectExt : bulkheadAspectExts) {
+                if (bulkheadAspectExt.canHandleReturnType(returnType)) {
+                    return bulkheadAspectExt.handle(
+                            proceedingJoinPoint,
+                            new ThreadPoolBulkheadAdapter(threadPoolBulkhead),
```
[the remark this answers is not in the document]
charb29: It was the best approach I was able to come up with, without the need to modify the `BulkheadAspectExt` interface. I thought of casting the `bulkhead` to a `ThreadPoolBulkhead` object but that causes a `ClassCastException` for non-null values.

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/PlainObjectBulkheadAspectExt.java
charb29: Would adding the `Order()` annotation to the implementations of `BulkheadAspectExt` be a viable option for this?

resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadAspect.java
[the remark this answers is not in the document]
charb29: I do see the issues now with the use of the `ThreadPoolBulkheadAdapter`. If I remove the `handleSemaphoreBulkhead` method, it causes the `testAsyncRecovery` method inside of `BulkheadRecoveryTest` to fail. I can create a different implementation for `ThreadPoolBulkheads` rather than using an adapter.

RobWin (changes requested):

Issue #2200, opened 2024-09-18
ignoreExceptions spring boot configuration - possible to ignore unknown classes to not fail on startup?


cabbonizio:
Resilience4j version: 2.2.0
Java version: 17

Hi, this is a general question.   We make use of the "ignoreExceptions" property for circuit breakers and our application exception classes underwent a package change.   When we deployed the code (with the new package location) but the config with the old location, it failed on startup due to r4j calling out the class is not found (which makes sense).

com.sample.app.SampleException (old)
com.sample.app.new.SampleException (new)

We have our app configuration externalized not part of the build, so it's quite the timing issue trying to match up the code and configuration to ensure things so smooth.   We have some ways to mitigate this, but my question is can we "tune" r4j to "ignore" an unknown classname in the "ignoreExceptions" list so it does not fail on startup?

```
resilience4j:
  circuitbreaker:
      global:
        ignoreExceptions:
          -  com.sample.app.SampleException
          -  com.sample.app.new.SampleException
```

Many thanks and I hope I was clear on the ask!

charb29:
Hi @cabbonizio, I'd be glad to work on this!

cabbonizio:
> Hi @cabbonizio, I can work on this issue/feature!

@charb29 many thanks for offering to do this.  It will help deal with compatibility issues.   Please let me know how else I can help.

charb29:
> @charb29 many thanks for offering to do this. It will help deal with compatibility issues. Please let me know how else I can help.

@cabbonizio no worries. I'll keep you updated on the progress.

cabbonizio:
Thoughts on whether to make this with a config flag so devs can opt-in or simply make it default behavior to ignore classes not found on classpath?

charb29:
@cabbonizio I'm working on replicating the issue at the moment, so I'm not entirely sure yet how we'll solve this. However, combining the default behaviour with a flag to try and provide the best of both worlds.

After speaking with @RobWin though, it's a priority that the new feature is backwards compatible.

cabbonizio:
> @cabbonizio I'm working on replicating the issue at the moment, so I'm not entirely sure yet how we'll solve this. However, combining the default behaviour with a flag to try and provide the best of both worlds.
> 
> After speaking with @RobWin though, it's a priority that the new feature is backwards compatible.

Makes sense @charb29.

charb29:
@cabbonizio
When you get the error on the application startup, does it look similar to this?

```
***************************
[...]

PR #2233, opened 2024-11-03
Feature #2200: Ignore Unknown Exceptions in Circuit Breaker Configuration

charb29:
This feature gives the Circuit Breaker the ability to ignore unknown exceptions in the `ignoreExceptions` field. To avoid `ClassNotFoundException` during compilation when `ignoreUnknownExceptions` is set to `true`, a custom converter has been implemented to convert any unknown exception classes to a placeholder exception, preventing compilation issues. If the `ignoreUnknownExceptions` field is not defined, it defaults to `false`, resulting in the usual compilation error.

For further context, please refer to the related issue: [Issue #2200](https://github.com/resilience4j/resilience4j/issues/2200)

When `ignoreUnknownExceptions` is set to true, users will be notified during compile time through logs.

To use the new property, define `ignoreUnknownExceptions` in your configuration file as follows:
```
resilience4j:
  circuitbreaker:
      global:
        ignoreUnknownExceptions: true
        ignoreExceptions:
          -  com.sample.app.SampleException
          -  com.sample.app.new.SampleException
 ```

### CODE REVIEW

File: resilience4j-circuitbreaker/build.gradle [hunk 1]
```diff
@@ -1,5 +1,6 @@
 dependencies {
     api project(':resilience4j-core')
+    implementation 'org.springframework.boot:spring-boot-starter:2.5.5'
```
RobWin, line 3: Hi,

resilience4j-circuitbreaker must not have a dependency to spring-boot-starter.
The library is framework agnostic and should not introduce any framework-specific dependencies.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -169,6 +171,10 @@ public boolean isWritableStackTraceEnabled() {
         return writableStackTraceEnabled;
     }
 
+    public boolean isIgnoreUnknownExceptionsEnabled() {
+        return ignoreUnknownExceptions;
```
RobWin, line 175: Let's rename it ignoreClassBindingExceptions or ignoreClassConversationExceptions.
Otherwise people thing it means the property is used to configure that the CircuitBreaker would ignore any unknown exception which is not the case.

RobWin (changes requested):

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -171,6 +173,10 @@ public boolean isWritableStackTraceEnabled() {
         return writableStackTraceEnabled;
     }
 
+    public boolean ignoreClassBindingExceptions() {
+        return ignoreUnknownExceptions;
```
RobWin, line 177: I like the name `ignoreClassBindingExceptions`
Can we also rename the variables and constants?

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -711,6 +719,27 @@ public Builder ignoreException(Predicate<Throwable> predicate) {
...
+         *
+         * <p>
+         *     The default value is {@code false}, meaning unknown exceptions will trigger a {@code ClassNotFoundException}
+         *     during startup.
+         * </p>
+         *
+         * @param ignoreUnknownExceptions {@code true} to ignore unknown exceptions during startup.
+         * @return CircuitBreakerConfig.Builder
+         */
+        public final Builder ignoreUnknownExceptions(@Nullable boolean ignoreUnknownExceptions) {
```
RobWin, line 738: Let's use ignoreClassBindingExceptions.
It does not configure that the CircuitBreaker ignores unknown exceptions. This is confusing. 
IgnoreExceptions configures which exceptions are ignore at runtime to secure a HTTP.
`ignoreClassBindingExceptions` is a setting which should not effect the runtime behavior of the CircuitBreaker, but instead make sure that Classes which are not found are ignored and don't lead to a failure at startup phase..

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerTest.java
```diff
@@ -1194,6 +1195,41 @@ public void shouldDecorateFutureAndReturnWithCallNotPermittedException() {
         assertThat(metrics.getNumberOfNotPermittedCalls()).isEqualTo(1);
     }
 
+    @Test
+    public void shouldIgnoreUnknownExceptionsWhenSetToTrue() {
```
RobWin, line 1199: I don't think that these tests are helpful.
The simulateApplicationStartup is not testing any internal logic of the CircuitBreaker.

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
RobWin: Actually I think we should not add this config parameter to the CircuitBreakerConfig.
Since it does not change the behavior of the CircuitBreaker at all.
If the config parameter is only useful in frameworks such as Spring Boot, we should move the config property to the CommonCircuitBreakerConfigurationProperties

RobWin (changes requested):
Thank you.
I have some small findings

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
```diff
@@ -39,6 +39,7 @@ public class CircuitBreakerConfig implements Serializable {
 
     private static final long serialVersionUID = -5429814941777001669L;
 
+    public static final boolean DEFAULT_IGNORE_UNKNOWN_EXCEPTIONS_ENABLED = false;
```
RobWin, line 42: Let's move this completly out of the CircuitBreakerConfig. It's not needed anymore, correct?

RobWin (changes requested):

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
charb29: Correct, I must have missed it in my previous commit. Will remove it.

RobWin (approved):

RobWin (approved):

PR #2249, opened 2024-11-28
Closes #Issue 2246 and adds handling for the case non checked retry consumer is interpted on sleep

nikolayandr:


RobWin (approved):

RobWin:
Thank you.That was fast

Issue #1699, opened 2022-05-30
Disable CircuitBreaker by config in different environments 


AlexBar56:
Resilience4j version: 1.7.1

Java version: 1.8

Hi Team,

For my springbok app I have implemented circuitbreaker using the @CircuitBreaker annotation and in the yml file I have added the configuration. But my question is there any property available to enable/disable circuitBreaker for different environments? 
If you could please help me with this that would be really helpful.

Thank you.

Alex

tylerlance:
You can inject the CircuitBreakerRegistry bean and properties and pull off any circuit breakers with a custom state defined in the properties. Then do something like this:
```
private void transitionCircuitBreakerState(CircuitBreaker circuitBreaker, CircuitBreaker.State state) {
        switch(state) {
            case CLOSED:
                circuitBreaker.transitionToClosedState();
                break;
            case OPEN:
                circuitBreaker.transitionToOpenState();
                break;
            case HALF_OPEN:
                circuitBreaker.transitionToOpenState();
                circuitBreaker.transitionToHalfOpenState();
                break;
            case FORCED_OPEN:
                circuitBreaker.transitionToForcedOpenState();
                break;
            case DISABLED:
                circuitBreaker.transitionToDisabledState();
                break;
            case METRICS_ONLY:
                circuitBreaker.transitionToMetricsOnlyState();
                break;
        }
    }
```

PR #2266, opened 2025-01-11
Added support for disabling circuit breaker from config

agarwalbharat:
Fixes: https://github.com/resilience4j/resilience4j/issues/1699

Adds support to start and keep circuit breaker in disabled state through config and can be controlled in environments

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -83,6 +83,7 @@ public class CircuitBreakerConfig implements Serializable {
     private int minimumNumberOfCalls = DEFAULT_MINIMUM_NUMBER_OF_CALLS;
     private boolean writableStackTraceEnabled = DEFAULT_WRITABLE_STACK_TRACE_ENABLED;
     private boolean automaticTransitionFromOpenToHalfOpenEnabled = false;
+    private boolean disabled = false;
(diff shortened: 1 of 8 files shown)


kumsma:
@RobWin could you please look into this, this is nice addition to disable through config.

RobWin:
Hello,
would you like to explain or describe your PR?
It's always helpful to read your concept and ideas before reviewing the code.

agarwalbharat:
@RobWin 
Currently to disable Circuit breaker we need to handle differently, so basic idea was to control enable / disable of circuit breaker from config, or while through builder, it will help to control in different environment or fleet or shard differently as per need.

So about PR, it's reading disabled flag while either creating circuit breaker instance or through builder or from application properties or yaml file, and while initializing CB state reference in CircuitBreakerStateMachine instead of initialization with closed state, checking if disabled, initializing state reference to disabled state, else initializing with closed state.

I hope this helps, 
Thank you :)

RobWin:
Thanks.
Do you think the disabled flag is sufficient or would it be more flexible if you could configure the initial state, e.g. you could even start circuitbreaker in METRICS state or in OPEN state?

agarwalbharat:
> Thanks. Do you think the disabled flag is sufficient or would it be more flexible if you could configure the initial state, e.g. you could even start circuitbreaker in METRICS state or in OPEN state?

Umm.. Yes, I agree, I initially thought of doing same way, where give open to start CB in particular state, whereas as per implementation and logic OPEN and other few states needs reference to existing state for example open state, it take current state attempt and current state metrics. 

But I hear your point, I think I should be able to initialize with default values based on last state to be CLOSED. Will raise a new PR with that change, Thanks for your suggestion.

PR #2268, opened 2025-01-14
Added support of initializing circuitBreaker in desired state from config

agarwalbharat:
Adding support to initializing circuit breaker in desired state, like initializing it in METRICS_ONLY, DISABLED etc. 

Basic idea was to give flexibility to control initial state of circuit breaker, and can be managed through configs instead of injected CircuitBreakerStateMachine and manually transiting to state. 

Details about implementation:

- Reading initialState from config file, or while building instance of Circuit breaker through builder. 
- In CircuitBreakerStateMachine for stateReference instead of having hardcoded CLOSED state, initializing stateReference with the given config based state. 
- While creating instance of State, considering initial attempt to be 1 and for OPEN state, providing instance of metrics from considering current state to be CLOSED with initial metric values as open state instance expects current state metric reference (`CircuitBreakerMetrics.forClosed(getCircuitBreakerConfig())`).


agarwalbharat:
@RobWin could you please help in reviewing, as we discussed in last thread regarding making it more flexible https://github.com/resilience4j/resilience4j/pull/2266#issuecomment-2589390415

yavor300:
In Hystrix, we had the convenience of using an `enabled: false` flag to disable the circuit breaker globally throughout the application's runtime. This approach was straightforward and explicitly communicated the intention to disable the circuit breaker.

The introduction of an `initialState` flag feels less intuitive for achieving the same goal. While it offers flexibility, it may not provide the same level of clarity for users aiming to completely disable the circuit breaker.

As a user of the library, I found the previous implementation ([#2666](https://github.com/resilience4j/resilience4j/pull/2266)) more user-friendly and direct in addressing this use case (#1699). It would be great to revisit this aspect to ensure the solution remains intuitive and easily configurable for developers. I am also willing to contribute if help is needed.

RobWin:
Hey, thank you for your feedback.
Its important to get feedback from users.
So you would prefer to just have a disabled flag?

yavor300:
Yes, having the option to set `enabled: false` for a circuit breaker via an environment configuration flag is much more convenient for me. It clearly states that the circuit breaker is disabled.

agarwalbharat:
Do you think it make more sense to keep both options on the table? 
If someone wants just a disable they can have flag where simply disable CB, but in case someone wants a more control, wants to have metric only or other states they could prefer this way to setting initial state?

yavor300:
Yes, it would be nice. Implement it the way you want, I just share some thoughts.

RobWin:
I believe having two separate configuration options that can conflict with each other may lead to confusion and unintended behavior. For instance, setting disabled=true while also having initialState=OPEN, or disabled=false with initialState=DISABLED, creates scenarios where the behavior could be ambiguous or incorrect.

To avoid this, we should simplify it by using only one configuration option instead of both. What is your vote?

yavor300:
In favor of flexibility and potential future development, we can stick to the `initialState` suggestion and provide documentation on how to disable the circuit using that property. While it may be less clear, the documentation will address this.

agarwalbharat:
> Yes, then let's follow that road and stick with initialState .

Could you please review the PR

dhuome:
Hi @RobWin and @agarwalbharat I hope you’re both doing well! Just checking in on this PR—do you have an estimated timeline for when it might be completed? I’m really looking forward to this feature and can’t wait to use it! 🚀🔥 Thanks!

dhuome:
@agarwalbharat Also, just to confirm—this disabled state applies to each instance individually, right?

agarwalbharat:
> do you have an estimated timeline for when it might be completed?
I have completed from my side waiting on @RobWin to review and merge 

> this disabled state applies to each instance individually, right?
Yes @dhuome each instance can have desired inital state configured individually

RobWin:
Seem a test is failing

```
io.github.resilience4j.circuitbreaker.internal.CircuitBreakerStateMachineTest > circuitBreakerInitializeInOpenState FAILED
    org.junit.ComparisonFailure at CircuitBreakerStateMachineTest.java:952
    ```

agarwalbharat:
It was working since I last checked.
@RobWin Can you rerun the checks? I have added transition method to move it to half open in test cases.

RobWin (approved):

PR #2301, opened 2025-04-03
Remove dependency on kotlin-stdlib-jdk8 from resilience4j-core

### LINKED ISSUE #2300: Why is kotlin-stdlib-jdk8 a dependency of resilience4j-core
I was looking through our dependency tree and noticed that resilience4j-core is depending on kotlin-stdlib-jdk8. Is this necessary? Can it be excluded?

Many thanks!

gavlyukovskiy:
Taking into account the comment https://github.com/resilience4j/resilience4j/pull/2241#discussion_r1860550374, moving these tests to resilience4j-kotlin would be a better compromise, rather than bringing whole Kotlin runtime together with a lightweight core module.

The problem is that the kotlin plugin automatically adds stdlib to dependencies, which is now added to all projects using resilience4j (https://mvnrepository.com/artifact/io.github.resilience4j/resilience4j-core/2.3.0), even though it never compiles Kotlin code nor needs it.

Alternatively, I have tried to
1. Searched YouTrack but did not find anything to exclude kotlin stdlib
2.  Tried to manually remove kotlin dependencies from the relevant configurations - did not work
    ```groovy
    configurations.implementation.dependencies.removeAll {
        it.group == "org.jetbrains.kotlin"
    }
    ```
3. Adding explicit exclusion:
    ```groovy
    configurations.all {
        exclude group: 'org.jetbrains.kotlin', module: 'kotlin-stdlib-jdk8'
    }
    ```
    but it just results in weird publication POM:
    ```xml
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib-jdk8</artifactId>
      <version>1.9.0</version>
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <artifactId>kotlin-stdlib-jdk8</artifactId>
          <groupId>org.jetbrains.kotlin</groupId>
        </exclusion>
      </exclusions>
    </dependency>
    ```

Fixes #2300


sigpwned:
I am just a random user, but I would also love to see this merged. @gavlyukovskiy, appreciate the work you did to get the PR together!

PR #1462, opened 2021-06-16
Issue #1452: base config's exceptionPredicate should be null, when exceptions are instance-specific.

melochin:
fix #1452

PR #2302, opened 2025-04-03
Bring back 'slidingWindow' with default synchronization strategy

### LINKED ISSUE #2272: Lock free sliding window implementation broke API going from 2.2.0 -> 2.3.0
The implementation for #2241 broke the API, and didn't provide a  default `SlidingWindowSynchronizationStrategy` in the API (nor a deprecation of the old method for constructing a circuitBreaker with slidingWindow:

```
		CircuitBreakerConfig circuitBreakerConfig = CircuitBreakerConfig.custom()
			.failureRateThreshold(50)
			.waitDurationInOpenState(Duration.ofSeconds(60))
			.permittedNumberOfCallsInHalfOpenState(10)
			.slidingWindow(100, 100, SlidingWindowType.COUNT_BASED)
			.build();

		circuitBreakerRegistry = CircuitBreakerRegistry.of(circuitBreakerConfig);
```
broke because `.slidingWindow()` now requires the synchronization strategy.

A better way is to keep the original method, and let that perform a default strategy and introduce the new method with the strategy, perhaps deprecating the old method suggesting users of the API to pick a synchronization strategy, and then in the next major release removing the deprecated API.

Also the JavaDoc didn't suggest which default was the original strategy (if it had any).

gavlyukovskiy:
Fixes #2272

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java
@@ -672,6 +672,54 @@ public class CircuitBreakerConfig implements Serializable {
         }
 
         /**
+         * Configures the sliding window which is used to record the outcome of calls when the
+         * CircuitBreaker is closed. {@code slidingWindowSize} configures the size of the sliding
+         * window. Sliding window can either be count-based or time-based, specified by {@code
+         * slidingWindowType}. {@code minimumNumberOfCalls} configures the minimum number of calls
+         * which are required (per sliding window period) before the CircuitBreaker can calculate
+         * the error rate. For example, if {@code minimumNumberOfCalls} is 10, then at least 10
+         * calls must be recorded, before the failure rate can be calculated. If only 9 calls have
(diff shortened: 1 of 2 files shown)


gavlyukovskiy:
It looks like all builds are broken due to old version of `actions/cache`, so not related to my changes. This should be fixed if another PR is merged first: https://github.com/resilience4j/resilience4j/pull/2252

Is there anyone who can review this? @RobWin perhaps you could take a look?

gavlyukovskiy:
@RobWin thank you for updating the actions! I have updated this MR (and also https://github.com/resilience4j/resilience4j/pull/2301) and now the build passes 😃 

P.S. I've noticed that `ThreadPoolBulkheadTest` is a bit flaky ([1](https://github.com/resilience4j/resilience4j/actions/runs/14794326757/job/41537981463), [2](https://github.com/resilience4j/resilience4j/actions/runs/14795175346/job/41540564468), also [this comment](https://github.com/resilience4j/resilience4j/pull/1462#issuecomment-862822590) mentioned the same). From the first glance it looks like one thread is trying to occupy the bulkhead, so that one of the other two gets an exception, but due to a race condition, if `second` or `third` thread manage to get into bulkhead before the `fist` (sleeping), then all three can succeed. Let me know if you want me to open another MR with a fix.

PR #2329, opened 2025-06-26
Provide key for components using SpEL for names

doumdoum:
When using SpEL for resilience4j components, default configuration is taken.
Say that i have:
`@RateLimiter(name="#param")
int foo(String param) { /* do something */ }
`
Calling it with `foo("bar")` will look for `bar` configuration whereas `foo("baz")` will look for `baz`configuration
which are very unlikely to exist, therefore the default ratelimiter configuration will be used.

My proposal is to add a new param in the resilience4j annotations to ensure that a predefined configuration is used.
`@RateLimiter(name="#param", configurationKey="foo")` to use `foo` rate limiter configuration if exists.

See code attached to this PR.

Let me know if that suits you before i move forwards.
Thx for your help.

### DIFF
--- a/resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/bulkhead/configure/BulkheadAspect.java
+++ b/resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/bulkhead/configure/BulkheadAspect.java
@@ -151,7 +151,7 @@ public class BulkheadAspect implements Ordered {
     private io.github.resilience4j.bulkhead.Bulkhead getOrCreateBulkhead(String methodName,
         String backend, String configKey) {
         BulkheadConfig config = bulkheadRegistry.getConfiguration(configKey)
-            .orElse(bulkheadRegistry.getDefaultConfig());
+            .orElseGet(bulkheadRegistry::getDefaultConfig);
         var bulkhead = bulkheadRegistry.bulkhead(backend, config);
 
         if (logger.isDebugEnabled()) {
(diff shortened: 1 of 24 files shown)


RobWin:
Hello,
yes I think the idea/concept makes sense. 
Would you adjust all Annotations?

doumdoum:
@RobWin many thx for your message.
I hope to submit a PR this week

doumdoum:
@RobWin There it is!

resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/bulkhead/configure/BulkheadAspect.java
```diff
@@ -116,8 +116,8 @@ public Object bulkheadAroundAdvice(ProceedingJoinPoint proceedingJoinPoint,
                 () -> proceedInThreadPoolBulkhead(proceedingJoinPoint, methodName, returnType, backend);
             return fallbackExecutor.execute(proceedingJoinPoint, method, bulkheadAnnotation.fallbackMethod(), bulkheadExecution);
         } else {
-            io.github.resilience4j.bulkhead.Bulkhead bulkhead = getOrCreateBulkhead(methodName,
-                backend);
+            String configKey = bulkheadAnnotation.configuration().isEmpty() ? backend : bulkheadAnnotation.configuration();
```
RobWin: I think when `bulkheadAnnotation.type() == Bulkhead.Type.THREADPOOL` the `backend` is not overridden when the config key is set. I think this is missing.

doumdoum: In fact threadpool bulkhead was always using the default config.
I've completed implementation.
Many thx for your review.

RobWin (approved):

Issue #2224, opened 2024-10-19
Support for Virtual Threads in Resilience4j


ykhfree:
Resilience4j version: v2.1.0

Java version: jdk21

**1. Description**

Hello Resilience4j Team,

I would like to propose an enhancement to support Virtual Threads in Resilience4j. With the increasing adoption of Virtual Threads in Java, it would be beneficial for Resilience4j to support this feature to improve performance and scalability in applications that rely on heavy blocking operations.

**2. Motivation**

Virtual Threads offer a lightweight concurrency model that can handle a large number of concurrent tasks efficiently. By supporting Virtual Threads, Resilience4j can provide better performance for applications that perform blocking operations, such as I/O tasks, without the overhead of traditional threads.

**3. Planned Approach for Virtual Threads**

I plan to implement the following changes:

- Utilize Virtual Threads specifically for tasks that involve heavy blocking operations to maximize their efficiency.
- Carefully avoid using Thread methods that are incompatible with Virtual Threads to prevent any potential issues.
- Ensure that tasks run by Virtual Threads do not include synchronized code, as this can lead to performance bottlenecks and pinned issues, which negate the benefits of using Virtual Threads.

**4. Benefits**

Improved performance and scalability for applications using Resilience4j with Virtual Threads.
Enhanced compatibility with modern Java concurrency features.

**5. Additional Information**

I am willing to contribute to this enhancement and implement these changes. Please let me know if there are any specific guidelines or processes I should follow to get started.

Thank you for considering this proposal. I look forward to your feedback and the possibility of contributing to Resilience4j.

RobWin:
Hello,

thank you for your well structured ideas.
Enhancements are very welcome. There is no specific guideline or process to follow, except that you can create a MR and explain you thoughts and changes.
Just keep in mind that we try to keep the API backwards-compatible. If this is not possible we would need to work on a Resilince4j v3 which supports Virtual Threads.

ykhfree:
Thank you for your response.

I think existing API compatibility is one of the most important things to work on when it comes to applying virtual threads, I'll keep that in mind.
Also, since you mentioned v3, as you know, **virtual threads have been officially supported since JDK 21**.
[...]

PR #2335, opened 2025-07-16
Feature #2224 : Apply virtual thread

ykhfree:
issue #2224

# Add Virtual Thread Support to Resilience4j

## Summary

This PR introduces comprehensive support for Java 21's Virtual Threads (Project Loom) across all Resilience4j modules, enabling high-performance concurrent operations with significantly reduced resource overhead compared to traditional platform threads.

### Core Infrastructure
- **New `ExecutorServiceFactory`**: Central factory for thread management with automatic virtual/platform thread detection
- **Configuration Property**: `resilience4j.thread.type` system property to control thread type (`virtual` or `platform`)
- **Backward Compatibility**: Seamless fallback to platform threads when virtual threads are unavailable

### Module Updates
- **Bulkhead**: `FixedThreadPoolBulkhead` now supports virtual thread pools with proper naming conventions
- **CircuitBreaker**: Enhanced `SchedulerFactory` with thread-safe virtual thread support and automatic configuration detection
- **RateLimiter**: Updated `SemaphoreBasedRateLimiter` implementation for virtual thread compatibility
- **Retry**: Virtual thread support in retry mechanisms
- **TimeLimiter**: Enhanced timeout handling with virtual threads

### ConcurrentHashMap → Smart CompletableFuture Usage
Applied selective CompletableFuture optimization based on operation complexity to prevent virtual thread pinning while maintaining performance:

#### ✅ **CompletableFuture Applied (Complex Operations)**
- **`InMemoryRegistryStore.entryMap`**: Uses optimized `putIfAbsent` pattern with complex object creation via `mappingFunction.apply()`
    - **Virtual Thread Benefits**: When virtual threads encounter blocking operations (I/O, database calls, complex initialization), they can **park without blocking carrier threads**, enabling much higher concurrency compared to synchronized blocks that cause thread pinning

**Key Optimizations Applied:**
- **putIfAbsent Pattern**: Replaced `computeIfAbsent` with more efficient `putIfAbsent` for better virtual thread performance
- **Exception Handling**: Only cleanup failed futures when `completeExceptionally()` returns `true` to prevent race conditions
- **JDK 25 Preparation**: Added TODO comments indicating future simplification when JDK 25 optimizes `computeIfAbsent` internally

**Result**: Optimized memory usage and CPU performance while maintaining virtual thread benefits where they provide genuine value.

**Performance Test Results:** (with I/O 500-1500ms delays)
    - Added a hash collision situation to the scenario
| Implementation | Thread Type | Throughput | Max Concurrent | Virtual Thread Advantage | Status |
|---|---|---|---|---|---|
| **Future-based HashMap** | **Virtual Thread** | **101.2 ops/s** | **100** | **+370.7%** 🥇 | ✅ **Selected** |
| ConcurrentSkipListMap | Virtual Thread | 89.5 ops/s | 100 | +194.4% 🥈 |  |
| ConcurrentHashMap | Virtual Thread | 25.9 ops/s | 85* | +63.9% | ❌ Thread Pinning |
| Future-based HashMap | Platform Thread | 21.5 ops/s | 20** | baseline | - |
| ConcurrentSkipListMap | Platform Thread | 30.4 ops/s | 20** | baseline | - |
| ConcurrentHashMap | Platform Thread | 15.8 ops/s | 20** | baseline | - |

*Limited by thread pinning issues  
**Platform threads limited to 20 due to high memory overhead (~2MB per thread vs ~1KB for virtual threads)

### Testing Strategy
- **Parameterized Tests**: Existing tests now run on both platform and virtual threads via `ThreadModeTestBase`
- **Dedicated Virtual Thread Tests**: New test suites specifically for virtual thread scenarios
- **Performance Validation**: Thread metrics and behavior verification across both thread types
- **Compatibility Tests**: Ensuring consistent behavior between platform and virtual thread modes

### Spring Boot Integration
- **Auto-Configuration**: New `Resilience4jThreadAutoConfiguration` for seamless Spring Boot integration
- **Thread Metrics**: `ThreadMetricsAutoConfiguration` with Micrometer integration for monitoring
- **Configuration Properties**: Spring Boot properties support for thread type configuration

### Documentation
- **README Updates**: Clear instructions for virtual thread configuration and usage
- **Spring Boot Guide**: Enhanced documentation with virtual thread examples
- **Migration Guide**: Step-by-step instructions for adopting virtual threads

### Performance Benefits
- **Reduced Memory Footprint**: Virtual threads use significantly less memory than platform threads
- **Higher Concurrency**: Support for millions of concurrent operations
- **Better Resource Utilization**: Efficient handling of I/O-bound operations

#### System Property
```bash
-Dresilience4j.thread.type=virtual
```

#### Spring Boot Configuration
```yaml
resilience4j:
  thread:
    type: virtual
```

#### Environment Variable
```bash
export resilience4j.thread.type=virtual
```

### Thread Naming Conventions
- **Platform Threads**: `bulkhead-{name}-{number}` (e.g., `bulkhead-test-1`)
- **Virtual Threads**: `bulkhead-{name}-v-{number}` (e.g., `bulkhead-test-v-0`)

### ExecutorServiceFactory Design
- **Singleton Pattern**: Thread-safe factory with lazy initialization
- **Configuration Detection**: Real-time detection of thread type changes
- **Resource Management**: Proper shutdown and cleanup of thread pools
- **Fallback Strategy**: Graceful degradation to platform threads

### Thread Pool Behavior
- **Bulkhead**: Virtual thread pools maintain the same semantics as platform thread pools
- **CircuitBreaker**: Scheduler factory automatically adapts to thread type changes
- **Context Propagation**: Enhanced `ContextPropagator` for virtual thread compatibility

### Testing Architecture
- **ThreadModeTestBase**: Base class for parameterized testing across thread types
- **Test Coverage**: >95% code coverage maintained across both thread modes
- **Performance Benchmarks**: Validation of virtual thread performance benefits

### Requirements
- **Java 21+**: Virtual threads require Java 21 or later
- **Backward Compatibility**: Full compatibility with existing Resilience4j applications
- **Automatic Fallback**: Graceful degradation on pre-Java 21 environments

### Migration Path
1. **No Code Changes Required**: Existing applications work unchanged
2. **Opt-in Activation**: Virtual threads must be explicitly enabled
3. **Gradual Adoption**: Can be enabled per module or globally

### Enhanced Existing Tests
- Parameterized tests running on both platform and virtual threads
- Thread naming validation for both thread types
- Behavior consistency verification across thread modes

## Breaking Changes

**None** - This is a fully backward-compatible addition. Existing applications will continue to work unchanged.

## Future Considerations

- **Structured Concurrency**: Potential integration with Java's structured concurrency features
- **Performance Monitoring**: Enhanced metrics for virtual thread pools
- **Framework Integration**: Deeper integration with reactive frameworks


ykhfree:
@RobWin 
Please review this PR

resilience4j-core/src/test/java/io/github/resilience4j/core/ThreadModeTestBaseTest.java
```diff
@@ -0,0 +1,159 @@
...
+     * Concrete implementation of ThreadModeTestBase for testing purposes.
+     * Exposes protected methods for verification.
+     */
+    private static class TestableThreadModeTestBase extends ThreadModeTestBase {
+        
+        public TestableThreadModeTestBase(ThreadType threadType) {
+            super(threadType);
+        }
+        
+        public void setUpThreadMode() {
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/ThreadModeTestBaseTest.java
```diff
@@ -0,0 +1,159 @@
...
+            super.setUpThreadMode();
+        }
+        
+        public void cleanUpThreadMode() {
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/ThreadModeTestBaseTest.java
```diff
@@ -0,0 +1,159 @@
...
+            super.cleanUpThreadMode();
+        }
+        
+        public boolean isVirtualThreadMode() {
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/ThreadModeTestBaseTest.java
```diff
@@ -0,0 +1,159 @@
...
+            return super.isVirtualThreadMode();
+        }
+        
+        public String getThreadModeDescription() {
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/ContextPropagatorTest.java
```diff
@@ -257,4 +296,186 @@ public void contextPropagationRunnableMultipleFailureTest() {
...
+    @Test
+    public void contextPropagationWithMultiplePropagators() throws Exception {
+        System.out.println("Running contextPropagationWithMultiplePropagators in " + getThreadModeDescription());
+        
+        // Create multiple ThreadLocals and propagators 
+        ThreadLocal<String> stringThreadLocal = new ThreadLocal<>();
+        ThreadLocal<Integer> intThreadLocal = new ThreadLocal<>();
+        
+        String stringValue = "string-value-" + threadType;
+        Integer intValue = 42 + threadType.hashCode(); // Different value per mode
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-hedge/src/test/java/io/github/resilience4j/hedge/HedgeConcurrencyTest.java
```diff
@@ -0,0 +1,367 @@
...
+
+        for (int i = 0; i < 5; i++) {
+            final int threadId = i;
+            testExecutor.submit(() -> {
+                try {
+                    startLatch.await();
+
+                    // Slow execution to potentially trigger hedge
+                    Callable<String> task = () -> {
+                        Thread.sleep(50 + (threadId * 10)); // Longer than hedge delay
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/VirtualThreadTestWithHashCollisionsAndRealIO.java
```diff
@@ -0,0 +1,964 @@
...
+                    
+                    // Simulate some processing time for read operations (lighter than write)
+                    if (result != null) {
+                        performLightProcessing(readKey, testName);
+                    }
+                    
+                } else {
+                    // WRITE OPERATION: Create new entry with blocking I/O
+                    String writeKey;
+                    if (op < COLLISION_KEYS.length && threadId < COLLISION_KEYS.length) {
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/metrics/CASLoopBenchmarkTest.java
```diff
@@ -0,0 +1,244 @@
...
+        
+        // Force GC
+        System.gc();
+        Thread.sleep(100);
+        
+        // Actual benchmark
+        return runBenchmark(metrics, threadCount, useVirtualThreads, BENCHMARK_ITERATIONS);
+    }
+
+    private BenchmarkResult runBenchmark(Metrics metrics, int threadCount,
```
gavlyukovskiy: These results are unlikely to be accurate, it would be better to use [JMH](https://github.com/openjdk/jmh) for writing benchmarks. You can follow an example here: https://github.com/resilience4j/resilience4j/tree/master/resilience4j-core/src/jmh

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
```diff
@@ -1,28 +1,38 @@
...
+    public RateLimitersImplementationTest(ThreadType threadType) {
+        super(threadType);
+    }
 
     protected abstract RateLimiter buildRateLimiter(RateLimiterConfig config);
 
     @Test
     public void acquireBigNumberOfPermitsAtStartOfCycleTest() {
+        System.out.println("Running acquireBigNumberOfPermitsAtStartOfCycleTest in " + getThreadModeDescription());
+
```
gavlyukovskiy: These comments are a bit unnecessary - IDE and Gradle usually give a fine indication between tests. Gradle can be configured to print before / after test messages if needed.

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
```diff
@@ -104,30 +127,61 @@ public void drainCycleWhichAlreadyHashNoPremitsLeftTest() {
...
-            } catch (InterruptedException ex) {
-                throw new IllegalStateException(ex);
-            }
+        System.out.print("Waiting for refresh: ");
+        try {
+            await()
+                .pollInterval(25, MILLISECONDS)
+                .atMost(config.getLimitRefreshPeriod().multipliedBy(3).toMillis(), MILLISECONDS)
+                .until(() -> {
+                    System.out.print(printedWhileWaiting);
```
gavlyukovskiy: Looks like leftover from debugging?

resilience4j-ratelimiter/build.gradle
```diff
@@ -1,4 +1,5 @@
 dependencies {
     api project(':resilience4j-core')
+    testImplementation project(':resilience4j-core').sourceSets.test.output
```
gavlyukovskiy: For things like this it's better to use [testFixtures](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures)

resilience4j-core/src/test/java/io/github/resilience4j/core/ThreadModeTestBaseTest.java
```diff
@@ -0,0 +1,163 @@
...
+     * Concrete implementation of ThreadModeTestBase for testing purposes.
+     * Exposes protected methods for verification.
+     */
+    private static class TestableThreadModeTestBase extends ThreadModeTestBase {
+        
+        public TestableThreadModeTestBase(ThreadType threadType) {
+            super(threadType);
+        }
+        
+        @Override
```
gavlyukovskiy: Why not make original methods public? It's test code anyway

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/thread/autoconfigure/Resilience4jThreadAutoConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+ * <p>Supported values: {@code virtual} or {@code platform} (default).</p>
+ *
+ * <p>This configuration is activated iff the property {@code resilience4j.thread.type} is present.
+ * If the system property is already set (e.g. via JVM arg) it is left untouched so that
+ * command-line options win over Spring configuration.</p>
+ * 
+ * @author kanghyun.yang
+ * @since 3.0.0
+ */
+@AutoConfiguration
```
gavlyukovskiy: Better to add explicit order so that this auto-configuration is executed before any other resilience4j auto-configuration. Otherwise there's no guarantee that the property will be set before registries are created.

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/thread/autoconfigure/Resilience4jThreadAutoConfiguration.java
```diff
@@ -0,0 +1,32 @@
+package io.github.resilience4j.springboot3.thread.autoconfigure;
```
gavlyukovskiy, line 1: Can you please also update `spring-boot4` module introduced in https://github.com/resilience4j/resilience4j/pull/2384?

resilience4j-core/src/main/java/io/github/resilience4j/core/registry/InMemoryRegistryStore.java
```diff
@@ -17,50 +17,120 @@
...
     }
 
     @Override
-    public E putIfAbsent(String key, E value) { return entryMap.putIfAbsent(key, value); }
+    public E putIfAbsent(String key, E value) {
+        Objects.requireNonNull(key, "Key cannot be null");
+        Objects.requireNonNull(value, "Value cannot be null");
+        
+        CompletableFuture<E> future = entryMap.putIfAbsent(key, CompletableFuture.completedFuture(value));
+        if (isSuccessfullyCompleted(future)) {
```
gavlyukovskiy: Shouldn't it join if `future != null`, but not completed? Otherwise it would change the contract for the caller - currently it would wait on a lock until other thread computes the value and would return that computed value back.

gavlyukovskiy: Reproducer code:
```java
    public static void main(String[] args) throws InterruptedException {
        var map = new ConcurrentHashMap<String, String>();

        var cdl = new CountDownLatch(1);
        CompletableFuture.runAsync(() -> {
            map.computeIfAbsent("k1", k -> {
                System.out.println(System.currentTimeMillis() + ": start computing k1");
                cdl.countDown();
                try {
                    Thread.sleep(1000);
                } catch (InterruptedException e) {
                    throw new RuntimeException(e);
                }
                System.out.println(System.currentTimeMillis() + ": done computing k1");
                return "v1";
            });
        });

        cdl.await();
        System.out.println(System.currentTimeMillis() + ": putIfAbsent k1");
        var value = map.putIfAbsent("k1", "v2");
        System.out.println(System.currentTimeMillis() + ": putIfAbsent return " + value);
        System.out.println(System.currentTimeMillis() + ": map value " + map.get("k1"));
    }
```
output:
```
1770070098963: start computing k1
1770070098966: putIfAbsent k1
1770070099972: done computing k1
1770070099972: putIfAbsent return v1
1770070099976: map value v1
```

gavlyukovskiy: The same would apply for `remove` and `replace` - previous implementation would wait and return the computed value. But, unlike for `putIfAbsent`,  I think it's very unlikely that it matters whether previous value is returned if it has not been computed yet.

resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
```diff
@@ -20,75 +20,58 @@
...
     public void onEvent(@Nullable EventConsumer<T> onEventConsumer) {
-        lock.lock();
-
-        try {
-            this.onEventConsumers.add(onEventConsumer);
-            this.consumerRegistered = true;
-        } finally {
-            lock.unlock();
-        }
+        if (onEventConsumer == null) return;
```
gavlyukovskiy: Despite what annotation says, it doesn't seem like it's supposed to accept nulls - it would fail on any event since it never checks whether consumer is null before calling `consumeEvent`. I would rather change nullable annotation

resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
```diff
@@ -20,75 +20,58 @@
...
     }
 
     public <E extends T> boolean processEvent(E event) {
         boolean consumed = false;
+
         if (!onEventConsumers.isEmpty()) {
-            for (EventConsumer<T> onEventConsumer : onEventConsumers) {
-                onEventConsumer.consumeEvent(event);
-            }
+            for (EventConsumer<T> c : onEventConsumers) c.consumeEvent(event);
```
gavlyukovskiy, line 65: Using if/for without braces is a bit confusing, it would be better to leave it as is.

gavlyukovskiy (commented):
Very good work! I've left a couple of comments.

Also one thing I wanted to ask - have you explored using multi-release JAR instead of increasing the required JDK version to 21? Even though it's been 2.5 years since release, it feels a bit early to do that already. Spring 7 / Spring Boot 4, even though released late last year, still use JDK 17 as a baseline.

resilience4j-core/src/main/java/io/github/resilience4j/core/registry/InMemoryRegistryStore.java
ykhfree: You're right - the current implementation changes the contract. When another thread is computing a value via computeIfAbsent, putIfAbsent should wait for that computation to complete and return the computed value, just like ConcurrentHashMap does.

The current isSuccessfullyCompleted() check skips futures where isDone()=false, causing immediate null return instead of waiting. I'll fix this by calling future.join() when future is not null.

So I will fix the issue you pointed out.

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/thread/autoconfigure/Resilience4jThreadAutoConfiguration.java
ykhfree: Good point! Without explicit ordering, there's a race condition where registries might be created before the thread type property is set, causing virtual threads to not be enabled as expected.

I'll use `@AutoConfiguration(before = {...})` to explicitly declare ordering against other resilience4j auto-configurations.

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/thread/autoconfigure/Resilience4jThreadAutoConfiguration.java
ykhfree: Absolutely! I'll add the same thread auto-configuration to the spring-boot4 module. I'll check if PR #2384 has been merged first - if so, I'll include this in the current PR. If not, I can either wait for it to be merged or create a separate follow-up PR to avoid conflicts.

The implementation will mirror the spring-boot3 module structure with appropriate package naming for spring-boot4.

resilience4j-core/src/test/java/io/github/resilience4j/core/ThreadModeTestBaseTest.java
ykhfree: You're right - since `ThreadModeTestBase` is test infrastructure code, there's no need for the protected visibility on the utility methods. I'll fix it.

resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
ykhfree: You're correct. The `@Nullable` annotation is misleading - while the method currently handles null by early return (silent no-op), the API contract is unclear. I'll remove `@Nullable` and add `Objects.requireNonNull()` to make the contract explicit.

Since this is a major version (Resilience4j 3.0), this breaking change is acceptable. Previously onEvent(null) was silently ignored; after this change it will throw `NullPointerException`.

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
ykhfree: Yes, these are debugging leftovers from development. I'll remove all `System.out.println`/`System.out.print` statements from the test methods and the `waitForRefresh` helper.

resilience4j-core/src/main/java/io/github/resilience4j/core/EventProcessor.java
ykhfree: Unless you have a strong opinion, I'll leave the original formatting as is. Thanks for the clarification.

resilience4j-core/src/test/java/io/github/resilience4j/core/metrics/CASLoopBenchmarkTest.java
ykhfree: Thank you for the suggestion and the reference! You're right that JUnit-based benchmarks don't provide accurate results due to JVM warmup, JIT compilation, and other factors that JMH handles properly.

I'd like to address this in a separate follow-up PR to keep this one focused. I'll migrate the benchmark to the existing JMH infrastructure under `src/jmh/java/` and include proper warmup iterations, measurement configurations, and more meaningful metrics. 

Would that be acceptable?

resilience4j-ratelimiter/build.gradle
ykhfree: Agreed - using `sourceSets.test.output` is indeed an anti-pattern in Gradle. The `testFixtures` plugin provides a cleaner way to share test utilities across modules.

Since this change affects multiple modules and requires moving files to a new source set, I'd prefer to handle this in a separate PR to minimize the scope of this one. I'll create a follow-up PR that:

What do you think?

ykhfree:
> Very good work! I've left a couple of comments.
> 
> Also one thing I wanted to ask - have you explored using multi-release JAR instead of increasing the required JDK version to 21? Even though it's been 2.5 years since release, it feels a bit early to do that already. Spring 7 / Spring Boot 4, even though released late last year, still use JDK 17 as a baseline.

Hmmm...that's a troubling point.

I think multi-release JARs have their advantages, but they also have their disadvantages.
I consider a major update to 3.0.0 to be a breaking change that can also change the JDK support version, so I think a single release jar is neat.

What's the scenario you're thinking of, where the jars for JDK17 can't use virtual threads and force the use of platform threads?

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
ykhfree: Yep, that's right. Let's remove it.

resilience4j-ratelimiter/build.gradle
gavlyukovskiy: Fair enough 👍

resilience4j-spring-boot3/src/main/java/io/github/resilience4j/springboot3/thread/autoconfigure/Resilience4jThreadAutoConfiguration.java
gavlyukovskiy: The PR has been merged quite a while ago, but after this one was opened

ykhfree:
@RobWin 
@gavlyukovskiy 

OK, I create PR for back to jdk21
https://github.com/resilience4j/resilience4j/pull/2446

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachineThreadModeTest.java
```diff
@@ -0,0 +1,155 @@
...
+    public void setUpThreadMode() {
+        super.setUpThreadMode();
+        SchedulerFactory.getInstance().reset();
+    }
+    
+    @Override
+    public void cleanUpThreadMode() {
+        super.cleanUpThreadMode();
+        SchedulerFactory.getInstance().reset();
+    }
```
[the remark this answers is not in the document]
ykhfree: Copilot claims that overriding `setUpThreadMode()` without re-adding `@Before` causes the base class setup to never run. This is **incorrect for JUnit 4**: the framework discovers `@Before` on the parent's method declaration, and at runtime, Java polymorphism invokes the child's overridden version. The overridden method **does execute correctly**, including the additional `SchedulerFactory.getInstance().reset()` call.

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/internal/SchedulerFactoryConcurrencyTest.java
```diff
@@ -0,0 +1,231 @@
...
+    public void setUpThreadMode() {
+        super.setUpThreadMode();
+        SchedulerFactory.getInstance().reset();
+    }
+    
+    @Override
+    public void cleanUpThreadMode() {
+        super.cleanUpThreadMode();
+        SchedulerFactory.getInstance().reset();
+    }
```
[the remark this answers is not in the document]
ykhfree: Copilot claims that overriding `setUpThreadMode()` without re-adding `@Before` causes the base class setup to never run. This is **incorrect for JUnit 4**: the framework discovers `@Before` on the parent's method declaration, and at runtime, Java polymorphism invokes the child's overridden version. The overridden method **does execute correctly**, including the additional `SchedulerFactory.getInstance().reset()` call.

resilience4j-core/src/test/java/io/github/resilience4j/core/metrics/CASLoopBenchmarkTest.java
ykhfree: I'll modify my plan to introduce JMH to this PR.

ykhfree:
@RobWin
@gavlyukovskiy

In addition to the jdk21 rollback [PR](https://github.com/resilience4j/resilience4j/pull/2446) I already shared, this PR (applying virtual threads) is also ready.

resilience4j-bulkhead/src/test/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadTest.java
```diff
@@ -38,13 +45,25 @@
...
 
-public class ThreadPoolBulkheadTest {
+@RunWith(Parameterized.class)
+public class ThreadPoolBulkheadTest extends ThreadModeTestBase {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return threadModes();
+    }
+
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-bulkhead/src/test/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadTest.java
```diff
@@ -54,8 +73,15 @@ public void setUp() {
             .build();
     }
 
+    @After
+    public void tearDown() {
+        cleanUpThreadMode(); // Clean up thread mode from ThreadModeTestBase
+    }
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/internal/SchedulerFactoryTest.java
```diff
@@ -1,12 +1,40 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class SchedulerFactoryTest {
+@RunWith(Parameterized.class)
+public class SchedulerFactoryTest extends ThreadModeTestBase {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return threadModes();
+    }
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/test/java/io/github/resilience4j/core/metrics/LockFreeMetricsContentionTest.java
```diff
@@ -0,0 +1,322 @@
...
+ * Integration tests for lock-free metrics under high contention scenarios.
+ * Tests that CAS backoff strategies work correctly with both platform and virtual threads.
+ */
+@RunWith(Parameterized.class)
+public class LockFreeMetricsContentionTest extends ThreadModeTestBase {
+
+    @Parameterized.Parameters(name = "threadMode={0}")
+    public static Collection<Object[]> data() {
+        return threadModes();
+    }
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiterImplTest.java
```diff
@@ -44,9 +44,18 @@
...
 import static org.mockito.Mockito.*;
 
-
+@RunWith(Parameterized.class)
 public class SemaphoreBasedRateLimiterImplTest extends RateLimitersImplementationTest {
 
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return threadModes();
+    }
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-core/src/main/java/io/github/resilience4j/core/ExecutorServiceFactory.java
```diff
@@ -0,0 +1,144 @@
...
+     * Create a single–thread {@link ScheduledExecutorService} with the specified thread type.
+     */
+    public static ScheduledExecutorService newSingleThreadScheduledExecutor(String poolName, ThreadType threadType) {
+        if (threadType == ThreadType.VIRTUAL) {
+            /*
+             * The JDK currently does not ship a dedicated virtual-thread scheduler factory
+             * (java.util.concurrent.Executors API is still incubating).
+             * We therefore create a scheduled executor backed by a virtual-thread per task
+             * executor.  Implementation detail: use {@code Runnable::run} as the thread
+             * factory and rely on virtual threads for actual concurrency.
```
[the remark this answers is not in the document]
ykhfree: Fixed

ykhfree:
@RobWin
@gavlyukovskiy

I also addressed additional review findings from the copilot.

resilience4j-retry/src/test/java/io/github/resilience4j/retry/internal/RetryThreadModeTest.java
```diff
@@ -0,0 +1,245 @@
...
+        // Track retry attempts and thread types
+        AtomicInteger attempts = new AtomicInteger(0);
+        AtomicBoolean allCorrectThreadType = new AtomicBoolean(true);
+        
+        // Create a supplier that will fail twice then succeed, tracking thread types
+        // Use appropriate executor based on thread mode
+        ExecutorService taskExecutor = isVirtualThreadMode() ? 
+            Executors.newVirtualThreadPerTaskExecutor() : 
+            Executors.newSingleThreadExecutor();
+
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/RateLimitersImplementationTest.java
```diff
@@ -100,34 +109,57 @@ public void drainCycleWhichAlreadyHashNoPremitsLeftTest() {
...
+                .atMost(config.getLimitRefreshPeriod().multipliedBy(3).toMillis(), MILLISECONDS)
+                .until(() -> metrics.getAvailablePermissions() == config.getLimitForPeriod());
+        } catch (Exception e) {
+            throw new AssertionError("Refresh did not complete within expected time. " +
+                "Current permits: " + metrics.getAvailablePermissions() + 
+                ", Expected: " + config.getLimitForPeriod() + 
+                ". Error: " + e.getMessage(), e);
         }
-        System.out.println();
     }
```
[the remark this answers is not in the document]
ykhfree: Fixed

resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/SchedulerFactory.java
```diff
@@ -1,27 +1,104 @@
...
+                || scheduler.isShutdown()
+                || scheduler.isTerminated()) {
+
+                ScheduledExecutorService fresh =
+                    ExecutorServiceFactory.newSingleThreadScheduledExecutor(
+                        "CircuitBreakerAutoTransitionThread");
+
+                old = scheduler;
+                scheduler = fresh;
+                virtual = desiredVirtual;
```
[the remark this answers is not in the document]
ykhfree: Fixed

RobWin:
Hi, a lot of tests contain ` System.out.`
Could you remove them, then we can merge it. Thank you so much

ykhfree:
> Hi, a lot of tests contain ` System.out.` Could you remove them, then we can merge it. Thank you so much

I changed all of the `System.out`s that existed in my work to loggers.

Issue #2354, opened 2025-09-15
Get rid of Kotlin dependencies from resilience4j-core


drnow4u:
Hi team,

Would it be possible to remove the Kotlin dependencies from the resilience4j-core module?
Our company follows an internal Lifecycle Policy, and the inclusion of Kotlin dependencies introduces compliance and maintenance challenges for us.

Thank you for considering this request!

Best regards,
Marcin

internetstaff:
The EOL version of Kotlin being shipped also causes us compliance issues.

It looks like some work to stop shipping that may have been done earlier this year, but was never released?

Thanks!

drnow4u:
PR #2359 is merged. Closing issue. Thanks

PR #2359, opened 2025-10-08
[GH-2354] Remove kotlin-stdlib from resilience4j-core

josalmi:
As can be seen from https://mvnrepository.com/artifact/io.github.resilience4j/resilience4j-core/2.3.0 the resilience4j-core currently includes kotlin-stdlib as a compile dependency. This happens automatically when kotlin plugin is applied in a gradle project and if if the intention was only to allow tests to be written it kotlin there are any kotlin sources in the core package.

```
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib</artifactId>
      <version>1.9.25</version>
      <scope>compile</scope>
    </dependency>
```

### DIFF
--- a/resilience4j-core/build.gradle
+++ b/resilience4j-core/build.gradle
@@ -1,21 +1,9 @@
-import org.jetbrains.kotlin.gradle.dsl.JvmTarget
-
-plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.9.25'
-}
-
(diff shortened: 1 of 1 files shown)


josalmi:
If merged before https://github.com/resilience4j/resilience4j/pull/2352 need to make sure that we don't accidentally pull the kotlin back in with that PR

Khalidmamdou7:
We need this.

PR #2386, opened 2025-12-15
feat: add setter for bulkheadAspectOrder property

### LINKED ISSUE #1657: Confusing documentation on aspect order
Resilience4j version: v1.7.0
Java version: 17

[Official documentation](https://resilience4j.readme.io/docs/getting-started-3#aspect-order) states the following default order of aspects:
`Retry ( CircuitBreaker ( RateLimiter ( TimeLimiter ( Bulkhead ( Function ) ) ) ) )`

Where Bulkhead has the highest priority and Retry the lowest.
This is already a bit difficult to understand, because if you think that we enter the chain from the left and move to the right, then it's more logical to say that Retry is actually has the highest priority, since we enter it first. But I guess the logic is different: although we enter from the left and move to the right, the way back is from right to left and it also determines when each aspect is finished. This means that CircuitBreaker is completely finished before Retry and in case of exceptions, they're also propagated from right to left. Is my understanding correct?

But the main problem is with changing that order. For example, the default order could lead to a situation when Bulkhead exceptions trigger CircuitBreaker to flip to OPEN state which is undesired for us. According to the documentation, it should be possible to change the order by setting `resilience4j.circuitbreaker.circuitBreakerAspectOrder` and `resilience4j.bulkhead.bulkheadAspectOrder`, where `circuitBreakerAspectOrder` would have a higher value than `bulkheadAspectOrder` (the higher value = the higher priority). When I tried to do that, I got `ConfigurationPropertiesBindException` exception:

```
org.springframework.boot.context.properties.ConfigurationPropertiesBindException: Error creating bean with name
 'resilience4j.bulkhead-io.github.resilience4j.bulkhead.autoconfigure.BulkheadProperties': Could not bind properties to
 'BulkheadProperties' : prefix=resilience4j.bulkhead, ignoreInvalidFields=false, ignoreUnknownFields=true; nested exception is
 org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'resilience4j.bulkhead' to
 io.github.resilience4j.bulkhead.autoconfigure.BulkheadProperties
```

And looking at the `BulkheadConfigurationProperties` code, I see that it actually doesn't allow setting `bulkheadAspectOrder` property. Is this a bug in documentation?

I of course can ignore `BulkheadFullException` in the CircuitBreaker settings but that's beside the point. :)

NiMv1:
Fixes #1657 - Added setter for bulkheadAspectOrder to allow configuration via application.properties. Previously only getter with fixed value existed, making it impossible to configure via properties.

### DIFF
--- a/resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadConfigurationProperties.java
+++ b/resilience4j-spring/src/main/java/io/github/resilience4j/bulkhead/configure/BulkheadConfigurationProperties.java
@@ -20,12 +20,21 @@ import org.springframework.core.Ordered;
 public class BulkheadConfigurationProperties extends
     io.github.resilience4j.common.bulkhead.configuration.CommonBulkheadConfigurationProperties {
 
+    private int bulkheadAspectOrder = Ordered.LOWEST_PRECEDENCE;
+
     /**
      * As of release 0.16.0 as we set an implicit spring aspect order for bulkhead to cover the
-     * async case of threadPool bulkhead
+     * async case of threadPool bulkhead but user can override it still if he has different use case
      */
     public int getBulkheadAspectOrder() {
-        return Ordered.LOWEST_PRECEDENCE;
+        return bulkheadAspectOrder;
+    }
+
+    /**
+     * @param bulkheadAspectOrder bulkhead aspect order
+     */
+    public void setBulkheadAspectOrder(int bulkheadAspectOrder) {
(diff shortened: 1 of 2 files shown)


NiMv1:
CI for this PR shows status 'action_required' (Actions run: https://github.com/resilience4j/resilience4j/actions/runs/20240091955). It looks like the workflow needs manual approval / re-run (typical for PRs from forks / first-time contributors). Could a maintainer please approve and re-run the Build workflow?

PR #2442, opened 2026-04-01
feat: add customTagsExtractor support to TaggedRetryMetrics

### LINKED ISSUE #2228: Way to add customTag which value is calculated from RetryEvent with TaggedRetryMetrics
Resilience4j version:
'io.github.resilience4j:resilience4j-spring-boot3:2.2.0'
"io.github.resilience4j:resilience4j-reactor:2.2.0"

Java version:
17

Is there a simple way to add a custom metric to be published to the set of tagged metrics? This custom metric has its value calculated dinamicaly from retryEvent `event.getLastThrowable().getClass().getSimpleName()`

Yelagandula:
## Summary
Adds a `customTagsExtractor` function parameter to `TaggedRetryMetrics` that allows users to compute dynamic tags from `RetryEvent` at runtime.

Closes #2228

## Motivation
Currently `TaggedRetryMetrics` only supports static tags defined on the `Retry` instance itself. Users who need dynamic tags calculated from the `RetryEvent` — such as `event.getLastThrowable().getClass().getSimpleName()` — are forced to override internal publisher methods, which is brittle and unsupported.

## Changes
- Added two new factory methods to `TaggedRetryMetrics`:
  - `ofRetryRegistry(RetryRegistry, Function<RetryEvent, List<Tag>>)`
  - `ofRetryRegistry(RetryMetricNames, RetryRegistry, Function<RetryEvent, List<Tag>>)`
- Existing factory methods default to `Collections.emptyList()` — fully backwards compatible
- Added Javadoc with usage example

## Example
```java
TaggedRetryMetrics.ofRetryRegistry(
    retryRegistry,
    event -> event.getLastThrowable() != null
        ? List.of(Tag.of("exception", event.getLastThrowable().getClass().getSimpleName()))
        : Collections.emptyList()
);

### DIFF
--- a/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/AbstractRetryMetrics.java
+++ b/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/AbstractRetryMetrics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Ingyu Hwang, Mahmoud Romeh
+ * Copyright 2019 Yevhenii Voievodin, Robert Winkler
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/tagged/AbstractRetryMetrics.java [hunk 1]
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Ingyu Hwang, Mahmoud Romeh
+ * Copyright 2019 Yevhenii Voievodin, Robert Winkler
```
andrzejsydor, line 2: Why do you remove the old authors?

PR #2447, opened 2026-04-11
fix: prevent NPE in waitIntervalAfterException when thread is interrupted

### LINKED ISSUE #2368: NPE when retry thread is interrupted
### Given
Retry is configured to await for condition to be passed.
```java
Supplier<List<Path>> filesSupplier = () -> getSomething();
RetryConfig config = RetryConfig.<List<Path>>custom()
        .maxAttempts(10)
        .waitDuration(Duration.ofMillis(1000))
        .retryOnResult(List::isEmpty)        
        .build();
Retry retry = RetryRegistry.of(config)
    .retry("myRetry");
```

### When
Thread is interrupted, for instance due to scaling down application pod

### Then
Retry is not handled gracefully and complete with NPE:
version `2.2.0`
```java
java.lang.NullPointerException: Cannot throw exception because the return value of "java.util.concurrent.atomic.AtomicReference.get()" is null
	at io.github.resilience4j.retry.internal.RetryImpl$ContextImpl.waitIntervalAfterRuntimeException(RetryImpl.java:291)
	at io.github.resilience4j.retry.internal.RetryImpl$ContextImpl.onResult(RetryImpl.java:198)
	at io.github.resilience4j.retry.Retry.lambda$decorateSupplier$4(Retry.java:212)
	at io.github.resilience4j.retry.Retry.executeSupplier(Retry.java:361)
```

version `2.3.0`
```java
java.lang.NullPointerException: Cannot throw exception because the return value of "java.util.concurrent.atomic.AtomicReference.get()" is null
	at io.github.resilience4j.retry.internal.RetryImpl$ContextImpl.waitIntervalAfterRuntimeException(RetryImpl.java:290)
	at io.github.resilience4j.retry.internal.RetryImpl$ContextImpl.onResult(RetryImpl.java:197)
	at io.github.resilience4j.retry.Retry.lambda$decorateSupplier$5(Retry.java:302)
	at io.github.resilience4j.retry.Retry.executeSupplier(Retry.java:523)
```

### Expected
When thread is interrupted, retry should complete gracefully. 
When value of lastRuntimeException is missing, current exception should be thrown instead. 

```diff
try {
    sleepFunction.accept(interval);
} 
...
catch (Throwable ex) {
-   throw lastRuntimeException.get();
+   throw Optional.ofNullable(lastRuntimeException.get()).orElse(ex);
}
```

anuragg-saxenaa:
## Problem

When `Retry` is configured with `retryOnException` and the sleeping thread is interrupted during the wait interval, `waitIntervalAfterException` calls:

```java
throw lastException.get();
```

without null-checking the `AtomicReference`. If the reference is null (e.g. when retrying on a result predicate), this produces:

```
NullPointerException: Cannot throw exception because the return value of
"java.util.concurrent.atomic.AtomicReference.get()" is null
    at RetryImpl$ContextImpl.waitIntervalAfterException(RetryImpl.java:271)
    at RetryImpl$ContextImpl.onResult(RetryImpl.java:198)
```

## Fix

Mirror the null-guard already present in `waitIntervalAfterRuntimeException`:

- If `lastException` is null during `InterruptedException`: re-throw the `InterruptedException` (interrupt status already restored via `Thread.currentThread().interrupt()`)
- If `lastException` is null during any other `Throwable`: wrap in `IllegalStateException` rather than NPE

```java
} catch (InterruptedException ex) {
    Thread.currentThread().interrupt();
    Exception last = lastException.get();
    if (last != null) {
        throw last;
    }
    throw ex;
} catch (Throwable ex) {
    Exception last = lastException.get();
    if (last != null) {
        throw last;
    }
    throw new IllegalStateException("Unexpected exception during retry wait interval", ex);
}
```

## Testing

Added `shouldThrowInterruptedExceptionWhenInterruptedDuringRetryOnError` to `SupplierRetryTest` — verifies that interrupting the sleep during an exception-based retry path throws `InterruptedException` (not NPE) and leaves the thread interrupt flag set.

Fixes #2368

### DIFF
--- a/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
+++ b/resilience4j-retry/src/main/java/io/github/resilience4j/retry/internal/RetryImpl.java
@@ -268,9 +268,17 @@ public class RetryImpl<T> implements Retry {
                 sleepFunction.accept(interval);
             } catch (InterruptedException ex) {
                 Thread.currentThread().interrupt();
-                throw lastException.get();
+                Exception last = lastException.get();
+                if (last != null) {
+                    throw last;
+                }
+                throw ex;
             } catch (Throwable ex) {
-                throw lastException.get();
+                Exception last = lastException.get();
+                if (last != null) {
+                    throw last;
+                }
+                throw new IllegalStateException("Unexpected exception during retry wait interval", ex);
             }
         }
 
--- a/resilience4j-retry/src/test/java/io/github/resilience4j/retry/internal/SupplierRetryTest.java
+++ b/resilience4j-retry/src/test/java/io/github/resilience4j/retry/internal/SupplierRetryTest.java
@@ -394,6 +394,32 @@ public class SupplierRetryTest {
 
 
     @Test
+    public void shouldThrowInterruptedExceptionWhenInterruptedDuringRetryOnError() {
+        CheckedConsumer<Long> previousSleepFunction = RetryImpl.sleepFunction;
+        try {
+            RetryImpl.sleepFunction = sleep -> {
+                throw new InterruptedException("Interrupted!");
+            };
+
+            RetryConfig retryConfig = RetryConfig.<String>custom()
+                .maxAttempts(3)
+                .build();
+            Retry retry = Retry.of("id", retryConfig);
+
+            given(helloWorldService.returnHelloWorld()).willThrow(new HelloWorldException());
+            CheckedSupplier<String> decorated = Retry.decorateCheckedSupplier(retry,
+                helloWorldService::returnHelloWorld);
+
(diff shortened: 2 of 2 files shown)


andrzejsydor (approved):

anuragg-saxenaa:
Thanks for the review comments @Copilot. Acknowledged — both points are valid:

1. The null-lastException test gap: I'll add a dedicated test for the path where retrying via onError enters with a fresh lastException state.
2. The shouldThrowInterruptedExceptionWhenInterruptedDuringRetryOnError assertion type mismatch: confirmed the code path stores and re-throws the supplier exception, not the retry framework's wrapper. Will correct the test assertion.

Note: this PR is currently CONFLICTING with the latest master — I'll rebase and address both review points in the next push.

Issue #2343, opened 2025-08-22
Bump up from JDK 21 to 25 to avoid virtual thread pinned issue


ykhfree:
Upgraded to JDK 21 for virtual thread enforcement. ([PR](https://github.com/resilience4j/resilience4j/pull/2331))

However, there is still a risk of virtual thread pinned issues in JDK 21, and even though we have removed the pinned issue causing code from this code in resilence4j, the 3rd party libraries that resilence4j uses may still have pinned issue causing code (most notably Synchronize).

However, in JDK 24, the pinned issue caused by Synchronize has been eliminated. ([link](https://openjdk.org/jeps/491))

This requires an upgrade to the next LTS version, JDK 25. ([JDK releases](https://www.java.com/releases/))

JDK 25 will be released on September 16, 2025. I want to upgrade after that date.

PR #2449, opened 2026-04-21
(draft) Upgrade to jdk25 to completely resolve pinned issue of virtual threads

ykhfree:
Issue #2343

## Background

Resilience4j 3.x upgraded to JDK 21 to support virtual threads. However, on JDK 21–24, virtual threads could get **pinned** to their carrier thread when executing inside a `synchronized` block. To work around this, `InMemoryRegistryStore` adopted a `FutureHashMap`-style pattern that was significantly more complex than the original `ConcurrentHashMap.computeIfAbsent()` implementation.

**JDK 25 ships [JEP 491 (Synchronize Virtual Threads without Pinning)](https://openjdk.org/jeps/491)**, which eliminates that restriction. At the same time, **[JEP 506 (Scoped Values)](https://openjdk.org/jeps/506)** is promoted to a permanent feature, offering a more efficient alternative to `ThreadLocal` for context propagation. This PR raises the project baseline to JDK 25 so that we can actually take advantage of both JEPs.

### 1. Toolchain / Dependency Upgrades

| Item | Before | After | Reason |
|------|--------|-------|--------|
| `sourceCompatibility` / `targetCompatibility` | `VERSION_21` | `VERSION_25` | Use JEP 491/506 |
| Gradle Wrapper | 8.10 | **9.2.1** | Official JDK 25 support |
| GitHub Actions JDK | 21 | **25** | Build on 25 in CI |
| Kotlin JVM plugin | 2.1.21 | **2.3.0** | Supports `JvmTarget.JVM_25` |
| `org.sonarqube` | 3.5.0.2730 | 6.0.1.5171 | Gradle 9.x compatibility |
| `org.asciidoctor.jvm.convert` | 2.4.0 | 4.0.4 | Gradle 9.x compatibility |
| `org.gradle.test-retry` | 1.5.10 | **1.6.4** | Bundled ASM 9.9 — without this the plugin crashes with `Unsupported class file major version 69` when parsing Java 25 test classes |
| `me.champeau.jmh` | 0.6.8 | 0.7.3 | Gradle 9.x compatibility |
| `org.mockito:mockito-core` / `mockito-junit-jupiter` | 5.16.1 | **5.23.0** | 5.16.1 inline mock maker (ByteBuddy) fails on JDK 25 with `Could not modify all classes`; 5.23.0 includes the JDK 25 fixes |
| `spring6` | 6.1.1 | **6.2.18** | Spring 6.1.x bundles an older ASM repackage that cannot parse class file major version 69 (JDK 25), causing test-time component scanning of `resilience4j-spring6` classes to fail; 6.2.x bundles ASM 9.8 and 6.2.x is the final 6.x feature branch with JDK 25 as an LTS target |
| `spring-boot3` | 3.2.0 | **3.5.13** | Shares the Spring 6.1.x ASM issue via `spring-context`; Spring Boot officially supports JDK 25 from **3.5.5** (3.5.13 is the latest patch) |
| `spring-cloud3` | 3.1.5 | **4.3.2** | Spring Cloud release train matched to Spring Boot 3.5 is **2025.0 (Northfields)**; individual artifacts (`spring-cloud-context`, `spring-cloud-openfeign-*`) are at 4.3.x |

### 2. JEP 491 — Simplified `InMemoryRegistryStore`

Removed the `FutureHashMap(CompletableFuture<E>)` wrapping that was only there to avoid virtual-thread pinning, and restored the straightforward `ConcurrentHashMap<String, E>` + `computeIfAbsent()` implementation.

**Before (JDK 21–24):**

```java
private final ConcurrentHashMap<String, CompletableFuture<E>> entryMap;

// Winner/Loser pattern so mappingFunction runs outside the map lock
CompletableFuture<E> created = new CompletableFuture<>();
CompletableFuture<E> future = entryMap.putIfAbsent(key, created);
if (future == null) {
    try {
        E value = mappingFunction.apply(key);
        future.complete(value);
    } catch (Throwable t) { ... }
}
return future.join();
```

**After (JDK 25):**

```java
private final ConcurrentHashMap<String, E> entryMap;

// JEP 491: synchronized / computeIfAbsent no longer pins virtual threads
return entryMap.computeIfAbsent(key, mappingFunction);
```

`putIfAbsent`, `remove`, `replace`, and `values` have all been simplified to direct calls without any `CompletableFuture` wrapping. Net deletion: **~102 lines**.

### 3. JEP 506 — Scoped Values Based Context Propagator

The existing `ThreadLocal`-based `ContextPropagator` carried non-trivial overhead in virtual-thread environments (per-thread storage cost, plus the historical pinning concern). Two new types are added as a modern alternative:

* **`ScopedValuePropagator<T>`** — propagates a single `ScopedValue<T>` via the `ContextPropagator<T>` interface.
* **`ScopedValueContext`** — captures multiple `ScopedValuePropagator`s into one scope and exposes `runWithContext(Runnable)` / `callWithContext(Callable)`.

Example usage:

```java
private static final ScopedValue<String> REQUEST_ID = ScopedValue.newInstance();
ScopedValuePropagator<String> p = ScopedValuePropagator.of(REQUEST_ID);

ThreadPoolBulkheadConfig config = ThreadPoolBulkheadConfig.custom()
    .contextPropagator(p)
    .build();

ScopedValue.where(REQUEST_ID, "req-123").run(() ->
    bulkhead.executeSupplier(() -> doWork(p.getCapturedValue().orElse("")))
);
```

Accompanying tests — `ScopedValuePropagatorTest`, `ScopedValueContextTest`, and `ScopedValueBulkheadIntegrationTest` — are added (~1,300 lines total).

### 4. Misc

* Added `kotlin.jvm.target.validation.mode=IGNORE` to `gradle.properties`. Kotlin 2.3.0 supports `JvmTarget.JVM_25`, but strict validation is too aggressive when Kotlin code is consumed by mixed-JVM-target modules in this multi-module build.
* A handful of tests were reworked around `ThreadModeTestBase` so they run parameterized across platform and virtual threads.

### 5. Review Follow-ups

Two review findings on the initial commit were addressed in a follow-up:

* **`ScopedValuePropagator.retrieve()` fallback** — the original `retrieve()` only checked `scopedValue.isBound()`, so once a decorated task ran on a worker thread (where the real `ScopedValue` is never re-bound — `copy()` only writes to a per-thread capture map), any _second_ decoration on that worker thread would read `Optional.empty()` and silently drop the value. `retrieve()` now falls back to the capture map when `isBound()` is false, so multi-hop propagation works. A new test `shouldPropagateValueAcrossMultipleAsyncHops` covers the regression.
* **`AbstractRegistry.computeIfAbsent` event emission** — the simplified `InMemoryRegistryStore.computeIfAbsent` exposed a latent issue where `EntryAddedEvent` was published _inside_ the CHM mapping function. A consumer that recursively added another entry (especially one hashing to the same bin) would trip `ConcurrentHashMap`'s recursive-update detection (`IllegalStateException`). The event is now emitted _after_ the CHM call returns, gated by an `AtomicBoolean` that records whether the mapping function actually ran. A new test `shouldAllowRecursiveRegistrationFromEntryAddedConsumer` covers this. The JEP 491 simplification in `InMemoryRegistryStore` is preserved.

## Known Limitations

* **JaCoCo coverage report temporarily disabled**: JaCoCo 0.8.12 does not yet understand class file major version 69 (Java 25). `jacocoTestReport` and `testCodeCoverageReport` are set to `enabled = false` and will be re-enabled once an upstream JaCoCo release adds Java 25 support. **Tests themselves still execute normally.**
* **Gradle 9.2.1 shutdown NPE**: On JDK 25 + Gradle 9.2.1 we occasionally see an NPE in `GroupingProgressLogEventGenerator.bufferOutput()`. This is an internal Gradle logging-subsystem cleanup bug and does not affect test outcomes. Raw results are still written to `build/test-results/test/*.xml`. Expected to be fixed in Gradle 9.3+. (Note: this is a separate issue from the test-retry ASM crash, which is resolved by the 1.6.4 bump in the table above.)

## Migration / Compatibility

* **Consumer impact**: applications using Resilience4j as a library will now require **JDK 25 or later**.
* **No API breakage**: existing `ContextPropagator`-based `ThreadLocal` code continues to work unchanged. `ScopedValuePropagator` is strictly **additive**.
* **Behavioral compatibility**: the `InMemoryRegistryStore` simplification does not change its externally observable contract (register / find / remove semantics).
* **Spring baseline bumps are test-scope only**: the `spring6` / `spring-boot3` / `spring-cloud3` version upgrades are declared in `compileOnly` and `testImplementation` dependency scopes in the affected modules. Downstream applications do NOT need to upgrade their Spring Framework / Spring Boot / Spring Cloud to the new versions; they only need to continue meeting the modules' existing `compileOnly` contracts (Spring 6.x for `resilience4j-spring6`, Spring Boot 3.x for `resilience4j-spring-boot3`).

## References

* Issue [Bump up from JDK 21 to 25 to avoid virtual thread pinned issue #2343](https://github.com/resilience4j/resilience4j/issues/2343)
* [JEP 491: Synchronize Virtual Threads without Pinning](https://openjdk.org/jeps/491)
* [JEP 506: Scoped Values](https://openjdk.org/jeps/506)
* Preceding virtual thread PR: [Feature #2224 : Apply virtual thread #2335](https://github.com/resilience4j/resilience4j/pull/2335) (already merged into master)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ allprojects {
     }
 
     jacoco {
-        toolVersion = "0.8.12"
+        toolVersion = "0.8.14"
     }
 }
 
@@ -89,8 +89,8 @@ configure(project.coreProjects) {
     }
 
     tasks.withType(JavaCompile) {
-        sourceCompatibility = JavaVersion.VERSION_21
-        targetCompatibility = JavaVersion.VERSION_21
+        sourceCompatibility = JavaVersion.VERSION_25
+        targetCompatibility = JavaVersion.VERSION_25
         options.deprecation = true
         options.encoding = "UTF-8"
         options.compilerArgs += ["-Xlint:unchecked", "-parameters"]
--- a/.github/workflows/gradle-build.yml
+++ b/.github/workflows/gradle-build.yml
@@ -30,10 +30,10 @@ jobs:
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
 
-      - name: Set up JDK 21
+      - name: Set up JDK 25
         uses: actions/setup-java@v5
         with:
-          java-version: '21'
+          java-version: '25'
           distribution: 'temurin'
           cache: 'gradle'
 
--- a/docs/superpowers/plans/2026-04-22-jdk25-spring-bump-plan.md
+++ b/docs/superpowers/plans/2026-04-22-jdk25-spring-bump-plan.md
@@ -1,0 +1,387 @@
+# JDK 25 Spring Bump Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** `resilience4j-spring6` / `resilience4j-spring-boot3` 모듈의 JDK 25 테스트 실패(ASM "Unsupported class file major version 69")를 Spring 6 / Spring Boot 3 / Spring Cloud 3 라이브러리 버전 bump으로 해결한다.
+
+**Architecture:** `gradle/libs.versions.toml`의 3개 버전 변수(`spring6`, `spring-boot3`, `spring-cloud3`)를 JDK 25를 지원하는 현행 최신 GA 버전으로 올린 뒤, 버전 점프 과정에서 발생하는 테스트 코드 fallout을 최소 범위로 수정한다. 프로덕션 코드는 수정하지 않는다.
+
+**Tech Stack:** Gradle 9.2.1, Java 25, Spring Framework 6.2.18, Spring Boot 3.5.13, Spring Cloud 4.3.2 (Northfields 2025.0 release train), JUnit 5 (Jupiter 5.x via JUnit BOM 6.0.3)
+
+**Working directory:** `/Users/al02628768/private/resilience4j` (branch `feature/2343-bump-up-jdk-to-25`, worktree 아님 - 기존 feature branch 직접 작업)
+
+**Design doc:** `docs/superpowers/specs/2026-04-22-jdk25-spring-bump-design.md`
+
+---
+
+## File Structure
+
+Primary change:
+- **Modify:** `gradle/libs.versions.toml` (versions block, 3 lines)
+
+Potentially touched during fallout fixes (실제 실패 발생 시에만):
+- `resilience4j-spring6/src/test/java/**/*.java` (테스트 코드)
+- `resilience4j-spring-boot3/src/test/java/**/*.java` (테스트 코드)
+- `resilience4j-spring-boot3/src/main/java/**/*.java` (autoconfigure spring.factories 경로 변경 등 필요시)
+- `resilience4j-spring6/src/main/java/**/*.java` (불가피할 경우만)
+- `resilience4j-spring-boot3/src/test/resources/application*.yml` 또는 `.properties` (property key deprecation)
+
+프로덕션 코어 코드(`resilience4j-core` 등)는 건드리지 않는다.
+
+---
+
+## Task 1: Record baseline failures
+
+`spring6`와 `spring-boot3` 모듈의 현재 실패 상태를 기록한다. 이후 수정이 의도한 에러만 해결하는지 검증 기준이 된다.
+
+**Files:**
+- Read-only. 결과는 터미널 출력으로만 기록.
+
+- [ ] **Step 1: Run spring6 tests, capture failure summary**
+
+Run:
+```bash
+./gradlew :resilience4j-spring6:test --no-daemon --continue 2>&1 | tail -40
+```
+
+Expected: FAIL. 실패 메시지에 다음 문자열이 반드시 포함되어야 한다:
+```
+java.lang.IllegalArgumentException: Unsupported class file major version 69
+```
+
+검증: `grep -c 'Unsupported class file major version 69' resilience4j-spring6/build/test-results/test/*.xml` 결과 ≥ 1.
+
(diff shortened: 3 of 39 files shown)


build.gradle
```diff
@@ -103,9 +103,14 @@ configure(project.coreProjects) {
             failOnPassedAfterRetry = true
         }
         useJUnitPlatform()
+        // Disable JaCoCo for Java 25+ as JaCoCo 0.8.12 doesn't support class file major version 69
```
lobaorn-bitso: Yeah but jacoco 0.8.14 does support, why keep 0.8.12 in this circumstance?

build.gradle
```diff
@@ -143,4 +148,5 @@ sonarqube {
     }
 }
 
-tasks.check.dependsOn tasks.testCodeCoverageReport
```
lobaorn-bitso, line 146: Jacoco 0.8.14 does support it.

build.gradle
ykhfree: Thank you for your comment.  This PR is currently in draft status.  Since various aspects have not yet been completed, I plan to check the lib version again.

build.gradle
ykhfree: Thank you for your comment.  This PR is currently in draft status.  Since various aspects have not yet been completed, I plan to check the lib version again.

ykhfree: Fixed

build.gradle
ykhfree: Fixed

gdupontf:
Do you think you could deliver this without the context propagation work?
With [#2428](https://github.com/resilience4j/resilience4j/pull/2428) I've set it up so context propagation in Resilience4J can be phased out, eventually redirecting users to micrometer context-propagation, where you might be able to contribute this scoped value-based propagation. What do you think?

RobWin:
@copilot resolve the merge conflicts in this pull request

Issue #2451, opened 2026-04-23
Virtual thread scheduler optimization - Hedge lifecycle + Core one-shot primitive


ykhfree:
While investigating HedgeConcurrencyTest, noticed that all five tests skip virtual-thread mode with Assume.assumeFalse citing "daemon thread limitations" — meaning Hedge's virtual-thread code path has no test coverage and virtual-mode regressions can slip in silently.
                                                                                                                                                                                                                              
Digging into the root cause: Hedge owns an internal ContextAwareScheduledThreadPoolExecutor but exposes no way to shut it down. In platform mode NamingThreadFactory forces setDaemon(false) so the JVM cleans up eventually; in virtual mode, virtual threads are permanently daemon (setDaemon(false) throws), so scheduler threads leak across test runs with no deterministic cleanup — hence the skip.
                                                                                                                                                                                                                              
Separately, while looking at the internal scheduler, realized Hedge's hot path is "schedule a trigger timer, cancel it when primary wins" — cancellation is the common case, not firing. The STPE-based pattern carries queue-management overhead that this workload does not need.
                                                                                                                                                                                                                              
These two findings share the same locus (Hedge's internal scheduler), so tackling them together was natural: fix the lifecycle contract (AutoCloseable) and, while the area is already being touched, introduce a primitive that matches the actual workload.

PR #2452, opened 2026-04-23
issue 2451: Implement AutoCloseable for Hedge and create OneShotDelayedScheduler

ykhfree:
issue #2451

## Summary

Two related changes delivered together:

1. **`Hedge` now implements `AutoCloseable`** — standardizes the lifecycle of the internal scheduler that Hedge owns, unblocking virtual-thread-mode tests that had to be skipped.
2. **New Core primitive `OneShotDelayedScheduler`** — a dedicated one-shot delayed-task scheduler that replaces the STPE-based trigger timer inside `HedgeImpl`. Uses an **asymmetric implementation**: per-timer virtual threads in virtual mode (3× faster than the previous STPE pattern) and a shared STPE singleton in platform mode (no regression vs. the previous pattern). Benchmarked in both modes.

### Part 1 — Hedge lifecycle gap was blocking virtual-thread testing

`HedgeConcurrencyTest` has 5 methods that all start with:

```java
Assume.assumeFalse(
    "Hedge has known issues with virtual threads due to daemon thread limitations",
    isVirtualThreadMode());
```

This guard was added because `HedgeImpl` silently owns a `ContextAwareScheduledThreadPoolExecutor` and never exposes a way to shut it down. The issue only materializes under virtual threads because:

- In **platform mode**, `NamingThreadFactory` forces `setDaemon(false)`, so the scheduler workers keep the JVM alive while pending scheduled tasks exist. Tests tolerate this because the suite completes and the JVM eventually exits cleanly.
- In **virtual mode**, JVM spec makes virtual threads **permanently daemon** (`setDaemon(false)` throws `IllegalArgumentException`). `NamingThreadFactory` correctly skips that branch for virtual threads. Result: tests that create many Hedges accumulate leaked scheduler threads across runs with no deterministic cleanup path.

This is not a hedge race-algorithm bug — it's a **resource ownership / lifecycle gap** that the virtual-thread path made visible. The fix is to follow the precedent already present in the same codebase: `ThreadPoolBulkhead extends AutoCloseable` with an explicit `close()` that shuts down the owned executor.

### Part 2 — The scheduler model itself was suboptimal for Hedge's workload

While closing Part 1, the scheduler pattern used by `HedgeImpl` was inspected. `HedgeImpl` used `ScheduledThreadPoolExecutor` (via `ContextAwareScheduledThreadPoolExecutor`) with virtual-thread workers to schedule its hedge-trigger timer. Two observations:

- Hedge's dominant runtime path is **"schedule a timer, cancel it when primary completes first"** — a one-shot delayed event where cancellation is the common case.
- `ScheduledThreadPoolExecutor` is built around a fixed pool of long-lived workers pulling from a `DelayedWorkQueue`. That design is appropriate for sustained periodic scheduling, but for "schedule + almost always cancel" it adds queue-management and worker-dispatch overhead that a one-shot-per-timer model avoids.

To quantify, three alternatives were compared with JMH (see `HedgeSchedulerBenchmark`):

- **Option A** — `CompletableFuture.delayedExecutor(...)` + `cancel()`.
- **Option B** — `Thread.ofVirtual() / Thread.ofPlatform()` per timer + `parkNanos()` + `interrupt()` (one-shot-per-timer).
- **Option C** — shared `ScheduledThreadPoolExecutor` + `schedule(...).cancel(...)` (the pattern Hedge used before this change).

Two preliminary runs ruled out Option A: `CompletableFuture.delayedExecutor` + `cancel(true)` does **not** release the pending delayed dispatch (documented JDK behaviour). Under Hedge's cancel-heavy load this produces unbounded memory growth that OOMs the JVM in both virtual and platform modes.

The B vs C comparison is mode-dependent:

- **Virtual mode**: per-timer virtual threads (B) are ~3× faster than the shared STPE (C). Virtual thread creation cost is ~1 μs; no queue management.
- **Platform mode**: per-timer platform threads (B) are **~130× slower** than the shared STPE (C). Platform thread creation costs ~1 ms + ~1 MB stack, dominating every schedule + cancel.

Neither single primitive wins in both modes, so `OneShotDelayedScheduler` adopts an **asymmetric implementation**: Option B in virtual mode, Option C (shared daemon STPE singleton) in platform mode. The observable contract — race-free state machine (`PENDING → FIRED | CANCELLED`), `ContextPropagator` + MDC propagation, daemon lifetime — is identical in both modes.

## Benchmark evidence

Ran `HedgeSchedulerBenchmark` on JDK 21 (Corretto 21.0.6), 4 threads, -Xmx2g, JMH 1.35, warmup 2×2s, measurement 3×3s. Variants B and C were each measured in both `virtual` and `platform` thread modes (Option A excluded — OOMs in both modes, see Motivation).

### `scheduleAndCancel` — fast-primary path (dominant in production)

Schedule a 10 ms timer and immediately cancel. Measures scheduling + cancellation overhead when the timer never fires.

| Mode | Variant | Throughput (ops/μs) | AvgTime (μs/op) |
|---|---|---:|---:|
| **virtual** | **B — per-timer thread** (this PR) | **9.750 ±1.756** | **0.431 ±0.098** |
| virtual | C — shared STPE (previous baseline) | 3.270 ±1.382 | 1.280 ±0.497 |
| platform | B — per-timer thread | 0.029 ±0.002 | 142.715 ±8.304 |
| **platform** | **C — shared STPE** (this PR) | **3.756 ±1.785** | **1.109 ±0.279** |

**Results**:

- **Virtual mode**: Option B over Option C → throughput **3.0×**, latency **3.0×** improvement (0.431 μs/op vs 1.280 μs/op).
- **Platform mode**: Option B over Option C is a **~130× regression** (142.7 μs/op vs 1.1 μs/op) — platform thread creation at ~1 ms/timer dominates the cancel path. This is why `OneShotDelayedScheduler` reuses a shared daemon STPE in platform mode.
- **Shipped helper**: virtual → 0.43 μs/op (3× faster than the previous per-Hedge STPE); platform → 1.11 μs/op (on par with the previous per-Hedge STPE, no regression).

### `scheduleAndFire` — slow-primary path

Schedule a 1 ms timer and wait for it to fire. All four combinations land around ~1300 μs/op and ~0.003 ops/μs — dominated by OS timer resolution + JMH overhead. Differences are within noise, which is expected: this path is bounded by the delay itself, not scheduling overhead.

Reproduce:
```bash
./gradlew :resilience4j-hedge:jmh -PjmhIncludes=HedgeSchedulerBenchmark
```

Raw results: `resilience4j-hedge/build/results/jmh/results.txt`.

### Caveat

The benchmark compares raw primitives. The shipped `OneShotDelayedScheduler` adds thin wrappers (MDC snapshot/restore, `ContextPropagator.decorateRunnable`, `AtomicReference<State>` race-free cancel). Wrapper overhead is minimal compared to `parkNanos` + `LockSupport` cost (virtual) and to STPE queue dispatch cost (platform), so real-world helper performance should track the Option-of-choice in each mode closely, directionally if not exactly.

### New

- `resilience4j-core/.../OneShotDelayedScheduler.java` — public helper:
  - `schedule(Duration, List<? extends ContextPropagator>, Runnable) → Cancellation`
  - Virtual mode: `Thread.ofVirtual()` + `parkNanos()` per timer.
  - Platform mode: shared daemon-backed `ScheduledExecutorService` singleton (lazy-initialised, `availableProcessors()` pool size).
  - Race-free state machine using `AtomicReference<State>` — identical observable contract in both modes.
  - MDC + `ContextPropagator` captured on caller, restored on firing thread.
- `resilience4j-core/.../OneShotDelayedSchedulerTest.java` — 20 cases (10 methods × platform/virtual) covering delay, cancel idempotency, post-fire cancel, MDC propagation, context propagator, concurrent schedules, null validation.
- `resilience4j-hedge/.../HedgeLifecycleTest.java` — 8 cases covering rejection after close, close idempotency, try-with-resources, registry-cascading close.
- `resilience4j-hedge/src/jmh/.../HedgeSchedulerBenchmark.java` — the benchmark used above; retained for future regression measurement.

### Modified

- `resilience4j-hedge/.../Hedge.java` — `extends AutoCloseable`, `close()` declared with no checked exception.
- `resilience4j-hedge/.../HedgeRegistry.java` — `default void close()` that cascades to all managed hedges.
- `resilience4j-hedge/.../internal/HedgeImpl.java` —
  - Implements `close()` using the graceful-then-force pattern from `FixedThreadPoolBulkhead.close()` (shutdown → 5s await → `shutdownNow()`).
  - Trigger timer scheduling migrated to `OneShotDelayedScheduler.schedule(...)`.
  - Fail-fast guard added: `decorateCaller` throws `RejectedExecutionException` if `configuredHedgeExecutor.isShutdown()` (preserves the documented close contract).
- `resilience4j-hedge/.../HedgeConcurrencyTest.java` — `Assume.assumeFalse(... isVirtualThreadMode())` removed from all 5 methods; `track()` helper and `@After` cleanup added so virtual-mode runs are deterministic.

### DIFF
--- a/resilience4j-core/src/main/java/io/github/resilience4j/core/OneShotDelayedScheduler.java
+++ b/resilience4j-core/src/main/java/io/github/resilience4j/core/OneShotDelayedScheduler.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright 2026 kanghyun.yang
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.resilience4j.core;
+
+import org.slf4j.MDC;
+
+import java.time.Duration;
+import java.util.*;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
+
+/**
+ * Primitive for one-shot delayed task execution with cheap cancellation.
+ * <p>
+ * Intended for the "schedule a timer, usually cancel it before it fires" pattern
+ * common in resilience primitives (hedge trigger timers, timeout guards, etc.).
+ * Implementation varies by {@link ExecutorServiceFactory#getThreadType()}:
+ * <ul>
+ *   <li><b>Virtual mode</b>: each timer is backed by its own virtual thread that
+ *       parks for the delay and runs the task on expiry. Leverages the cheap
+ *       creation cost of virtual threads and avoids queue management / worker
+ *       dispatch entirely. Cancellation interrupts the timer thread.</li>
+ *   <li><b>Platform mode</b>: a process-wide daemon-backed
+ *       {@link ScheduledExecutorService} singleton is reused across all calls.
+ *       Per-timer platform thread creation would cost ~1ms + ~1MB stack, so a
+ *       shared pool is preferred. Cancellation removes the task from the
+ *       scheduler queue.</li>
+ * </ul>
+ * <p>
+ * In both modes:
+ * <ul>
+ *   <li>A race-free state machine ({@code PENDING → FIRED | CANCELLED}) ensures
+ *       the task runs at most once and {@code cancel()} returns {@code true}
+ *       only if it prevented execution.</li>
+ *   <li>MDC and {@link ContextPropagator} values are captured on the caller and
+ *       restored on the firing thread, mirroring
+ *       {@code ContextAwareScheduledThreadPoolExecutor} semantics.</li>
+ * </ul>
+ * <p>
+ * This helper is a stateless gateway and does not expose {@code shutdown()}.
+ * The shared platform scheduler uses daemon threads so JVM shutdown is never
+ * blocked by pending timers.
+ *
+ * @author kanghyun.yang
+ * @since 3.0.0
+ */
+public final class OneShotDelayedScheduler {
+
+    private static final String VIRTUAL_PREFIX = "OneShotDelayed-v-";
+    private static final String PLATFORM_PREFIX = "OneShotDelayed-p-";
+
+    private enum State { PENDING, FIRED, CANCELLED }
+
+    /**
+     * Lazy holder for the shared platform-mode scheduler. Initialised on first
+     * use so processes running purely in virtual mode never pay for a platform
+     * thread pool. Threads are daemon so the pool never blocks JVM exit.
+     */
+    private static final class PlatformSchedulerHolder {
+        static final ScheduledExecutorService INSTANCE = newPlatformScheduler();
+    }
+
+    private OneShotDelayedScheduler() {
+    }
+
+    static ScheduledThreadPoolExecutor newPlatformScheduler() {
+        ScheduledThreadPoolExecutor scheduler = new ScheduledThreadPoolExecutor(
+            Math.max(1, Runtime.getRuntime().availableProcessors()),
+            Thread.ofPlatform().name(PLATFORM_PREFIX, 0).daemon(true).factory());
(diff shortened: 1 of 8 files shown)


gdupontf:
I'm trying to get the ball rolling on [#2428](https://github.com/resilience4j/resilience4j/pull/2428). While I was waiting for feedback I saw your PR. Do you think it would be easier to rebase your PR on top of it or it might be better the other way around?

ykhfree:
> gdupontf

I'm really sorry for the late reply.
Thanks for checking. I think it is better to keep #2452 independent for now and rebase it on current master rather than on #2428.

#2452 is a focused Hedge lifecycle / one-shot timer fix, while #2428 is a broader scheduler-injection change that may still be split based on maintainer feedback. I don’t see a strong direct file conflict at the moment, so making either PR depend on the other may add unnecessary coupling.

If the scheduler abstraction from #2428 later needs to cover Hedge as well, I’m happy to adapt #2452 once that direction is settled.

ykhfree:
@RobWin 
All fixes in this PR have been resolved.

PR #2461, opened 2026-05-08
#issue 2450: Enhance virtual thread support in RetryAspect and TimeLimiterAspect by using ExecutorServiceFactory for fallback schedulers

### LINKED ISSUE #2450: Identify and fix any files that ignore the user's resilience4j thread type specification.
- Introduced environment variables such as `resilience4j.thread.type=virtual` for virtual thread support.
- However, I noticed that some files (4) ignore the above settings and blindly use the platform thread.
    - `resilience4j-spring/.../TimeLimiterAspect.java`
    - `resilience4j-spring/.../RetryAspect.java`
    - `resilience4j-spring6/.../TimeLimiterAspect.java`
    - `resilience4j-spring6/.../RetryAspect.java`
 - I will fix this bug accordingly

ykhfree:
Closes #2450

## Summary

Spring AOP aspects for `Retry` and `TimeLimiter` silently bypassed the `resilience4j.thread.type` configuration whenever the optional `ContextAwareScheduledThreadPoolExecutor` bean was absent. Their fallback path called `java.util.concurrent.Executors.newScheduledThreadPool(...)` directly, which uses the default `ThreadFactory` and therefore **always produces platform threads** — even when the user enabled virtual threads via system property, environment variable, or Spring Boot YAML.

This PR routes those fallback constructions through `ExecutorServiceFactory.newScheduledThreadPool(size, poolName)` so the documented `resilience4j.thread.type` switch is honored end-to-end. Two parameterized JUnit 5 tests (`RetryAspectVirtualThreadTest`, `TimeLimiterAspectVirtualThreadTest`) lock the contract in place against both `platform` and `virtual` modes.

## Motivation

`README.adoc` and `CLAUDE.md` advertise a single switch — `resilience4j.thread.type=virtual` — that flips Resilience4j's internal schedulers onto Java 21+ virtual threads:

```yaml
resilience4j:
  thread:
    type: virtual
```

That switch is read by `ExecutorServiceFactory` (in `resilience4j-core`) and is the single entry point that decides between platform and virtual `ThreadFactory` instances. Every internal scheduler is supposed to go through this factory.

The Spring (legacy) and Spring 6 aspects, however, contained a fallback branch that constructed a scheduler **without consulting the factory**:

```java
this.xxxExecutorService = contextAwareScheduledThreadPoolExecutor != null ?
    contextAwareScheduledThreadPoolExecutor :
    Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
```

When the autoconfigured `ContextAwareScheduledThreadPoolExecutor` bean is present, the configured behavior holds. But in any "bare" wiring — a custom `@Configuration` that defines the aspect manually, or any environment that has the aspect bean but not the optional context-aware executor — the user's `virtual` setting was silently ignored. The aspect kept running on platform threads while the rest of the library obeyed the switch.

This is a **configuration-contract defect**, not a performance discussion. Operators who turn on virtual mode and size their services around its concurrency assumptions deserve all internal schedulers to actually be virtual — including the ones owned by the Spring aspects.

### Modified

Both files in `resilience4j-spring6` (Spring Boot 3 / Spring Framework 6):

- `resilience4j-spring6/.../spring6/timelimiter/configure/TimeLimiterAspect.java`
- `resilience4j-spring6/.../spring6/retry/configure/RetryAspect.java`

Each constructor now calls:

```java
ExecutorServiceFactory.newScheduledThreadPool(
    Runtime.getRuntime().availableProcessors(),
    "<AspectName>");
```

Pool names: `"TimeLimiterAspect"` and `"RetryAspect"` — visible in thread dumps and JFR (`TimeLimiterAspect-vthread-N` or `TimeLimiterAspect-thread-N`, depending on mode), giving clear ownership for any thread that originated from the aspect's fallback path.

The `import java.util.concurrent.*;` wildcard was also expanded to explicit imports (`CompletionException`, `CompletionStage`, `ScheduledExecutorService`, `TimeUnit`) in the two Spring 6 files; the only outwardly observable side effect is that `Executors` is no longer imported, since it is no longer referenced.

The legacy `resilience4j-spring` module contains the same defect pattern, but that module is no longer included in `settings.gradle` (removed in commit `ab7dca4b` together with the Spring Boot 2 line). It is therefore not part of any build artifact and is intentionally not touched here.

## Backwards compatibility

| Change | Impact | Notes |
|---|---|---|
| `resilience4j.thread.type=virtual` is now honored in the Spring 6 aspect fallback path | Behavioral fix, matches documented contract | Users who relied on the broken behavior to keep aspects on platform threads despite enabling virtual mode globally would see a change. There is no documented or supported way that intent is expressed today. |
| Fallback scheduler threads are now **daemon** (in platform mode) | Minor behavior shift | The previous `Executors.newScheduledThreadPool(...)` produced non-daemon platform threads via the default `ThreadFactory`. `ExecutorServiceFactory.newPlatformThreadFactory(...)` produces daemon platform threads. For an internal aspect-owned scheduler this is the safer default — it cannot delay JVM shutdown. The aspect already implements `AutoCloseable.close()` for clean shutdown, so the daemon flag is a defense-in-depth improvement, not the primary lifecycle mechanism. |
| Thread names changed from generic `pool-N-thread-M` to `TimeLimiterAspect-(v)thread-N` / `RetryAspect-(v)thread-N` | Observability improvement | Any monitoring grep that pinned to `pool-N-thread-M` would need to widen its pattern; the new names provide aspect-scoped attribution that the old names lacked. |
| Pool size, scheduler interface (`ScheduledExecutorService`), `close()` semantics | Unchanged | `Runtime.availableProcessors()` worker count and `AutoCloseable.close()` graceful-shutdown sequence are preserved exactly. |

### DIFF
--- a/resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/retry/configure/RetryAspect.java
+++ b/resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/retry/configure/RetryAspect.java
@@ -16,12 +16,13 @@
 package io.github.resilience4j.spring6.retry.configure;
 
 import io.github.resilience4j.core.ContextAwareScheduledThreadPoolExecutor;
+import io.github.resilience4j.core.ExecutorServiceFactory;
 import io.github.resilience4j.core.functions.CheckedSupplier;
 import io.github.resilience4j.core.lang.Nullable;
-import io.github.resilience4j.spring6.fallback.FallbackExecutor;
 import io.github.resilience4j.retry.RetryConfig;
 import io.github.resilience4j.retry.RetryRegistry;
 import io.github.resilience4j.retry.annotation.Retry;
+import io.github.resilience4j.spring6.fallback.FallbackExecutor;
 import io.github.resilience4j.spring6.spelresolver.SpelResolver;
 import io.github.resilience4j.spring6.utils.AnnotationExtractor;
 import org.aspectj.lang.ProceedingJoinPoint;
@@ -37,7 +38,10 @@ import org.springframework.core.Ordered;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.List;
-import java.util.concurrent.*;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 /**
  * This Spring AOP aspect intercepts all methods which are annotated with a {@link Retry}
@@ -93,7 +97,9 @@ public class RetryAspect implements Ordered, AutoCloseable {
         this.spelResolver = spelResolver;
         this.retryExecutorService = contextAwareScheduledThreadPoolExecutor != null ?
             contextAwareScheduledThreadPoolExecutor :
-            Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
+            ExecutorServiceFactory.newScheduledThreadPool(
+                Runtime.getRuntime().availableProcessors(),
+                "RetryAspect");
     }
 
     @Pointcut(value = "@within(retry) || @annotation(retry)", argNames = "retry")
--- a/resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/timelimiter/configure/TimeLimiterAspect.java
+++ b/resilience4j-spring6/src/main/java/io/github/resilience4j/spring6/timelimiter/configure/TimeLimiterAspect.java
@@ -17,14 +17,15 @@
 package io.github.resilience4j.spring6.timelimiter.configure;
 
 import io.github.resilience4j.core.ContextAwareScheduledThreadPoolExecutor;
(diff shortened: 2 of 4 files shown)


resilience4j-cache/build.gradle
```diff
@@ -2,6 +2,9 @@ dependencies {
     api(project(":resilience4j-core"))
     api(libs.jcache)
 
+    testRuntimeOnly(libs.junit.vintage.engine)
```
gavlyukovskiy: This looks unnecessary, why add JUnit 4 back?

ykhfree: You're right, this should not be necessary. I missed that `CacheConcurrencyTest` was still on JUnit 4. I'll convert it to JUnit Jupiter and remove the Vintage dependency.

ykhfree: I fixed.

ykhfree:
@RobWin 
All fixes in this PR have been resolved.

PR #2475, opened 2026-05-27
Partially addresses #2451: Hedge AutoCloseable lifecycle, daemon scheduler, VT test coverage

zbnerd:
Partially addresses #2451 by adding deterministic Hedge/HedgeRegistry lifecycle cleanup and daemon scheduler threads.

## What's included

- Hedge extends AutoCloseable with graceful shutdown (5s drain → shutdownNow)
- HedgeRegistry cascade close — closes all managed Hedge instances
- Daemon scheduler threads via `NamingThreadFactory` and `ContextAwareScheduledThreadPoolExecutor.Builder.daemon(boolean)`
- Virtual-thread concurrency tests enabled (VT skip removed — virtual threads are always daemon)
- HedgeConcurrencyTest: thread leak prevention via createdHedges tracking + tearDown close
- InMemoryHedgeRegistry.close() — exception-safe, continues closing remaining hedges if one fails
- JUnit 4 → 5 migration for hedge test suite (kept to avoid mixed JUnit 4/5 dependencies)
- NamingThreadFactory: removed deprecated SecurityManager check (JDK 17+)

## Not included (follow-up)

- Core one-shot scheduling primitive from #2452 (`OneShotDelayedScheduler`)
- If maintainers prefer solving #2451 in one step, #2452 is the more complete approach

## Test results

All hedge tests pass in both platform and virtual thread modes (10/10, 0 skipped, 0 failures).


ykhfree:
This PR seems to be the same as the [PR I registered](https://github.com/resilience4j/resilience4j/pull/2452).
What do you think?

zbnerd:
> This PR seems to be the same as the [PR I registered](https://github.com/resilience4j/resilience4j/pull/2452). What do you think?

Thanks for pointing that out.

I did review #2452 while working on this issue.

Both PRs seem to address the lifecycle problem from #2451, but my approach was a bit different. I focused on adding AutoCloseable support directly to Hedge/HedgeRegistry, daemon scheduler configuration, and virtual-thread compatibility around the existing scheduler infrastructure.

I think there is definitely overlap in the problem being solved, but the implementation direction is somewhat different.

It would be great to hear which approach the maintainers prefer.

zbnerd:
Hi maintainers,

I have addressed the Copilot review feedback and completely cleaned up the codebase:
- Resolved the `ExecutorService` thread leaks in `HedgeImplTest` and `InMemoryHedgeRegistryTest` by properly scoping them using try-with-resources.
- Strengthened `shouldDrainInFlightWorkOnClose` to strictly verify that `close()` blocks and gracefully drains actual in-flight work using cross-latch synchronization.
- Ensured `NamingThreadFactoryTest` starts and joins threads correctly to validate daemon behavior safely.

Regarding the overall scope of #2451, this PR intentionally focuses **only on the lifecycle cleanup and daemon scheduler threads** portion to establish a deterministic cleanup mechanism first.

### Key Points for Reviewers:
* **Partially addresses #2451:** It ensures `Hedge` and `HedgeRegistry` lifecycle management works deterministically via `AutoCloseable`.
* **Distinct from #2452:** This PR does **NOT** implement the `OneShotDelayedScheduler` proposed in #2452, nor does it remove the virtual-thread skips in `HedgeConcurrencyTest` yet (which should be handled as a follow-up).

If you prefer resolving the entire scope of #2451 (including replacing the scheduler architecture) in a single step, **#2452** might be the more complete approach. However, if you prefer a progressive, safer merge of the fundamental lifecycle infrastructure first, this PR is ready to go.

ykhfree:
> > This PR seems to be the same as the [PR I registered](https://github.com/resilience4j/resilience4j/pull/2452). What do you think?
> 
> Thanks for pointing that out.
> 
> I did review #2452 while working on this issue.
> 
> Both PRs seem to address the lifecycle problem from #2451, but my approach was a bit different. I focused on adding AutoCloseable support directly to Hedge/HedgeRegistry, daemon scheduler configuration, and virtual-thread compatibility around the existing scheduler infrastructure.
> 
> I think there is definitely overlap in the problem being solved, but the implementation direction is somewhat different.
> 
> It would be great to hear which approach the maintainers prefer.

Thank you for your response.

PR #2487, opened 2026-06-28
fix: HedgeConfig.ofDefaults() uses AVERAGE_PLUS as default supplier type

### LINKED ISSUE #2457: HedgeConfig.ofDefaults() creates an unusable default duration supplier
## Description

`HedgeConfig.ofDefaults()` currently creates a config whose duration supplier type is `PRECONFIGURED`, but the configured cutoff duration is `null`.

This means `HedgeDurationSupplier.fromConfig(HedgeConfig.ofDefaults())` creates a `PreconfiguredDurationSupplier` with a `null` cutoff. The supplier can then return `null`, and the default `Hedge` execution path may fail later when `HedgeImpl` calls `durationSupplier.get().toNanos()`.

There is a related copy-builder issue in `HedgeConfig.from(...)`: `HedgeConfig.Builder(HedgeConfig baseConfig)` does not copy `durationSupplierType` from the source config. As a result, a config using `AVERAGE_PLUS` can silently become `PRECONFIGURED` after being copied.

## Tests

The issue can be reproduced by adding the following tests to `HedgeConfigTest`. When executed, `shouldCreateUsableDefaultDurationSupplier` fails because the default duration supplier returns `null`, and `shouldInitializeDurationSupplierTypeFromOtherConfig` fails because the copied config's supplier type is `PRECONFIGURED` instead of the expected `AVERAGE_PLUS`.

```java
@Test
public void shouldCreateUsableDefaultDurationSupplier() {
    HedgeDurationSupplier supplier = HedgeDurationSupplier.fromConfig(HedgeConfig.ofDefaults());

    then(supplier.get()).isNotNull();
}

@Test
public void shouldInitializeDurationSupplierTypeFromOtherConfig() {
    HedgeConfig config = HedgeConfig.custom()
        .averagePlusAmountDuration(200, false, 100).build();

    HedgeConfig copiedConfig = HedgeConfig.from(config).build();

    then(copiedConfig.getDurationSupplier())
        .isEqualTo(HedgeConfig.HedgeDurationSupplierType.AVERAGE_PLUS);
}
```

Failure result:

```bash
HedgeConfigTest > shouldInitializeDurationSupplierTypeFromOtherConfig FAILED
    org.opentest4j.AssertionFailedError at HedgeConfigTest.java:70

HedgeConfigTest > shouldCreateUsableDefaultDurationSupplier FAILED
    java.lang.AssertionError at HedgeConfigTest.java:59
```

## Expected behavior

- `HedgeConfig.ofDefaults()` should create a usable default duration supplier.
- `HedgeConfig.from(existingConfig).build()` should preserve the source config's duration supplier type.

Prabhanshu-dev:
Fixes #2457

Problem:
HedgeConfig.ofDefaults() sets hedgeDurationSupplierType to PRECONFIGURED 
with cutoff = null by default. This causes supplier.get() to return null, 
leading to NPE in HedgeImpl at runtime when durationSupplier.get().toNanos() 
is called.

Fix:
Changed default hedgeDurationSupplierType from PRECONFIGURED to AVERAGE_PLUS 
in Builder, making ofDefaults() produce a usable configuration out of the box.

Tests:
Added shouldCreateUsableDefaultDurationSupplier() to HedgeConfigTest.
All 7 tests pass.

### DIFF
--- a/resilience4j-hedge/src/main/java/io/github/resilience4j/hedge/HedgeConfig.java
+++ b/resilience4j-hedge/src/main/java/io/github/resilience4j/hedge/HedgeConfig.java
@@ -130,7 +130,7 @@ public class HedgeConfig implements Serializable {
 
     public static class Builder {
 
-        private HedgeDurationSupplierType hedgeDurationSupplierType = HedgeDurationSupplierType.PRECONFIGURED;
+        private HedgeDurationSupplierType hedgeDurationSupplierType = HedgeDurationSupplierType.AVERAGE_PLUS;
         private boolean shouldUseFactorAsPercentage = false;
         private int hedgeTimeFactor = 0;
         private boolean shouldMeasureErrors = true;
--- a/resilience4j-hedge/src/main/java/io/github/resilience4j/hedge/internal/AverageDurationSupplier.java
+++ b/resilience4j-hedge/src/main/java/io/github/resilience4j/hedge/internal/AverageDurationSupplier.java
@@ -35,6 +35,7 @@ public class AverageDurationSupplier implements HedgeDurationSupplier {
     final boolean shouldMeasureErrors;
     final int factor;
     final FixedSizeSlidingWindowMetrics metrics;
+    private static final Duration MIN_HEDGE_DELAY = Duration.ofMillis(1);
 
     /**
      * @param shouldUseFactorAsPercentage whether to use factor as a percentage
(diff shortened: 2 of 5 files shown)


Prabhanshu-dev:
Added a 1 ms minimum delay in AverageDurationSupplier.get() so the default configuration never returns Duration.ZERO, even when there is no metrics data available. This ensures AVERAGE_PLUS continues to work as the adaptive default instead of falling back to the fixed PRECONFIGURED cutoff.

Updated the existing tests to match the new behavior, added new test cases for different scenarios (cold start, warm window, and different factors), and improved the regression test to verify the delay is greater than Duration.ZERO. All hedge module tests are passing, and the pull request is ready for review.

Prabhanshu-dev:
Hi @RobWin, just following up on this PR — happy to make any changes if there's feedback. Let me know if anything needs adjusting. Thanks for your time!

Issue #2324, opened 2025-06-20
If `ignoreThrowablePredicate` throws it can lead to a spin.


HuwCampbell:
This is specific to the Kotlin wrapper.

We found that if `ignoreThrowablePredicate` throws it can lead to a thread / CPU spinning.

The reasons is quite simple, as if there is a [throw](https://github.com/resilience4j/resilience4j/blob/7d6895d281ffd94daf38e4705702ee274376a80c/resilience4j-kotlin/src/main/kotlin/io/github/resilience4j/kotlin/circuitbreaker/CircuitBreaker.kt#L52-L58) inside the circuit breaker handler, then none of `releasePermission`, `onError`, or `onResult` are called; a resource has been acquired which will never be released.

While "don't do that" is good advice, I think an additional try catch over this user supplied function would be prudent.

davsclaus:
is this only kotlin, maybe its also possible with java - it may be worth to double check there too.

RobWin:
Hello,
yes, good finding. Would you like to create a PR?

HuwCampbell:
Ok. There's some choices to be made there as well though.

HuwCampbell:
Look at the [Java code now](https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java#L205-L222). I think it has the same issue.

When the functions `recordExceptionPredicate` and `ignoreExceptionPredicate` are executed they are not wrapped in a try catch, and it appears they could cause a similar resource leak.

I don't know why the Kotlin wrapper actually has the function now at all, seeing as it appears to be duplicating the functionality of the Java client? Maybe just so it's per call?

AryamannSingh7:
Following up on this — #2332 addresses the Kotlin wrapper, but as @HuwCampbell noted above, the same issue exists in the core Java `CircuitBreakerStateMachine`. In `handleThrowable` (and `onResult`), the user-supplied `ignoreException` / `recordException` / `recordResult` predicates are evaluated without a `try/catch`. If one of them throws, the call's permit is neither released nor recorded — so in `HALF_OPEN` the breaker leaks its trial permit and can get permanently wedged (no further trial calls, never transitions).

[...]

PR #2488, opened 2026-06-30
Issue #2324: Prevent permit leak when a CircuitBreaker record/ignore predicate throws

AryamannSingh7:
Addresses the Java-core part of #2324 (complements the Kotlin wrapper fix in #2332, which does not touch `CircuitBreakerStateMachine`).

### Problem
In `CircuitBreakerStateMachine`, the user-supplied `ignoreException` / `recordException` predicates (in `handleThrowable`) and `recordResult` predicate (in `onResult`) are evaluated without a `try/catch`. A permission is acquired before the protected call; if one of these predicates throws while handling the outcome, neither `releasePermission()` nor `stateReference.onError()/onSuccess()` is reached, so the acquired permit is **neither released nor recorded**.

In the `HALF_OPEN` state, where the number of permitted calls is limited, this leaks a trial permit. With enough such failures the breaker can no longer permit calls and can never transition — it is wedged.

### Fix
Predicate evaluations now go through a small helper that releases the permission and rethrows the original exception if the predicate fails, leaving the breaker in a consistent state. `Exception` (not `Throwable`) is caught so `Error`s and interrupts are not affected; the original exception is rethrown so the caller still observes the failure. `releasePermission()` is already a no-op in `CLOSED` and restores the permit in `HALF_OPEN`, so this is safe in every state and consistent with the existing ignored-exception branch.

### Tests
Added `CircuitBreakerThrowingPredicateLeakTest` with a minimal reproduction for both the exception-predicate and result-predicate paths: a `HALF_OPEN` breaker with a throwing predicate, asserting the permit is released after the predicate throws. Both tests fail without the change and pass with it; the full `resilience4j-circuitbreaker` suite is green.

Open to adjusting the approach (e.g. `Exception` vs `Throwable`, or recording the failure instead of rethrowing) based on your preference.

### DIFF
--- a/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
+++ b/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/internal/CircuitBreakerStateMachine.java
@@ -41,6 +41,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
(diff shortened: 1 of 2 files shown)


davsclaus (approved):

AryamannSingh7:
Hi @davsclaus, thanks again for the review and approval. Just a gentle nudge on this one — is there anything else you'd like me to address before it can be merged, or is it good to go? Happy to rebase if it's fallen behind master. Thanks!

davsclaus:
I am not a commiter on this project

AryamannSingh7:
Thanks for clarifying, @davsclaus — appreciate the review regardless!

@RobWin, would you have a chance to take a look at this one when you get time? It's a small, well-scoped fix (#2324) with tests, and Claus has kindly reviewed it. Happy to rebase if needed. Thanks!

Issue #2489, opened 2026-06-30
SemaphoreBasedRateLimiter: lowering limitForPeriod below available permits stops the scheduled refresh task


zbnerd:
### Summary

`SemaphoreBasedRateLimiter.refreshLimit()` computes `permissionsToRelease = limitForPeriod - semaphore.availablePermits()` and calls `semaphore.release(permissionsToRelease)`. After `changeLimitForPeriod(newLimit)` lowers the limit below the currently available permits, `permissionsToRelease` becomes negative on the next refresh. `Semaphore.release(int)` throws `IllegalArgumentException` for negative values, and because `refreshLimit` runs via `scheduleAtFixedRate`, that uncaught exception suppresses all subsequent executions (JDK contract). The refresh task stops executing and the limiter no longer replenishes permits for the lifetime of that instance.

Verified against the current upstream `master`: the implementation of `refreshLimit()` still contains the same code path.

```java
void refreshLimit() {
    int permissionsToRelease =
        this.rateLimiterConfig.get().getLimitForPeriod() - semaphore.availablePermits();
    semaphore.release(permissionsToRelease);
}
```

### Steps to reproduce

```java
RateLimiterConfig cfg = RateLimiterConfig.custom()
    .limitForPeriod(5)
    .limitRefreshPeriod(Duration.ofMillis(5))
    .timeoutDuration(Duration.ofSeconds(1))
    .build();
SemaphoreBasedRateLimiter rl = new SemaphoreBasedRateLimiter("demo", cfg);

rl.changeLimitForPeriod(1);   // availablePermits(5) > new limit(1)
// next scheduled refreshLimit(): permissionsToRelease = 1 - 5 = -4
//   → semaphore.release(-4) throws IllegalArgumentException
//   → scheduleAtFixedRate suppresses all future executions
// limiter never refreshes again for the lifetime of this instance
```

### Expected

The JavaDoc of `RateLimiter#changeLimitForPeriod` states:

> "New limit won't affect current period permissions and will apply only from next one."

There is no documented restriction that the new limit must be greater than or equal to the current number of available permits, nor is lowering the limit documented as unsupported.

### Actual

The first `refreshLimit` after a downward `changeLimitForPeriod` throws `IllegalArgumentException`; the periodic refresh task terminates and no longer replenishes permits.

### Why this appears to be a bug

[...]

PR #2490, opened 2026-06-30
fix(ratelimiter): prevent SemaphoreBasedRateLimiter refresh task from terminating after a downward limit change

zbnerd:
## Problem

`SemaphoreBasedRateLimiter.refreshLimit()` computes

```java
int permissionsToRelease =
    rateLimiterConfig.get().getLimitForPeriod()
        - semaphore.availablePermits();
semaphore.release(permissionsToRelease);
```

After `changeLimitForPeriod()` lowers the configured limit, the number of available permits may temporarily exceed the new limit until the next refresh period. In that situation, `permissionsToRelease` becomes negative.

Since `Semaphore.release(int)` throws `IllegalArgumentException` for negative values, the scheduled refresh task terminates unexpectedly. According to the JDK contract of `ScheduledExecutorService.scheduleAtFixedRate(...)`, an uncaught exception suppresses all subsequent executions of that periodic task.

As a result, the scheduled refresh task stops executing and the rate limiter no longer replenishes permits for the lifetime of that instance.

## Why this appears to be a bug

The JavaDoc of `RateLimiter#changeLimitForPeriod` states:

> "New limit won't affect current period permissions and will apply only from next one."

There is no documented restriction that the new limit must be greater than or equal to the current number of available permits, nor is lowering the limit documented as unsupported.

The existing implementation of `AtomicRateLimiter` supports decreasing the limit without terminating its internal state update mechanism. While the two implementations use different internal algorithms, they expose the same public API and currently behave differently when the configured limit is reduced.

## Proposed fix

Handle all three cases in `refreshLimit`:

```java
if (permissionsToRelease > 0) {
    semaphore.release(permissionsToRelease);           // deficit: top up to the limit
} else if (permissionsToRelease < 0) {
    semaphore.tryAcquire(-permissionsToRelease);      // excess: clamp down to the new limit
}
```

- **deficit** (`> 0`): release to top up
- **equal** (`0`): no-op
- **excess** (`< 0`): `tryAcquire` the excess so available permits reach the new limit

The clamp mirrors `AtomicRateLimiter`'s behaviour of capping excess permits at each cycle boundary (`min(nextPermissions + accumulated, permissionsPerCycle)`) and aligns with the `changeLimitForPeriod` JavaDoc that the new limit "will apply only from next one".

**Concurrency safety.** `Semaphore.tryAcquire(int)` is non-throwing and all-or-nothing: if permits are concurrently consumed between the `availablePermits` read and the `tryAcquire` call so that the full excess is no longer available, it returns `false` (never throws) and the surplus persists until the next refresh tick clamps it. The scheduled refresh task therefore survives any concurrent acquire traffic. The companion test `refreshLimitClampsExcessPermitsWithoutThrowing` exercises this contract directly.

The goal remains intentionally limited: clamp excess + keep the scheduler alive. If maintainers prefer a different semantic for downward limit changes (for example adjusting the semaphore state inside `changeLimitForPeriod()`), I'm happy to update this PR accordingly.

## Regression test

This PR adds two regression tests:

- `refreshSchedulerMustSurviveLimitDecreaseBelowAvailablePermits` uses a real `ScheduledThreadPoolExecutor` (a counting subclass, not a mock), lowers `limitForPeriod` below the currently available permits, and asserts that the periodic refresh keeps running and that excess permits are clamped down to the new limit.
- `refreshLimitClampsExcessPermitsWithoutThrowing` exercises the clamp branch directly by calling `refreshLimit()` after lowering the limit and verifies both that excess permits are removed and that no exception escapes (the `Semaphore.tryAcquire(int)` non-throwing contract).

Waits are condition-based with a nanoTime deadline (no `Thread.sleep`), and the tests depend only on JUnit 5, so no extra test libraries are needed on the module classpath.

Without this fix:

- the first test fails because `Semaphore.release(int)` throws `IllegalArgumentException` and the scheduled refresh task terminates (the periodic task stops firing and excess permits are never clamped)
- the second test fails directly with `IllegalArgumentException` from `refreshLimit()`

With this change:

- both tests pass

The existing `changeLimitForPeriod` tests only cover increasing the limit and use a mocked scheduler, so this execution path was previously untested.

The commits are intentionally split into a regression test followed by the fix, making it easy to verify that the tests fail before the fix and pass afterwards. The full `resilience4j-ratelimiter` test suite passes.

Relates to #2489.

### DIFF
--- a/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
+++ b/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiter.java
@@ -131,7 +131,11 @@ public class SemaphoreBasedRateLimiter implements RateLimiter {
     void refreshLimit() {
         int permissionsToRelease =
             this.rateLimiterConfig.get().getLimitForPeriod() - semaphore.availablePermits();
-        semaphore.release(permissionsToRelease);
+        if (permissionsToRelease > 0) {
+            semaphore.release(permissionsToRelease);
+        } else if (permissionsToRelease < 0) {
+            semaphore.tryAcquire(-permissionsToRelease);
+        }
     }
 
     /**
--- a/resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiterRefreshLimitRegressionTest.java
+++ b/resilience4j-ratelimiter/src/test/java/io/github/resilience4j/ratelimiter/internal/SemaphoreBasedRateLimiterRefreshLimitRegressionTest.java
@@ -1,0 +1,199 @@
+package io.github.resilience4j.ratelimiter.internal;
+
+import io.github.resilience4j.ratelimiter.RateLimiterConfig;
+import org.junit.jupiter.api.Test;
+
+import java.time.Duration;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BooleanSupplier;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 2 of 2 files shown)


zbnerd:
@RobWin I’ve addressed the Copilot review comments and pushed the updates. When you have a chance, could you PTAL?

PR #2493, opened 2026-07-09
Issue #2492: Fixed ThreadPoolBulkheadConfig.from() mutating the base config

### LINKED ISSUE #2492: ThreadPoolBulkheadConfig.from(config) mutates the base config instead of copying it

```java
    this.config = bulkheadConfig;
}
```

### Impact

```java
@Test
        .maxThreadPoolSize(4)
        .coreThreadPoolSize(2)
        .queueCapacity(10)
        .build();

        .maxThreadPoolSize(20)
        .coreThreadPoolSize(8)
        .queueCapacity(50)
        .build();

}
```

### Environment

Develop-KIM:
`ThreadPoolBulkheadConfig.Builder(ThreadPoolBulkheadConfig)` aliased the passed-in config (`this.config = bulkheadConfig`) instead of copying its fields, so `from(base)...build()` mutated `base` and returned the same instance. This copies each field into a fresh config, matching `BulkheadConfig`, `RateLimiterConfig`, `TimeLimiterConfig` and `CircuitBreakerConfig`. The context propagator list is copied into a new `ArrayList` so `build()`'s `addAll` no longer leaks into the base config.

Added `ThreadPoolBulkheadConfigTest` cases that keep a reference to the base config and assert it stays unchanged (the existing `createFromBaseConfig` test passed the base inline, so it could not catch this).

Fixes #2492. This was written with AI assistance (see the `Co-Authored-By` trailer).

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadConfig.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadConfig.java
@@ -133,7 +133,14 @@ public class ThreadPoolBulkheadConfig {
         private ThreadPoolBulkheadConfig config;
 
         public Builder(ThreadPoolBulkheadConfig bulkheadConfig) {
-            this.config = bulkheadConfig;
+            this.config = new ThreadPoolBulkheadConfig();
+            this.config.maxThreadPoolSize = bulkheadConfig.maxThreadPoolSize;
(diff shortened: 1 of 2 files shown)


arjun2075:
Independently hit this same bug and arrived at the same fix — can confirm this resolves #2492 (verified locally on JDK 21, both tests fail without the change and pass with it). +1 to getting this merged.

Develop-KIM:
Closing this one — I'm stepping back from open-source contributions and won't be able to follow through. Thanks for the review.

PR #2510, opened 2026-09-05
Issue #2299: Fixed missing Timer metrics in Prometheus

### LINKED ISSUE #2299: resilience4j.micrometer.Timer + Prometheus: some metrics are not published because of different tags
Resilience4j version: newest

Hello,

I'm using resilience4j.micrometer.Timer in a Spring Boot app that publishes the metrics for Prometheus.

In this setup not all timer metrics are published.
Either metrics with kind="success" are published or metrics with  kind="failure", but never both, depending on which kind occurred first.
This is caused by io.micrometer.prometheusmetrics.PrometheusMeterRegistry. It logs a warning:

> 2025-04-01T10:41:11.254+02:00  WARN 34466 --- i.m.p.PrometheusMeterRegistry            : The meter (MeterId{name='resilience4j.timer.calls', tags=[tag(kind=successful),tag(name=testcall)]}) registration has failed: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'resilience4j.timer.calls' containing tag keys [failure, kind, name]. The meter you are attempting to register has keys [kind, name]. Note that subsequent logs will be logged at debug level.

A pragmatic fix would be, to always record the metrics with the same tags.
E.g. change the lines following https://github.com/resilience4j/resilience4j/blob/d6e80d7023d621db1a982daf768c8ec1b469766f/resilience4j-micrometer/src/main/java/io/github/resilience4j/micrometer/internal/TimerImpl.java#L140 to:
```java
      io.micrometer.core.instrument.Timer.Builder calls = builder(timerConfig.getMetricNames())
          .description("Timed decorated operation calls")
          .tag(NAME, name)
          .tag(KIND, resultKind)
          .tag(FAILURE_TAG, throwable == null ? "" : timerConfig.getOnFailureTagResolver().apply(throwable));
```

A slightly cleaner alternative would be to use different names for the metric for success and failure (which can have different tags).
This would break existing queries and dashboards (but they are already broken).

Both alternatives have been suggested in
https://github.com/micrometer-metrics/micrometer/issues/877 and
https://github.com/prometheus/client_java/issues/696.

Similar issues already have been mentioned since 2020 in multiple places.
But it seems there is no common understanding, if this should be fixed in Prometheus itself, the Prometheus-client, Micrometer, or the client-code that uses Micrometer (e.g.  resilience4j.micrometer.Timer).
So it probable doesn't make sense to wait for the root cause to be resolved somewhere lower in the stack.

zbnerd:
Prometheus only exports the first recorded Timer outcome because successful and failed calls have different tag keys.
Always add the `failure` tag, using an empty value for successful calls, so both outcomes are exported regardless of
registration order.

Add six regression cases using actual Prometheus scrape output, covering both outcome orders, different Timer names,
and custom metric names, tags and failure resolvers. Update the shared Timer assertion helper for the consistent tag
set.

  Validation:

  - Micrometer module: 152 tests passed, including six regression cases confirmed to fail before the fix.
  - Timer integration tests in Reactor, Spring 6, Spring Boot 3 and Spring Boot 4: 24 tests passed.
  - `./gradlew check --no-daemon` fails at `CircuitBreakerConfigTest.testToString()`: the test expects
  `CircuitBreakerConfig {`, while the implementation returns `CircuitBreakerConfig{...`. The same failure was reproduced
  on unmodified `master` (`a8a33164`) by running that test in a separate checkout.

  Fixes #2299

### DIFF
--- a/resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfigTest.java
+++ b/resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfigTest.java
@@ -576,7 +576,7 @@ class CircuitBreakerConfigTest {
         String result = config.toString();
 
         assertThat(result)
-                .startsWith("CircuitBreakerConfig {")
+                .startsWith("CircuitBreakerConfig{")
                 .contains("slidingWindowSize=5")
                 .contains("recordExceptions=[class java.lang.RuntimeException]")
                 .contains("automaticTransitionFromOpenToHalfOpenEnabled=true")
--- a/resilience4j-hedge/src/test/java/io/github/resilience4j/hedge/HedgeConfigTest.java
+++ b/resilience4j-hedge/src/test/java/io/github/resilience4j/hedge/HedgeConfigTest.java
@@ -29,7 +29,8 @@ import io.github.resilience4j.hedge.internal.HedgeDurationSupplier;
 class HedgeConfigTest {
 
     private static final String HEDGE_DURATION_MUST_NOT_BE_NULL = "HedgeDuration must not be null";
-    private static final String HEDGE_TO_STRING = "HedgeConfig{false,0,true,100,null}";
+    private static final String HEDGE_TO_STRING = "HedgeConfig{shouldUseFactorAsPercentage=false, "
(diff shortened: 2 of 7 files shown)


zbnerd:
@RobWin Hi, just a gentle follow-up on this PR. CI is green, and I’d appreciate a review whenever you have a chance. Happy to address any feedback. Thanks!

PR #2511, opened 2026-09-07
Issue #2492: Copy the base config in ThreadPoolBulkheadConfig.from()

### LINKED ISSUE #2492: ThreadPoolBulkheadConfig.from(config) mutates the base config instead of copying it

GangEunzzang:
Fixes #2492.

`ThreadPoolBulkheadConfig.Builder(ThreadPoolBulkheadConfig)` kept the passed-in config by reference, so every builder call wrote through to it and `build()` returned the same instance. `from(baseConfig)` therefore mutated the base config — when several thread-pool bulkheads share one (Spring Boot's `default` config via `CommonThreadPoolBulkheadConfigurationProperties`), the first instance's overrides leaked into the base and context propagators accumulated across instances.

The other config builders (`BulkheadConfig`, `RateLimiterConfig`, `TimeLimiterConfig`, `CircuitBreakerConfig`) copy the base config's fields into a fresh instance; this does the same here. `contextPropagators` is copied into a new list so `build()`'s `addAll` cannot reach the base either.

Tests: `fromShouldNotMutateBaseConfig` (the reproduction from the issue, extended to all copied fields) and `fromShouldNotAccumulateContextPropagatorsOnBaseConfig` (base keeps 1 propagator, each derived config gets its own 2). Both fail on `master` and pass with the fix; `resilience4j-bulkhead` tests are otherwise unchanged.

### DIFF
--- a/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadConfig.java
+++ b/resilience4j-bulkhead/src/main/java/io/github/resilience4j/bulkhead/ThreadPoolBulkheadConfig.java
@@ -70,7 +70,8 @@ public class ThreadPoolBulkheadConfig {
     }
 
     /**
-     * Returns a builder to create a custom ThreadPoolBulkheadConfig.
+     * Returns a builder to create a custom ThreadPoolBulkheadConfig based on another
+     * ThreadPoolBulkheadConfig. The given config is copied and left unchanged.
      *
      * @return a {@link Builder}
      */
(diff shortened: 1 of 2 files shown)


GangEunzzang:
Closing this — I should have checked the issue before opening it. #2500 (and #2507, #2509) already carry the same fix, and #2500 has been green since August, so a fifth copy doesn't help anyone. Sorry for the noise.

PR #2513, opened 2026-09-10
Issue #2512: Check the initial interval when building capped backoff functions

### LINKED ISSUE #2512: Capped exponential backoff factories validate the initial interval on first use instead of at construction
Resilience4j version: 2.4.0 and current `master` (`a8a3316`)

Java version: 21

## Problem

The capped factories validate the maximum interval eagerly but build their uncapped delegate inside the returned lambda, so the initial interval is only validated on first use.

```java
IntervalFunction.ofExponentialBackoff(0, 2.0);          // throws IllegalArgumentException
IntervalFunction.ofExponentialBackoff(0, 2.0, 5000);    // returns normally
```

`ofExponentialRandomBackoff(long, double, double, long)` is the same.

Zero is legal elsewhere, since `RetryConfig.Builder.waitDuration` takes `Duration.ZERO` as retry immediately, but it cannot mean that here: `ofExponentialBackoff(initial, multiplier)` is `of(initial, x -> x * multiplier)`, so a zero initial interval stays zero forever.

Configuration reaches it. `wait-duration: 0` passes `CommonRetryConfigurationProperties.setWaitDuration`, which rejects only negatives. With `enable-exponential-backoff` and a multiplier set, `exponential-max-wait-duration` decides which factory is used: without it the two-argument overload rejects the value while the registry is built, with it the capped overload accepts it.

`RetryImpl` calls the interval function outside the `try` block that rethrows the exception being retried, so the validation error replaces the original failure. Today this fails with `IllegalArgumentException: Illegal argument interval: 0 millis is less than 1` rather than the exception it was retrying:

```java
@Test
void exceptionBeingRetriedShouldSurviveTheRetry() {
    RetryConfig config = RetryConfig.custom()
        .maxAttempts(3)
        .intervalFunction(IntervalFunction.ofExponentialBackoff(0, 2.0, 5000))
        .build();
    Retry retry = Retry.of("example", config);
    Supplier<String> supplier = Retry.decorateSupplier(retry, () -> {
        throw new IllegalStateException("real failure");
    });

    assertThatThrownBy(supplier::get)
        .isInstanceOf(IllegalStateException.class)
        .hasMessage("real failure");
}
```

## Compatibility

Checking eagerly is a behaviour change: an application with that configuration starts today, and would fail to start instead. #2108 made the same move for the randomization factor in one of these two factories.

jjh75607:
Fixes #2512

## Change

`ofExponentialBackoff(long, double, long)` and `ofExponentialRandomBackoff(long, double, double, long)` built their uncapped delegate inside the returned lambda. Hoisting it out moves the initial interval check to construction and stops the delegate being rebuilt on every attempt.

The explicit `checkRandomizationFactor` is now duplicated by the hoisted delegate. I kept it so an out-of-range factor is still reported before an invalid interval, and locked that ordering with an assertion.

The uncapped randomized overload rebuilds its delegate the same way. It already validates eagerly, so I left it out to keep this focused.

## Tests

Five assertions in `IntervalFunctionTest.shouldRejectNonPositiveDuration`, which covered a non-positive maximum interval for the capped overloads but not a non-positive initial one. They fail on `master`.

## Verification

JDK 21. `core`, `retry`, `framework-common`, `spring6`, `spring-boot3`, `reactor`, `kotlin` and `vavr` are green.

`CircuitBreakerConfigTest.testToString` fails, unrelated to this change: `master` has been red for the same reason since #2458 (run 33362579012). It expects `startsWith("CircuitBreakerConfig {")` while `CircuitBreakerConfig.toString()` produces `CircuitBreakerConfig{`.


jjh75607:
The `Build and publish` failure is pre-existing on `master`, not from this PR.

#2458 changed the `CircuitBreakerConfig.toString()` prefix from `"CircuitBreakerConfig {"`
to `"CircuitBreakerConfig{"` without updating the assertion, which still expects the space:

- `resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfig.java:250`
- `resilience4j-circuitbreaker/src/test/java/io/github/resilience4j/circuitbreaker/CircuitBreakerConfigTest.java:579`

`master` has been red since that commit
(https://github.com/resilience4j/resilience4j/actions/runs/33362579012), and every PR run
since then fails the same way. #2510 already carries the fix (commit 9c5776e). Once it
lands, a CI re-run here should be green: this PR only touches `IntervalFunction` and its test.

PR #2515, opened 2026-09-12
Issue #2183: Add lastResult to RetryOnRetryEvent

### LINKED ISSUE #2183: Add RetryOnRetryEvent.lastResult
I'm trying to implement a generic logging via `retry.getEventPublisher().onRetry(event -> log.warn("Retrying...`

The event provides access to the exception if `retry-exception-predicate` is used.
But if `result-predicate` is used its a similar situation, unfortunately there is no way to get the result value that triggered the retry.

It would be great if `RetryOnRetryEvent`/`AbstractRetryEvent` would be extended to contain the last result too.

elgamal-ahmed:
RetryOnRetryEvent only carried the last thrown Throwable, so retries triggered by a result predicate (retryOnResult) left event consumers with no way to see the value that caused the retry.

Add a getLastResult() accessor, defaulted to null on the RetryEvent interface so existing implementations keep compiling, and overridden in RetryOnRetryEvent via a new constructor overload. The previous constructor is kept and simply passes null, so this is source and binary compatible. RetryImpl now populates the result on both the synchronous and CompletionStage retry paths.

Closes #2183


elgamal-ahmed:
Hi @RobWin, could you approve the workflow run when you get a chance? It is waiting on approval since this is my first PR here. Happy to adjust anything in the change. Thanks!

RobWin:
Executed it, sry I'm super busy at work right now. Took some time.

